Issue #54, opened 2014-12-26
Transpose Multiply Operation


jonlawlor:
There are many cases where we want to multiply a transpose of a matrix by another.  For example, in the stat.CovarianceMatrix function, the covariance matrix is calculated by first centering the input, making a transpose copy of it, and then matrix multiplying the original by the copy.  It would be more memory and time efficient if we could avoid the transpose copy.  I think we should have an additional method of Dense which would be able to express that kind of transpose multiply operation.

BLAS usually allows the caller to indicate if an input matrix is transpose via an additional input argument - see for example DGEMM and DSYRK.  In R, the same effect is produced by using tcrossprod (http://stat.ethz.ch/R-manual/R-patched/library/Matrix/html/tcrossprod.html).  I'm not sure what MATLAB does behind the scenes, but I strongly suspect that the jitter recognizes these operations, or maybe they encode transpose via a flag in the matrix objects themselves.  I believe we've already discussed that approach and came out against it.

That leaves two ways to do this - we can either have a more general multiply operation, which would have additional inputs to indicate if the first and/or second matrix was a transpose, following the function signature of BLAS.  Alternatively, we could have an additional, specific TMul function, which would only perform A * B'.  (Or, perhaps a TMul which indicates A' * B, MulT which indicates A * B', and TMulT which indicates A' * B').

An additional possibility is that in the cases where multiply operations are taking place between two matrices and one of them is nil, we could take that to mean that the nil argument is the same as the other argument.  For example, a "MulT" from above, with m.MulT(A, nil) would mean that A * A' should be stored into m.  In this way we could (at some point) take advantage of the symmetry of the operation.

What are your thoughts?  I think I prefer more simple methods rather than fewer complicated methods, but the main thing is to be able to express these relationships.

jonlawlor:
#61 fixes this

PR #61, opened 2014-12-29
Mulgen

jonlawlor:
as per #54

### DIFF
--- a/mat64/dense_arithmetic.go
+++ b/mat64/dense_arithmetic.go
@@ -476,6 +476,9 @@ func (m *Dense) MulGen(a Matrix, at bool, b Matrix, bt bool) {
 		panic(ErrShape)
 	}
 
+	//TODO(jonlawlor): if we are performing a * a' or a' * a, then make a call to
+	// DSYRK instead of DGEMM.
+
 	if a, ok := a.(RawMatrixer); ok {
 		if b, ok := b.(RawMatrixer); ok {
 			amat, bmat := a.RawMatrix(), b.RawMatrix()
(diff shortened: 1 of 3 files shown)


mat64/dense_arithmetic.go
```diff
@@ -434,6 +434,168 @@ func (m *Dense) Mul(a, b Matrix) {
...
+	if a, ok := a.(Vectorer); ok {
+		if b, ok := b.(Vectorer); ok {
+			row := make([]float64, ac)
+			col := make([]float64, br)
+			if blasEngine == nil {
+				panic(ErrNoEngine)
+			}
+			if aTrans != blas.NoTrans {
+				if bTrans != blas.NoTrans {
+					for r := 0; r < ar; r++ {
```
jonlawlor: The trouble is that would require a type assert to access the mat field, and we don't currently have a type that implements Vectorer but doesn't implement RawMatrixer.

jonlawlor:
I've changed the code to use booleans, but I don't see how I can implement @btracey's suggestion.  Is it good to go without it?

kortschak:
LGTM but again wait for @btracey.

mat64/dense_arithmetic.go
kortschak: You don't need to assert to get the mat field; w is a Dense.

jonlawlor: Ah, yes you're right.

jonlawlor:
OK, I've implemented that as well.  I haven't benchmarked it though.

btracey:
LGTM.

PR #63, opened 2015-01-01
Make goblas the default BLAS implementation

kortschak:
Also allow cblas use in testing with `go test -tags cblas`.

### DIFF
--- a/mat64/cblas_test.go
+++ b/mat64/cblas_test.go
@@ -1,0 +1,15 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//+build cblas
+
+package mat64
+
+import (
+	"github.com/gonum/blas/cblas"
+)
+
+func init() {
+	Register(cblas.Blas{})
+}
--- a/mat64/dense_test.go
+++ b/mat64/dense_test.go
@@ -5,12 +5,11 @@
 package mat64
 
 import (
-	"github.com/gonum/floats"
 	"math"
-
 	"math/rand"
 	"testing"
 
+	"github.com/gonum/floats"
 	"gopkg.in/check.v1"
 )
 
@@ -492,7 +491,7 @@ func (s *S) TestMul(c *check.C) {
 
 		// These probably warrant a better check and failure. They should never happen in the wild though.
 		temp.mat.Data = nil
-		c.Check(func() { temp.Mul(a, b) }, check.PanicMatches, "cblas: index of c out of range", check.Commentf("Test %d"))
+		c.Check(func() { temp.Mul(a, b) }, check.PanicMatches, "general: insufficient length", check.Commentf("Test %d"))
 	}
 }
 
--- a/mat64/dense_test.go
+++ b/mat64/dense_test.go
@@ -604,6 +604,9 @@ func (s *S) TestPow(c *check.C) {
 					d[i] = math.NaN()
 				}
 				*a = *NewDense(10, 10, d).View(1, 1, 3, 3).(*Dense)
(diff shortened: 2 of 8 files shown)


jonlawlor:
It appears that it fails on FAIL: dense_test.go:551: S.TestPow.  I get the same failure on my machine.  Is there an issue with Mul?

kortschak:
This change highlights two issues.

1. The errors meesages used in goblas and cblas differ, so it is not possible to transparantly drop in one or the other and retain use of the recovered message. This should probably be changed.
2. Mul in goblas does not correctly handle cases where the receiver is full of NaN.

kortschak:
I think I know why this is happening, and I think it shows that cblas is actually the one that is incorrect. In goblas, @btracey does C <- aAB + bC, which means that if (as is the case in this test) C is full of NaN, C is NaN at completion. Why this doesn't happen in cblas? Dunno.

mat64/cblas_test.go
```diff
@@ -0,0 +1,15 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//+build cblas
+
+package mat64
+
```
btracey: Does this mean the package is tested with cblas? Is this file somehow avoided with the go test call?

kortschak: If you want to test with cblas, use `go test -tags cblas`. This does not happen on travis. If you do this locally, it will fail at the moment - we need to harmonise panic values between the two implementations. My proposal would be to drop the c from the beginning of the cblas panics and use those in both - the reasoning being that those messages are more informative (they give a stringer hint to which parameter was incorrect).

btracey: Use the cblas panic messages rather than the goblas ones? SGTM.

I see what this code does now. Thanks. I hadn't used build tags before.

btracey:
LGTM. 

Now that we have a default implementation registered, do we need the explicit nil checks in functions? We could instead move the nil check into register and then it's impossible for it to be nil. Doesn't need to happen here, just a note for the future.

jonlawlor:
Would we also want to have downstream packages (stat, for example) use build tags for goblas vs cblas?

btracey:
I'm under the impression this is a temporary tie-over until there is a single place to register BLAS (located in the blas package). Then there is one tag for all code whether you're interfacing with BLAS directly, mat64 indirectly, or something else.

jonlawlor:
LGTM

kortschak:
PTAL

jonlawlor:
LGTM as well.

kortschak:
>Then there is one tag for all code whether you're interfacing with BLAS directly, mat64 indirectly, or something else.

The tag only specifies the mat64 test behaviour; I don't want to use tags anywhere else for specifying which implementation, that's down to client code, but yes, once we've centralise the registration, a single Register call will do all the work of changing implementation.

mat64/dense_test.go
```diff
@@ -501,7 +501,7 @@ func (s *S) TestMul(c *check.C) {
 
 		// These probably warrant a better check and failure. They should never happen in the wild though.
 		temp.mat.Data = nil
-		c.Check(func() { temp.Mul(a, b) }, check.PanicMatches, "cblas: index of c out of range", check.Commentf("Test %d"))
+		c.Check(func() { temp.Mul(a, b) }, check.PanicMatches, "general: insufficient length", check.Commentf("Test %d"))
```
btracey: Should this panic message be "blas: index out of range" to match https://github.com/gonum/blas/pull/66/files

kortschak: Not yet. When goblas is being harmonised this should change too.

btracey:
LGTM

Issue #37, opened 2014-11-15
mat64.Dense doesn't implement Transposer


jonlawlor:
Looks like an oversight - it does implement TransposeCopier.

jonlawlor:
As a temporary measure, we could also just use Dgemm to multiply by the identity matrix with the transpose flag set.

kortschak:
Dgemm requires that C be allocated and different to A and B, and we then also need to allocate I. This doesn't help us.

jonlawlor:
Yes, you're right.  In the mean time, can we change the definition of Transposer from:

```go
// A Transposer can create a transposed view matrix from the matrix represented by the receiver.
// Changes made to the returned Matrix may be reflected in the original.
type Transposer interface {
	T() Matrix
}
```

to
```go
// A Transposer can transpose a matrix in place.
type Transposer interface {
	T()
}
```
Because I don't see how we can ever create a transpose view without some gnarly stuff behind the scenes.

kortschak:
SGTM

btracey:
SGTM

btracey:
This was instead solved with the Transpose struct. Feel free to reopen if you do not believe so @jonlawlor

jonlawlor:
Yes that definitely works.

This approach is interesting because it is a little matrix expression - it represents a transformation of a matrix as a type, instead of representing a literal matrix.  By coincidence I've been working on a similar line of thought, inspired by some of the ideas in [MadLINQ](http://research.microsoft.com/pubs/158914/euro135-qian.pdf), and also my previous effort with representing relational algebra in go as a set of expressions.  The code is still in a preliminary state, and I mostly intend it to be a proof of concept, but you can find it at [jonlawlor/matrixexp](https://github.com/jonlawlor/matrixexp).

The basic idea is to take the concept you're using in the `Transpose` struct and turn it up to 11.  Instead of evaluating matrix operations, the library builds a graph of expressions, and I intend to write a set of expression compilers that can transform the matrix expressions in various ways, like the automatic use of `blas.Transpose` flags in DGEMM that you have here, but also more complicated operations like ordering `a.Mul(b).Mul(c)`, and asynchronous computation, such as computing both sides of the addition in `(a.Mul(b)).Add(c.Mul(d))` concurrently.  The idea would be that the compilers would be run at init with a `MustCompile` method, so that optimizations could be amortized.

[...]

PR #65, opened 2015-01-02
Change Transposer to be a side effect

jonlawlor:
as per #37


kortschak:
LGTM

kortschak:
This allows Dense to satisfy Transposer by copying the result of TCopy() to itself.

jonlawlor:
Instead of using TCopy, I wrote a loop to do direct copying between slices, which should be faster.  It still makes a complete copy of the data though.

mat64/dense.go
```diff
@@ -379,6 +379,27 @@ func (m *Dense) TCopy(a Matrix) {
 	*m = w
 }
 
+func (m *Dense) T() {
+	// TODO(jonlawlor): use a fully in-place transpose, without
+	// creating another slice with the same size.
+	// create a new slice
+	d := make([]float64, len(m.mat.Data))
+
+	// copy values into the appropriate location
```
kortschak: Complete sentence please.

mat64/dense.go
```diff
@@ -379,6 +379,27 @@ func (m *Dense) TCopy(a Matrix) {
 	*m = w
 }
 
+func (m *Dense) T() {
+	// TODO(jonlawlor): use a fully in-place transpose, without
+	// creating another slice with the same size.
+	// create a new slice
+	d := make([]float64, len(m.mat.Data))
```
kortschak: This potentially allocates to much. When m is a view on another matrix that is wider there are elements that are not used in the resulting matrix data slice.

mat64/dense.go
```diff
@@ -379,6 +379,27 @@ func (m *Dense) TCopy(a Matrix) {
...
+
+	// copy values into the appropriate location
+	cycle := m.mat.Rows*m.mat.Cols - 1
```
kortschak: This assumes Cols == Stride - see above.

jonlawlor: I see your point.  I'll have to put together some tests to represent that case, because I'm not used to thinking that way.

mat64/dense.go
```diff
@@ -379,6 +379,27 @@ func (m *Dense) TCopy(a Matrix) {
 	*m = w
 }
 
+func (m *Dense) T() {
+	// TODO(jonlawlor): use a fully in-place transpose, without
```
btracey: Could we add a fast-path for m square where it's really easy?

jonlawlor:
I've added in a more memory efficient path for square matrices, but your criticism still stands and I'll address it.

mat64/dense.go
kortschak: There are test cases in TestPow that show how to do this.

kortschak:
We need to be aware that this results in two different types of behaviour wrt the backing data. If the matrix is square, the change is reflected in the original backing data (which may also be referenced by another *Dense), but if it is not square, this is not the case. I think this will be surprising. At the moment, the interface definition is silent on this. I think that needs to be addressed.

Since it is not possible to do an in-place transpose with a *Dense representation, presumably the decision we will come to is to say that the transpose is not in place.

kortschak:
To extend this, it is possible to do the fast path as it stands when Cols == Stride.

jonlawlor:
It is possible to do it entirely in place - should I hold off until I can work up a non-square, entirely in place transpose?

jonlawlor:
I've changed transpose to modify the original slice as well.  Now both code paths result in the same side effects.

kortschak:
It is possible to do entirely in place if `Rows*Cols == len(Data)` (i.e. the matrix is not a view).

kortschak:
I really think what you are trying to do here is an impossibility in the general case.

For example, what do you think the appropriate output should be from this code?

```
package main

import (
	"fmt"
)

type fm struct {
	mat64.Matrix
}

func (m fm) Format(fs fmt.State, c rune) {
	if c == 'v' && fs.Flag('#') {
		fmt.Fprintf(fs, "%#v", m.Matrix)
		return
	}
	mat64.Format(m.Matrix, 0, '.', fs, c)
}

func main() {
	d := []float64{
		1, 2, 3, 4,
		5, 6, 7, 8,
		9, 8, 11, 12,
		13, 14, 15, 16,
	}
	a := mat64.NewDense(4, 4, d)
	v := a.View(2, 1, 2, 3).(*mat64.Dense)
	fmt.Printf("a =\n%v\n", fm{a})
	v.T()
	fmt.Printf("a after v' =\n%v\n", fm{a})
	fmt.Printf("v' =\n%v\n", fm{v})
}
```

I think what you should be doing is not attempting to do things deeply in-place.

jonlawlor:
Yes, that would be impossible.  Therefore, View can't return a Transposer.

jonlawlor:
Well, we could actually implement 2 "DenseView" types, and regard the transpose operation as a transpose of the view instead of the underlying data.  In that case, your example above would not modify the underlying dense mat, but it would work.

kortschak:
What is the problem that this is trying to solve?

jonlawlor:
I want dense to be able to do transpose in place, which is a pretty fundamental operation.  Doing so quickly and without significant memory overhead would be a good bonus.

kortschak:
That's the Y in the XY problem. The issue that I have is that it is proliferating complexity for something that has not yet been required.

The question that I have is whether you stricly need it to be in place (i.e. in the same data slice). I can't see why that is important. Allocating a destination slice is (except in the case of square matrices) likely to be cheaper than resolution of cycles to perform the operation in place. This in conjunction with the complexity cost when we consider that Views must now be treated differently and will require separate types that duplicate almost all of the *Dense method set makes this something that I don't see as paying its way.

jonlawlor:
We might not have to have a different type.  In the example previously:
```
func main() {
    d := []float64{
        1, 2, 3, 4,
        5, 6, 7, 8,
        9, 8, 11, 12,
        13, 14, 15, 16,
    }
    a := mat64.NewDense(4, 4, d)
    v := a.View(2, 1, 2, 3).(*mat64.Dense)
    fmt.Printf("a =\n%v\n", fm{a})
    v.T()
    fmt.Printf("a after v' =\n%v\n", fm{a})
    fmt.Printf("v' =\n%v\n", fm{v})
}
```
I'm not entirely sure what would happen to the backing Dense, but it wouldn't do anything illegal, it would just switch around the order of its elements in an odd way.  Any operations on views that cause a side effect to the view is going to cause strange results in the backing matrix though.  The answer might just be to tell people "don't T() a view if you intend on using the backing Dense, instead use TCopy()"

WRT performance when compared with transpose copy, I'll run some benchmarks.  If it isn't faster, then it may remove any motivation to use it.

Normally I would agree that without an example of where to use it, we probably shouldn't add it.  This is just too fundamental, though, and I think we've got to try to fit it in if it can.  I'm not there yet on the implementation, so let me finish and we'll see how it shakes out, and if it doesn't seem to be a net benefit, we can discard it.

kortschak:
OK, we'll see how it goes.

kortschak:
@jonlawlor I'm going to close this since the goal of this PR was to achieve something that #156 and related PRs bring. Please open it again if you disagree.

PR #66, opened 2015-01-09
Make use of gonum/blas/blas64

kortschak:
This is part of a set of changes to rationalise and centralise BLAS implementation registration.

It is expected to fail until gonum/blas#83 has been merged. Review that PR first.

### DIFF
--- a/mat64/cblas_test.go
+++ b/mat64/cblas_test.go
@@ -7,9 +7,10 @@
 package mat64
 
 import (
-	"github.com/gonum/blas/cblas"
+	"github.com/gonum/blas/blas64"
+	"github.com/gonum/blas/cgo"
 )
 
 func init() {
-	Register(cblas.Blas{})
+	blas64.Use(cgo.Implementation{})
 }
--- a/mat64/dense.go
+++ b/mat64/dense.go
@@ -5,21 +5,8 @@
 package mat64
 
 import (
-	"github.com/gonum/blas"
-	"github.com/gonum/blas/goblas"
+	"github.com/gonum/blas/blas64"
 )
-
-// Set the default BLAS implementation.
-var blasEngine blas.Float64 = goblas.Blas{}
-
-func Register(b blas.Float64) {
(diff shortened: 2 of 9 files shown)


jonlawlor:
There are [60 downstream packages](http://godoc.org/github.com/gonum/matrix/mat64?importers) (!) that will be broken by this change:
* many are subpackages or forks of [golearn](http://godoc.org/github.com/sjwhitworth/golearn)
* [go-nets](https://github.com/antoine-lizee/go-nets/blob/master/network.go#L21) also uses it
* some are gonum (specifically, optimize.  stat uses it but looks ok)
* some are owned by @btracey

kortschak:
The only change required in those packages will be the deletion of the `mat64.Register()` call unless people are doing `RawMatrix` matrix construction (`RawMatrix()` calls will continue to work as previously unless people predeclare their vars). Also the changes at gonum/blas#83 already will have broken those packages.

The pain will be worthwhile in the long term.

btracey:
Nothing exported depends on BLAS except the call to Register. Register going away will break stuff, but the fix is either a deletion (if you were using goblas) or a one line fix (from cblas.Blas{} to cgo.Implementation{}). The removal of RawMatrix is potentially more damaging, but at least the types are identical. If you were using RawMatrix() and :=, possibly nothing will need to change. Otherwise, some names will need to change. This is unfortunate, but we're still too young an ecosystem to be unable to make breaking changes. This set of changes is how the blas organization should be, and hopefully we won't have to break this again.

btracey:
@sjwhitworth, @antoine-lizee, @dane-unltd, @jesand, @kse we're sorry for the break, but it's for the best long-term

jonlawlor:
I'll put together a notification for them tomorrow.

btracey:
LGTM. I just looked at the diff, so there may be unchanged comments, but if they got through comments are easy to fix.

btracey:
Okay, now we see how badly all of my code breaks ...

PR #67, opened 2015-01-10
Use Dsyrk in MulTrans when doing AA' or A'A

kortschak:


kortschak:
@jonlawlor @btracey

kortschak:
During previous discussions we considered the use of nil to indicate the same matrix, e.g. `m.MulTrans(a, true, nil, false)` being equivalent to `m.MulTrans(a, true, a, false)`. I now think that would be a mistake since it then requires people understand that special case.

btracey:
LGTM

jonlawlor:
I hope you don't mind, I added some benchmarks:

```
jonlawlor: ~/go/src/github.com/gonum/matrix/mat64 $ go test -bench MulTrans
OK: 33 passed
PASS
BenchmarkMulTransDense100Half	     500	   3095720 ns/op
BenchmarkMulTransDense100Tenth	     500	   3247291 ns/op
BenchmarkMulTransDense1000Half	       1	3084764617 ns/op
BenchmarkMulTransDense1000Tenth	       1	3057808541 ns/op
BenchmarkMulTransDense1000Hundredth	       1	3051772182 ns/op
BenchmarkMulTransDense1000Thousandth	       1	3081380968 ns/op
BenchmarkMulTransDenseSym100Half	    1000	   1733685 ns/op
BenchmarkMulTransDenseSym100Tenth	    1000	   1718710 ns/op
BenchmarkMulTransDenseSym1000Half	       1	1687840120 ns/op
BenchmarkMulTransDenseSym1000Tenth	       1	1718980158 ns/op
BenchmarkMulTransDenseSym1000Hundredth	       1	1692635500 ns/op
BenchmarkMulTransDenseSym1000Thousandth	       1	1694264557 ns/op
```
Looks great to me!

kortschak:
It's not guaranteed to be safe given the workflow I'm adopting (amend and force push), but the benchmarks LGTM.

PR #69, opened 2015-01-10
Fix incorrect test messages

kortschak:


jonlawlor:
LGTM.  Sorry about the MulTrans one.

PR #71, opened 2015-01-11
Use harmonised panic message in test

kortschak:
This depends on gonum/blas#85, review that first.


btracey:
Looks like this failed because some of the tests expect certain panic strings which this changes.

kortschak:
It will pass when gonum/blas#85 is merged.

btracey:
Oh, I see, thanks. 

LGTM when the time comes.

PR #77, opened 2015-01-13
Added Triangular type, At and Set methods, and tests

btracey:


mat64/triangular.go
```diff
@@ -0,0 +1,55 @@
...
+
+import (
+	"github.com/gonum/blas"
+	"github.com/gonum/blas/blas64"
+)
+
+var (
+	triangular *Triangular
+
+	_ Matrix  = triangular
```
sbinet: naive comment: isn't "cleaner" to just spell it out
`_ Matrix = (*Triangular)(nil)`

admittedly more verbose, but no "pollution" of the global scope.

mat64/index_bound_checks.go
```diff
@@ -28,10 +30,54 @@ func (m *Dense) Set(r, c int, v float64) {
...
+func (t *Triangular) set(r, c int, v float64) {
+	if r >= t.mat.N || r < 0 {
+		panic(ErrRowAccess)
+	}
+	if c >= t.mat.N || c < 0 {
+		panic(ErrColAccess)
+	}
+	if t.mat.Uplo == blas.Upper && r > c {
+		panic("mat64: triangular set out of bounds")
+	} else if t.mat.Uplo == blas.Lower && r < c {
```
fhs: Remove "else". It's idiomatic to write simple if statements. See [Effective Go](https://golang.org/doc/effective_go.html#if).

mat64/index_no_bound_checks.go
```diff
@@ -35,3 +37,47 @@ func (m *Dense) Set(r, c int, v float64) {
...
+func (t *Triangular) Set(r, c int, v float64) {
+	if r >= t.mat.N || r < 0 {
+		panic(ErrRowAccess)
+	}
+	if c >= t.mat.N || c < 0 {
+		panic(ErrColAccess)
+	}
+	if t.mat.Uplo == blas.Upper && r > c {
+		panic("mat64: triangular set out of bounds")
+	} else if t.mat.Uplo == blas.Lower && r < c {
```
fhs: ditto

mat64/index_bound_checks.go
```diff
@@ -28,10 +30,54 @@ func (m *Dense) Set(r, c int, v float64) {
...
+		}
+		return t.mat.Data[r*t.mat.Stride+c]
+	}
+	if r < c {
+		return 0
+	}
+	return t.mat.Data[r*t.mat.Stride+c]
+}
+
+// Set sets the element at row r and column c. Set panics if the
```
vladimir-ch: Unfinished comment

mat64/index_no_bound_checks.go
```diff
@@ -35,3 +37,47 @@ func (m *Dense) Set(r, c int, v float64) {
...
+		}
+		return t.mat.Data[r*t.mat.Stride+c]
+	}
+	if r < c {
+		return 0
+	}
+	return t.mat.Data[r*t.mat.Stride+c]
+}
+
+// Set sets the element at row r and column c. Set panics if the
```
vladimir-ch: Unfinished comment

mat64/triangular.go
btracey: I'm following the pattern that exists there now. We can change it if there's a better way, but consistency is better.

mat64/index_no_bound_checks.go
btracey: Done.

mat64/index_bound_checks.go
btracey: Done.

mat64/index_bound_checks.go
btracey: Done.

mat64/index_no_bound_checks.go
btracey: Done.

btracey:
Fixed comments and added a comment to NewTriangular about the underlying data. PTAL

jonlawlor:
Is QR going to be changed to:
```go
type QRFactor struct {
	QR    *Triangular
	rDiag []float64
}
```

btracey:
I would vote it gets changed in some form. You can see my thoughts on Eigen at Issue #79 , and I have similar thoughts for Cholesky and QRFactor. I'm not a dictator though, we'll need to discuss how we want these to look. Getting the basic types in will make it easier to think about.

btracey:
We should also consider if we want the Upper and Lower constants, or if we would rather have a boolean where true == Upper. There are only two possibilities, it would eliminate a panic possibility, and a boolean would match what we've chosen for Trans. However, it seemed to be that Upper and Lower read better, and it's not clear a priori which one should be true.

jonlawlor:
You can type Upper and Lower to boolean and callers can just use true and false.  http://play.golang.org/p/9FKN8h4SbT

btracey:
Removed TriType and made the field a Boolean. We can revisit if we want to create a typed Boolean, but this is in line with the current Trans behavior.

jonlawlor:
That bool change looks good to me; but I haven't reviewed all of the other changes in depth.

btracey:
Pulled from master

btracey:
Just made a new PR ( #94) that incorporates some changes from the symmetric discussion.

PR #78, opened 2015-01-13
mat64: implement Binary(Un)Marshaler

sbinet:
implement `encoding.BinaryMarshaler` and `encoding.BinaryUnmarshaler` for:
- `mat64.Dense`
- `mat64.Eigen`
- `mat64.LQ`
- `mat64.LU`
- `mat64.QR`
- `mat64.SVD`
- `mat64.Triangular`
- `mat64.Vec`

add tests.

### DIFF
--- a/mat64/dense.go
+++ b/mat64/dense.go
@@ -5,6 +5,9 @@
 package mat64
 
 import (
+	"bytes"
+	"encoding/binary"
+
 	"github.com/gonum/blas/blas64"
 )
 
@@ -496,4 +499,74 @@ func (m *Dense) Augment(a, b Matrix) {
 	m.Copy(a)
 	w := m.View(0, ac, br, bc).(*Dense)
 	w.Copy(b)
+}
+
+// MarshalBinary encodes the receiver into a binary form and returns the result.
+//
+// Dense is little-endian encoded as follows:
+//  0 -  8  number of rows    (int64)
+//  8 - 16  number of columns (int64)
+// 16 - ..  matrix data elements (float64)
+//          [0,0] [0,1] ... [0,ncols-1]
+//          [1,0] [1,1] ... [1,ncols-1]
+//          ...
+//          [nrows-1,0] ... [nrows-1,ncols-1]
+func (m Dense) MarshalBinary() ([]byte, error) {
+	buf := bytes.NewBuffer(make([]byte, 0, m.mat.Rows*m.mat.Cols*sizeFloat64+2*sizeInt64))
+	err := binary.Write(buf, defaultEndian, int64(m.mat.Rows))
+	if err != nil {
+		return nil, err
+	}
+	err = binary.Write(buf, defaultEndian, int64(m.mat.Cols))
+	if err != nil {
+		return nil, err
+	}
+
+	for i := 0; i < m.mat.Rows; i++ {
+		for _, v := range m.rowView(i) {
+			err = binary.Write(buf, defaultEndian, v)
+			if err != nil {
+				return nil, err
+			}
+		}
+	}
+	return buf.Bytes(), err
+}
+
+// UnmarshalBinary decodes the binary form into the receiver.
+// It panics if the receiver is a non-zero Dense matrix.
+//
+// See MarshalBinary for the on-disk layout.
+func (m *Dense) UnmarshalBinary(data []byte) error {
+	if !m.isZero() {
+		panic("mat64: unmarshal into non-zero matrix")
+	}
+
+	buf := bytes.NewReader(data)
+	var rows int64
+	err := binary.Read(buf, defaultEndian, &rows)
+	if err != nil {
+		return err
+	}
+	var cols int64
+	err = binary.Read(buf, defaultEndian, &cols)
+	if err != nil {
+		return err
+	}
+
+	m.mat.Rows = int(rows)
+	m.mat.Cols = int(cols)
+	m.mat.Stride = int(cols)
+	m.capRows = int(rows)
+	m.capCols = int(cols)
+	m.mat.Data = use(m.mat.Data, m.mat.Rows*m.mat.Cols)
+
+	for i := range m.mat.Data {
+		err = binary.Read(buf, defaultEndian, &m.mat.Data[i])
+		if err != nil {
+			return err
+		}
+	}
+
+	return err
 }
--- a/mat64/io.go
+++ b/mat64/io.go
@@ -1,0 +1,18 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package mat64
+
+import (
+	"encoding/binary"
+)
+
+var (
+	littleEndian  = binary.LittleEndian
+	bigEndian     = binary.BigEndian
+	defaultEndian = littleEndian
+
+	sizeInt64   = binary.Size(int64(0))
+	sizeFloat64 = binary.Size(float64(0))
+)
--- a/mat64/io_test.go
+++ b/mat64/io_test.go
@@ -1,0 +1,35 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package mat64
+
+import (
+	"gopkg.in/check.v1"
+)
+
+func (s *S) TestDenseRW(c *check.C) {
+	for i, test := range []*Dense{
+		NewDense(0, 0, []float64{}),
+		NewDense(2, 2, []float64{1, 2, 3, 4}),
+		NewDense(2, 3, []float64{1, 2, 3, 4, 5, 6}),
+		NewDense(3, 2, []float64{1, 2, 3, 4, 5, 6}),
+		NewDense(3, 3, []float64{1, 2, 3, 4, 5, 6, 7, 8, 9}),
+		NewDense(3, 3, []float64{1, 2, 3, 4, 5, 6, 7, 8, 9}).View(0, 0, 2, 2).(*Dense),
+		NewDense(3, 3, []float64{1, 2, 3, 4, 5, 6, 7, 8, 9}).View(1, 1, 2, 2).(*Dense),
+		NewDense(3, 3, []float64{1, 2, 3, 4, 5, 6, 7, 8, 9}).View(0, 1, 3, 2).(*Dense),
+	} {
+		buf, err := test.MarshalBinary()
+		c.Check(err, check.Equals, nil, check.Commentf("error encoding test #%d: %v\n", i, err))
+
+		nrows, ncols := test.Dims()
+		sz := nrows*ncols*sizeFloat64 + 2*sizeInt64
+		c.Check(len(buf), check.Equals, sz, check.Commentf("encoded size test #%d: want=%d got=%d\n", i, sz, len(buf)))
+
+		var got Dense
+		err = got.UnmarshalBinary(buf)
+		c.Check(err, check.Equals, nil, check.Commentf("error decoding test #%d: %v\n", i, err))
+
+		c.Check(got.Equals(test), check.Equals, true, check.Commentf("r/w test #%d failed\nwant=%#v\n got=%#v\n", i, test, &got))
+	}
+}


sbinet:
needs #77

mat64/dense.go
```diff
@@ -5,7 +5,10 @@
 package mat64
 
 import (
+	"bytes"
+
 	"github.com/gonum/blas/blas64"
+	"github.com/gonuts/binary"
```
btracey: I'd really prefer to not add a downstream dependency like this . At the very least we should vendor it. See note below, however.

mat64/dense.go
```diff
@@ -417,3 +420,25 @@ func (m *Dense) Augment(a, b Matrix) {
 	w := m.View(0, ac, br, bc).(*Dense)
 	w.Copy(b)
 }
+
+// MarshalBinary encodes the receiver into a binary form and returns the result.
+func (m Dense) MarshalBinary() ([]byte, error) {
+	buf := new(bytes.Buffer)
+	enc := binary.NewEncoder(buf)
```
btracey: Using this package seems like a bad idea to me. It supports variable-length encoding, but does so by adding the size. Here's how I think it should work. Dense (etc) implement GobEncoder, which you use to get compact representations of your matrix, if, for example, your matrix is sparse or something. Then, it also implements MarshalBinary, where it first encodes all of size parameters (as int64 rather than int), and then encodes the floats at fixed-width (8 bytes). This does two things. First, it means that this is a portable binary representation with a very simple spec. There are three int64 (row, col, Stride), and then row*Stride + col  IEEE float64 numbers each at 8 bytes. Very easy to decode into other languages/programs. Secondly, because the sizes are known a-priori, MarshalBinary can be coded in parallel, which can have a big speedup for large matrices. 

In short: GobEncoder to have a compact representation, MarshalBinary to have parallel I/O and portable format.

mat64/eigen.go
```diff
@@ -817,3 +820,49 @@ func (f EigenFactors) D() *Dense {
 	}
 	return dm
 }
+
+// MarshalBinary encodes the receiver into a binary form and returns the result.
+func (f EigenFactors) MarshalBinary() ([]byte, error) {
```
btracey: I see the utility of having marshalers for Dense, but do we really need the code for these other values (not just this, but all the supplemental structs)? I think we should keep the code minimal until there's a demonstrated need to marshal a particular struct. This is especially true as some of these structs may change, see for example Issue #79

kortschak:
This does way too much. We really just want to support *Dense IMO. The support types are overkill.

mat64/dense.go
kortschak: gonuts/binary is excessive - it provides support for a variety of types where we only need []float64. All we need to do is write an initial int64 and then the following data (in whatever format). It fosters a leave it to someoe else attitude that (below) gives us what I think is the wrong behaviour. An encode helper function would do the necessary prelude and have the correct behaviour.

mat64/dense.go
```diff
@@ -417,3 +420,25 @@ func (m *Dense) Augment(a, b Matrix) {
...
+	err := enc.Encode(&m.mat)
```
kortschak: This is not what we want to do.

Consider the case where m is a 10x10 view into a 1000x1000 matrix near the upper left. This would write almost the entire data slice of the 1000x1000, but would not be able to be used to recover the entire 1000x1000 matrix. We need to just write the 100 elements represented by m.

mat64/dense.go
kortschak: We can get away with just row, col and data. There is no need for stride in the serialised data.

mat64/dense.go
sbinet: @btracey : gonuts/binary is mine. I can easily vendor it under, say, gonum/internal/binary

sbinet: @kortschak : I started with a minimal writeBin/writeLen/readBin/readLen (as the commit logs show) then went to gonuts/binary for support of Triangular, SVD, ...

mat64/dense.go
sbinet: @btracey my rational for using an `encoding/binary` (or `gonuts/binary`) based solution is cross-language support.
`encoding/gob.GobEncoding` (and its `GobDecoding` sidekick) is obviously `go` oriented.

`varint` is rather easily implementable in various language (even `python`)

but I do recognize the usefulness of concurrent decoding capabilities that not using varint (and thus `int64`) buys us.

mat64/dense.go
sbinet: this can (and should) be tackled at the `blas64.General.MarshalBinary` layer.

kortschak: blas64 is a thin wrapper around blas. I don't think it is the job of that package to handle this. There are no methods on any of the types in that package.

mat64/eigen.go
sbinet: my rationale for adding them was that they might be intermediate steps of an analysis and one might want to restart from these intermediate states w/o having to go through the whole (possibly quite expensive) computation.
as a user, it might become a bit tedious to figure out that some critical fields are not exported and thus silently discarded from the compiler-provided default serialization functions.
as a library implementer, providing those seemed quite interesting.

but, ok. I can remove them, if the general consensus tips that way. :)

mat64/dense.go
kortschak: But it give use the wrong behaviour for *Dense and we don't want to support SVD etc. The writing that's done for Triangular is just a few lines.

mat64/eigen.go
kortschak: It's easy for clients to extract the relevant data that might be reused later and serialise that.

sbinet: @kortschak  extracting relevant data is one thing, re-synthesizing it (w/o too much of hysteresis) is another can of worms (especially when you don't have direct access to all data fields)

kortschak: The changes that @btracey has suggested to EigenFactors exports all the relevant fields.

mat64/dense.go
btracey: I'm with Dan. I too am tempted to add methods, but it's better to keep the complexity in mat64.

mat64/dense.go
btracey: BinaryMarshaler and BinaryUnmarshaler should be the simplest thing, which is to take the matrix, and dump its data in binary form (int64, int64, float64, float64 ...). This is also possibly the most cross language encoding, as it is the simplest to encode and decode. 

We can support multiple kinds of encoding, for example we could add GobEncoder and VarintEncoder, but BinaryMarshaler should be the most basic version.

sbinet:
PTAL.
- removed everything but the marshaler for `Dense`
- implemented "stride suppression"

(happy to squash everything into one big CL)

mat64/dense.go
```diff
@@ -427,3 +429,63 @@ func (m *Dense) Augment(a, b Matrix) {
...
+		return nil, err
+	}
+
+	err = writeBin(buf, int64(m.mat.Cols))
+	if err != nil {
+		return nil, err
+	}
+
+	for i := 0; i < m.mat.Rows; i++ {
+		// note we don't write out the length of the slice as
```
kortschak: s/note w/W/
s/m.mat.Stride/m.mat.Cols/

mat64/dense.go
```diff
@@ -427,3 +429,63 @@ func (m *Dense) Augment(a, b Matrix) {
...
+		return err
+	}
+
+	m.mat.Rows = int(rows)
+	m.mat.Cols = int(cols)
+	m.mat.Stride = int(cols)
+	m.mat.Data = make([]float64, 0, m.mat.Rows*m.mat.Cols)
+
+	for i := 0; i < m.mat.Rows; i++ {
+		// note we didn't write out the length of the slice as
```
kortschak: Matching above.

kortschak:
LGTM after addressing comments. Wait for at least one other review.

Please do squash this.

sbinet:
PTAL
(squashed)

mat64/dense.go
sbinet: done.

mat64/dense.go
sbinet: done.

mat64/io_test.go
```diff
@@ -0,0 +1,42 @@
...
+		NewDense(3, 3, []float64{1, 2, 3, 4, 5, 6, 7, 8, 9}).View(0, 0, 2, 2),
+		NewDense(3, 3, []float64{1, 2, 3, 4, 5, 6, 7, 8, 9}).View(1, 1, 2, 2),
+		NewDense(3, 3, []float64{1, 2, 3, 4, 5, 6, 7, 8, 9}).View(0, 1, 3, 2),
+	} {
+		buf := new(bytes.Buffer)
+		enc := gob.NewEncoder(buf)
+		err := enc.Encode(test)
+		if err != nil {
+			t.Fatalf("error encoding test #%d: %v\n", i, err)
+		}
```
btracey: Could you please add a test that the returned bytes is the correct length? It should be rows*cols*8 + 16 bytes.

mat64/dense.go
```diff
@@ -427,3 +429,63 @@ func (m *Dense) Augment(a, b Matrix) {
...
+		return nil, err
+	}
+
+	err = writeBin(buf, int64(m.mat.Cols))
+	if err != nil {
+		return nil, err
+	}
+
+	for i := 0; i < m.mat.Rows; i++ {
+		// We don't write out the length of the slice as
```
btracey: I found this comment confusing. I would prefer you get rid of this, and before the for loop write
````
// Write all of the elements in condensed form  
````
or something like that.

File: mat64/io.go [hunk 9]
```diff
@@ -0,0 +1,60 @@
...
+	littleEndian  = binary.LittleEndian
+	bigEndian     = binary.BigEndian
+	defaultEndian = littleEndian
+)
+
+func writeBin(w io.Writer, data interface{}) error {
+	return binary.Write(w, defaultEndian, data)
+}
+
+func writeInt(w io.Writer, n int) error {
```
btracey: This function is never used. Delete it.

mat64/io.go
```diff
@@ -0,0 +1,60 @@
...
+	"io"
+)
+
+var (
+	littleEndian  = binary.LittleEndian
+	bigEndian     = binary.BigEndian
+	defaultEndian = littleEndian
+)
+
+func writeBin(w io.Writer, data interface{}) error {
```
btracey: Remove this function. It's one line, it's used three times, and the only thing it saves is typing "defaultEndian". By using binary.Write directly, it's clear that nothing tricky is happening.

mat64/io.go
```diff
@@ -0,0 +1,60 @@
...
+	return binary.Write(w, defaultEndian, data)
+}
+
+func writeInt(w io.Writer, n int) error {
+	return writeBin(w, int64(n))
+}
+
+func writeSlice(w io.Writer, data []float64) error {
+	var err error
+
```
btracey: Remove \n

mat64/io.go
```diff
@@ -0,0 +1,60 @@
...
+	for _, v := range data {
+		err = writeBin(w, v)
+		if err != nil {
+			return err
+		}
+	}
+
```
btracey: Remove \n

mat64/dense.go
```diff
@@ -427,3 +429,63 @@ func (m *Dense) Augment(a, b Matrix) {
...
+		return err
+	}
+
+	m.mat.Rows = int(rows)
+	m.mat.Cols = int(cols)
+	m.mat.Stride = int(cols)
+	m.mat.Data = make([]float64, 0, m.mat.Rows*m.mat.Cols)
+
+	for i := 0; i < m.mat.Rows; i++ {
+		// We didn't write out the length of the slice as
```
btracey: Same comment as with Write
````
The elements are stored as a compact array, so read them in order
````

File: mat64/dense.go [hunk 14]
```diff
@@ -427,3 +429,63 @@ func (m *Dense) Augment(a, b Matrix) {
...
+		// it should be m.mat.Cols.
+		slice, err := readSlice(buf, m.mat.Cols)
```
btracey: Don't use readSlice. It allocates and doesn't make the code much cleaner. Just create Data as a full-length slice and do
````
// Compactness comment
for i := 0; i < m.mat.Rows; i++ {
    tmp := data[i*m.mat.Stride:i*m.mat.Stride + m.mat.Cols]
    for j := 0; j < m.mat.Cols; j++ {
        err := binary.Read(buf, defaultEndian, &tmp[j])
         if err != nil{
             return err
        }
}
````
Now the reader can see everything that's happening, and the code is basically as efficient as it can be.

File: mat64/io.go [hunk 15]
```diff
@@ -0,0 +1,60 @@
...
+	return err
+}
+
+func readBin(r io.Reader, data interface{}) error {
+	return binary.Read(r, defaultEndian, data)
+}
+
+func readInt(r io.Reader) (int, error) {
```
btracey: Unused. Delete.

mat64/io.go
```diff
@@ -0,0 +1,60 @@
...
+		err = writeBin(w, v)
+		if err != nil {
+			return err
+		}
+	}
+
+	return err
+}
+
+func readBin(r io.Reader, data interface{}) error {
```
btracey: See comment on writeBin. One line function that doesn't help much. Just put in place in the main function.

mat64/io.go
```diff
@@ -0,0 +1,60 @@
...
+	return binary.Read(r, defaultEndian, data)
+}
+
+func readInt(r io.Reader) (int, error) {
+	var n int64
+	err := readBin(r, &n)
+	return int(n), err
+}
+
+func readSlice(r io.Reader, n int) ([]float64, error) {
```
btracey: This can be deleted too. See loop comment.

mat64/io.go
btracey: Actually, given that I think it's clearer with readSlice removed, writeSlice is probably better removed instead.

mat64/dense.go
```diff
@@ -427,3 +429,63 @@ func (m *Dense) Augment(a, b Matrix) {
 	w := m.View(0, ac, br, bc).(*Dense)
 	w.Copy(b)
 }
+
+// MarshalBinary encodes the receiver into a binary form and returns the result.
```
btracey: Since we intend this to be cross-language and cross-program, I think a comment documenting how the matrix is marshaled is in order.

mat64/dense.go
```diff
@@ -427,3 +429,63 @@ func (m *Dense) Augment(a, b Matrix) {
...
+	var cols int64
+	err = readBin(buf, &cols)
+	if err != nil {
+		return err
+	}
+
+	m.mat.Rows = int(rows)
+	m.mat.Cols = int(cols)
+	m.mat.Stride = int(cols)
+	m.mat.Data = make([]float64, 0, m.mat.Rows*m.mat.Cols)
```
kortschak: Now that #82 is in, you need to set m.capRows = rows and m.capCols = cols.

You'll need to rebase to see those.

mat64/io_test.go
```diff
@@ -0,0 +1,42 @@
...
+		NewDense(3, 2, []float64{1, 2, 3, 4, 5, 6}),
+		NewDense(3, 3, []float64{1, 2, 3, 4, 5, 6, 7, 8, 9}),
+		NewDense(3, 3, []float64{1, 2, 3, 4, 5, 6, 7, 8, 9}).View(0, 0, 2, 2),
+		NewDense(3, 3, []float64{1, 2, 3, 4, 5, 6, 7, 8, 9}).View(1, 1, 2, 2),
+		NewDense(3, 3, []float64{1, 2, 3, 4, 5, 6, 7, 8, 9}).View(0, 1, 3, 2),
+	} {
+		buf := new(bytes.Buffer)
+		enc := gob.NewEncoder(buf)
+		err := enc.Encode(test)
+		if err != nil {
```
kortschak: Apologies, tests in mat64 should be done using gocheck until we revise that. Otherwise we have two test output formats.

mat64/dense.go
sbinet: done

sbinet:
PTAL.
- handle `m.capXYZ`
- document on-disk data layout
- migrate test to `gopkg.in/check.v1`

mat64/dense.go
```diff
@@ -497,3 +500,71 @@ func (m *Dense) Augment(a, b Matrix) {
...
+// Dense is little-endian encoded as follows:
+//  0 -  8  number of rows    (int64)
+//  8 - 16  number of columns (int64)
+// 16 - ..  matrix data elements (float64)
+//          [0,0] [0,1] ... [0,ncols-1]
+//          [1,0] [1,1] ... [1,ncols-1]
+//          ...
+//          [nrows-1,0] ... [nrows-1,ncols-1]
+func (m Dense) MarshalBinary() ([]byte, error) {
+	var err error
```
btracey: Remove and use := at the first one.

mat64/dense.go
```diff
@@ -497,3 +500,71 @@ func (m *Dense) Augment(a, b Matrix) {
...
+	buf := bytes.NewBuffer(make([]byte, 0, m.mat.Rows*m.mat.Cols*8+16))
+	err = binary.Write(buf, defaultEndian, int64(m.mat.Rows))
+	if err != nil {
+		return nil, err
+	}
+
```
btracey: Remove this newline so that the separation between encoding {rows,cols} and the rest is clearer

mat64/dense.go
```diff
@@ -497,3 +500,71 @@ func (m *Dense) Augment(a, b Matrix) {
...
+//
+// See MarshalBinary for the on-disk layout.
+func (m *Dense) UnmarshalBinary(data []byte) error {
+	buf := bytes.NewReader(data)
+	var rows int64
+	err := binary.Read(buf, defaultEndian, &rows)
+	if err != nil {
+		return err
+	}
+
```
btracey: rm \n (like above)

mat64/dense.go
```diff
@@ -497,3 +500,71 @@ func (m *Dense) Augment(a, b Matrix) {
...
+		}
+	}
+	return buf.Bytes(), err
+}
+
+// UnmarshalBinary decodes the binary form into the receiver.
+//
+// See MarshalBinary for the on-disk layout.
+func (m *Dense) UnmarshalBinary(data []byte) error {
+	buf := bytes.NewReader(data)
```
btracey: This should have an "isZero" check and panic otherwise.

mat64/dense.go
```diff
@@ -497,3 +500,71 @@ func (m *Dense) Augment(a, b Matrix) {
...
+	if err != nil {
+		return err
+	}
+
+	m.mat.Rows = int(rows)
+	m.mat.Cols = int(cols)
+	m.mat.Stride = int(cols)
+	m.capRows = int(rows)
+	m.capCols = int(cols)
+	m.mat.Data = make([]float64, m.mat.Rows*m.mat.Cols)
```
btracey: Instead of this make, use zero so that data can be reused if it exists

sbinet: hum... I am not sure `zero` quite cuts it. it uses the `copy` builtin, which won't work in case `cap` is too small.
rather, one should check `cap(m.mat.Data)` for enough space, if there isn't, we'll need to re-allocate: better do that in one big chunk rather than discovering it too late in the game (ie: a big `make`)
otherwise, if `len(m.mat.Data)` is too small, `append` in one big chunk.
finally, `m.mat.Data = m.mat.Data[: m.mat.Rows*m.mat.Cols]` (and perhaps also set the cap?)

besides, all the data will be overwritten by the for-loop and `binary.Read` so the for-loop of `zero` is a bit superfluous if you ask me.

so:

```go
sz := m.mat.Rows * m.mat.Cols
if cap(m.mat.Data) < sz {
   m.mat.Data = make([]float64, sz)
}
if len(m.mat.Data) < sz {
  m.mat.Data = append(m.mat.Data, make([]float64, sz-len(m.mat.Data))...)
}
if len(m.mat.Data) > sz {
  m.mat.Data = m.mat.Data[:sz]
  // or, even:
  // m.mat.Data = m.mat.Data[:sz:sz]
}
```

to be completely honest, I am not sure all of these (premature?) optimizations are worth it.

in 90% of cases, `UnmarshalBinary` will be used on just-created `Dense` matrices... won't it?

mat64/dense.go
sbinet: why?
`UnmarshalBinary` is meant to overwrite all the fields and state of `Dense`.
and its use is quite obvious. there won't be any hysteresis effect or weird state left in `Dense` after it's been called.

mat64/dense.go
sbinet: ok.

mat64/dense.go
sbinet: ok

mat64/dense.go
sbinet: ok

mat64/dense.go
btracey: Sorry, there shouldn't have been a space. I meant to point you to the already-implemented useZeroed (see Grow). AS you point out, this overwrites all of the elements, use() is good. 
Instead of
````
m.mat.Data = make([]float64, m.mat.Rows*m.mat.Cols)
````
Instead do
````
m.mat.Data = use(m.mat.Data, m.mat.Rows * m.mat.Cols)
````

File: mat64/dense.go (hunk 24 above)
Comment (btracey) [inline mat64/dense.go]: Exactly because if overwrites all the data in a matrix. It should only be called on a newly-created Dense or one that has been Reset() so that it can be used in arbitrary ways. An existing m x n matrix should not be silently be overwritten. Calling UnmarshalBinary on a Dense that contains data is incorrect; if the data doesn't matter anymore, you should call Reset() first.

File: mat64/dense.go (hunk 25 above)
Comment (sbinet) [inline mat64/dense.go]: done

File: mat64/dense.go (hunk 24 above)
Comment (sbinet) [inline mat64/dense.go]: hum... ok.
I'd err on less nany-ing, but, fine :)

done.

**Comment by sbinet** [comment]
PTAL.

**Comment by btracey** [comment]
LGTM. Confirm @kortschak ?

File: mat64/dense.go (hunk 24 above)
Comment (kortschak) [inline mat64/dense.go]: I'm a little surprised we are being so unforgiving here, though it is consistent with the way we behave in other cases. @sbinet, we are essentially proxying type through the dims and allowing the zero-dims state be dynamic. Where it is surprising is that this is not how it is done pretty much anywhere else. Because of this, the behaviour should be explicitly documented in the doc comment.

For an (entertaining) explanation of why we are so brutal, when you have 45min to spare, watch this: https://www.youtube.com/watch?v=lrm8iEMQZNw

File: mat64/dense.go [hunk 26]
```diff
@@ -497,3 +500,72 @@ func (m *Dense) Augment(a, b Matrix) {
...
+// Dense is little-endian encoded as follows:
+//  0 -  8  number of rows    (int64)
+//  8 - 16  number of columns (int64)
+// 16 - ..  matrix data elements (float64)
+//          [0,0] [0,1] ... [0,ncols-1]
+//          [1,0] [1,1] ... [1,ncols-1]
+//          ...
+//          [nrows-1,0] ... [nrows-1,ncols-1]
+func (m Dense) MarshalBinary() ([]byte, error) {
+	buf := bytes.NewBuffer(make([]byte, 0, m.mat.Rows*m.mat.Cols*8+16))
```
kortschak: s/8/binary.Size(float64(0))/
s/16/2*binary.Size(int64(0))/

or preferably add a var block defining sizeInt64 and sizeFloat64 based on above that are refered to here.

kortschak:
LGTM after comments. Thank you.

sbinet:
PTAL

kortschak:
LGTM

sbinet:
yay! :)
(just so there is no quid pro quo: I don't have enough karma to merge it myself)

kortschak:
As an addendum, if the unmarshaling into a non-zero matrix becomes an issue, we can revisit this.

Issue #73, opened 2015-01-12
mat64: use() and *Dense.View() don't play well together


kortschak:
Consider the following code:

```go
package main

import (
	"fmt"

	"github.com/gonum/matrix/mat64"
)

type fm struct {
	mat64.Matrix
}

func (m fm) Format(fs fmt.State, c rune) {
	if c == 'v' && fs.Flag('#') {
		fmt.Fprintf(fs, "%#v", m.Matrix)
		return
	}
	mat64.Format(m.Matrix, 0, '.', fs, c)
}

func main() {
	a := mat64.NewDense(4, 4, []float64{
		1, 2, 3, 4,
		5, 6, 7, 8,
		9, 10, 11, 12,
		13, 14, 15, 15,
	})
	v := a.View(1, 1, 1, 1).(*mat64.Dense)
	v.Reset()

	b := mat64.NewDense(3, 3, []float64{
		1, 2, 3,
		4, 5, 6,
		7, 8, 9,
	})

	fmt.Printf("a' before =\n%v\n", fm{a})

	v.Add(b, b)

	fmt.Printf("v' =\n%v\n", fm{v})

	fmt.Printf("a' after =\n%v\n", fm{a})
}
```

```
$ go run useView.go 
a' before =
⎡ 1   2   3   4⎤
⎢ 5   6   7   8⎥
⎢ 9  10  11  12⎥
⎣13  14  15  15⎦
v' =
⎡ 2   4   6⎤
⎢ 8  10  12⎥
⎣14  16  18⎦
a' after =
⎡ 1   2   3   4⎤
⎢ 5   2   4   6⎥
⎢ 8  10  12  14⎥
⎣16  18  15  15⎦
```

We would expect that elements of `a` outside the view `v` would be unaffected by the add operation. However, the combination of View, Reset and use (internally) result in reuse of those elements illegitimately.

There are five obvious solutions, though the problem is subtle.

1. Document that they don't interact well and deal with queries on gonum-dev and issues here.
2. Replace use with make and lose slice reuse.
3. Make Reset re-make the slice (essentially the same as 2.).
4. Increase the logic complexity of use as described below.
5. Increase the complexity of the *Dense struct to include 2 dimensions of cap as described in the tables proposal by @btracey. (Left as an exercise for the reader).

```go
func use(m blas64.General, i, j, r, c, s int) blas64.General {
	var data []float64
	if s != m.Stride || r > m.Rows || c > m.Cols {
		data = make([]float64, r*c)
[...]

PR #81, opened 2015-01-13
Make Reset clearly a destructive action

kortschak:
If a user calls Reset on a *Dense there are potential implications on views, so document that. Also since we are now free to be brutal, don't bother keeping the Stride intact and use it as the flag for zero state.

Updates #73.

### DIFF
--- a/mat64/dense.go
+++ b/mat64/dense.go
@@ -49,6 +49,8 @@ var (
 
 	_ RawMatrixSetter = matrix
 	_ RawMatrixer     = matrix
+
+	_ Reseter = matrix
 )
 
 type Dense struct {
@@ -82,7 +84,9 @@ func (m *Dense) SetRawMatrix(b blas64.General) { m.mat = b }
 func (m *Dense) RawMatrix() blas64.General { return m.mat }
 
 func (m *Dense) isZero() bool {
-	return m.mat.Cols == 0 || m.mat.Rows == 0
+	// It must be the case that m.Dims() returns
+	// zeros in this case. See comment in Reset().
+	return m.mat.Stride == 0
 }
 
 func (m *Dense) Dims() (r, c int) { return m.mat.Rows, m.mat.Cols }
@@ -152,6 +156,10 @@ func (m *Dense) rowView(r int) []float64 {
 }
 
 func (m *Dense) View(i, j, r, c int) Matrix {
+	mr, mc := m.Dims()
+	if i < 0 || i >= mr || j < 0 || j >= mc || r <= 0 || i+r > mr || c <= 0 || j+c > mc {
+		panic(ErrIndexOutOfRange)
+	}
 	t := *m
 	t.mat.Data = t.mat.Data[i*t.mat.Stride+j : (i+r-1)*t.mat.Stride+(j+c)]
 	t.mat.Rows = r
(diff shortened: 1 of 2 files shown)


btracey:
The doc comment also needs to change. Right now it says " zero value for the concrete type of the matrix". This is misleading, as the zero value of Dense is:

````
Dense {
     blas64.General {
          Rows: 0,
          Cols: 0,
          Stride: 0,
         Data: nil,
    }
}
````

but the actual test is if stride is zero.

kortschak:
Can you suggest wording? We don't expose the tests we make here or the implementation details that are relevant, so it's not entirely clear how to describe this.

btracey:
Can we just say "if it's a zero-size matrix"? The only way to explicitly create a zero-size matrix is to do so with NewDense(0,0, nil), which will have a zero stride, or to use Reset() which documents its own purpose.

kortschak:
But that's not true either. I guess if we require that Stride is never set to zero unless dims are as well then it is a contract we can keep.

kortschak:
PTAL

btracey:
I like the wording, but I just realized an edge case. If I do m := a.View(i, j, 0, 0), will m be overwritten on use? The standard coding for View would keep Stride nonzero, and so it will panic. I think this behavior is reasonable as it works nicely with slice slicing and with Grow, and otherwise there has to be an exception in View. It does make the documentation more complicated.

How about: "dimensions for the result or a zero-strided matrix. Zero-strided matrices will usually occur as the zero-value of a matrix or by calling Reset(). In this latter case, matrix data ... "

It's more nuanced than I'd like, but it's a nuanced situation. Fortunately the code behavior is clear even if the precise English is not.

kortschak:
At the moment a.View(i, j, 0, 0) panics. We can leave that. What I was thinking is that when caps are added in, if you want to work into a view you need to grow or view it to the correct size. This solve many of the issues.

kortschak:
I don't want to refer to strides at all in the package doc.

btracey:
I keep forgetting that View doesn't work down to 0, 0. The behavior is clear anyway with a quick look at the code. LGTM

kortschak:
I have fixed both the documentation of Viewer to be explicit there and the implementation to match the docs as they were written - we were not doing what we said and say.

PTAL

btracey:
LGTM

I'm leaving this comment for future readers, it's a description of my internal argument that in the end agrees with the current behavior, but I think we also need to special case Grow a bit.

Conceptually there's nothing wrong with View being allow to go to zero size, and not allowing it doesn't mesh with slicing, which can go to zero size. There's also a possible use for such a view: preallocating a matrix to store values. If the values are coming in on a channel, you may not know when the first value will come in. If View doesn't work down to zero then you have to special case the code for the first size.

This is not correct, because it is easy to preallocate a zero-sized matrix with the data there. 
m := mat64.NewDense(0, 0, make([]float64, m*n))
When the first value comes in, you can then use m.Grow(1,n)  (or whatever your unit size is). This requires that Grow is also flexible for size expansion when the value is zero, which does mesh with the behavior in the rest of the package.

kortschak:
Perhaps we should have a further discussion to flesh this out. I'll merge, but we can leave the zero dims view option on the table for later.

PR #82, opened 2015-01-14
Add Grow method and Grower interface

### LINKED ISSUE #74: Add capacities to Dense
There are a couple of problems that arise from having View but not having capacities (see Issue #73 and Issue #65). Additionally, adding capacities will support matrix growing (https://groups.google.com/forum/#!topic/gonum-dev/WEeFgNtF9yk) without additional complexities.

kortschak:
Fixes #74.

### DIFF
--- a/mat64/dense.go
+++ b/mat64/dense.go
@@ -19,6 +19,7 @@ var (
 	_ Cloner    = matrix
 	_ Viewer    = matrix
 	_ RowViewer = matrix
+	_ Grower    = matrix
 
 	_ Adder     = matrix
 	_ Suber     = matrix
@@ -55,6 +56,8 @@ var (
 
 type Dense struct {
 	mat blas64.General
+
+	capRows, capCols int
 }
 
 func NewDense(r, c int, mat []float64) *Dense {
@@ -64,12 +67,16 @@ func NewDense(r, c int, mat []float64) *Dense {
 	if mat == nil {
 		mat = make([]float64, r*c)
 	}
-	return &Dense{blas64.General{
-		Rows:   r,
-		Cols:   c,
-		Stride: c,
-		Data:   mat,
-	}}
+	return &Dense{
+		mat: blas64.General{
+			Rows:   r,
+			Cols:   c,
+			Stride: c,
+			Data:   mat,
+		},
+		capRows: r,
+		capCols: c,
+	}
 }
 
 // DenseCopyOf returns a newly allocated copy of the elements of a.
@@ -79,7 +86,10 @@ func DenseCopyOf(a Matrix) *Dense {
 	return d
 }
 
-func (m *Dense) SetRawMatrix(b blas64.General) { m.mat = b }
+func (m *Dense) SetRawMatrix(b blas64.General) {
+	m.capRows, m.capCols = b.Rows, b.Cols
+	m.mat = b
+}
 
 func (m *Dense) RawMatrix() blas64.General { return m.mat }
 
@@ -90,6 +100,8 @@ func (m *Dense) isZero() bool {
 }
 
 func (m *Dense) Dims() (r, c int) { return m.mat.Rows, m.mat.Cols }
+
+func (m *Dense) Caps() (r, c int) { return m.capRows, m.capCols }
 
 func (m *Dense) Col(dst []float64, j int) []float64 {
 	if j >= m.mat.Cols || j < 0 {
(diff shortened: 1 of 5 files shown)


mat64/dense.go
```diff
@@ -164,13 +175,53 @@ func (m *Dense) View(i, j, r, c int) Matrix {
...
+	switch {
+	case m.mat.Rows == 0 || m.mat.Cols == 0:
+		m.mat = blas64.General{
+			Rows:   r,
+			Cols:   c,
+			Stride: c,
+			Data:   use(m.mat.Data, r*c),
+		}
+	case r > m.capRows || c > m.capCols:
+		t := Dense{mat: blas64.General{
```
btracey: I'm not sure what the right behavior here is, do we want to use max(rows, capRows) etc. instead?

We have a matrix with Rows = 3 , CapRows = 4, Cols = 2, CapCols = 2. We want to add 1 column to the matrix. Does the new matrix have CapRows = 3 or CapRows = 4?

mat64/dense.go
```diff
@@ -164,13 +175,53 @@ func (m *Dense) View(i, j, r, c int) Matrix {
 	t.mat.Data = t.mat.Data[i*t.mat.Stride+j : (i+r-1)*t.mat.Stride+(j+c)]
 	t.mat.Rows = r
 	t.mat.Cols = c
+	t.capRows -= i
+	t.capCols -= j
 	return &t
 }
 
+func (m *Dense) Grow(r, c int) {
```
btracey: Should this return a matrix like view (and append)?
````
func (m *Dense) Grow(r, c int) Matrix
````
It is somewhat complicated to see the interactions with this and other functions, but they're roughly the same complexity as append, and the common case of  m = m.Grow(r,c) works okay.

File: mat64/dense.go (hunk 1 above)
Comment (kortschak) [inline mat64/dense.go]: What do you think you would end up using most?

File: mat64/dense.go (hunk 2 above)
Comment (kortschak) [inline mat64/dense.go]: Sure.

Comment (btracey) [inline mat64/dense.go]: I'm not sure what the cases are where you don't assign it to itself, but there are very few functions that self-modify in this package, so it seemed more consistent to return.

File: mat64/dense.go (hunk 1 above)
Comment (btracey) [inline mat64/dense.go]: I think this is the correct behavior. In order to have rows < capRows, you need to View the matrix and not use all of it. If you then Grow the matrix that's out of bounds with the previous size, you are effectively voiding the data that was there before, which means a new allocation should happen.

Comment (kortschak) [inline mat64/dense.go]: Yes, but do you want it to shrink in the orthogonal axis? I can make it stay at least as big.

Comment (kortschak) [inline mat64/dense.go]: Just to add, it might be a little surprising that it shrank.

Comment (btracey) [inline mat64/dense.go]: Doesn't look like the note got through:

On the flipside, if you grow it, it doesn’t decrease in size, and then you expand it again, you may be surprised to find you have zeros rather than the previous values.

This is a pretty bizarre case. If I have a 4x4 matrix, I view it to have a 4x3 matrix. I Grow an extra row to that, causing an allocation and giving a 5x3 matrix. My personal assumption would be that while I previously stated it had 4 rows, I voided that contract when I added an extra 3-sized row. This meshes with the idea that when I Grow it to get 4 rows, I don’t see the old values I had when it was a 4x4 matrix — new data was allocated. 

That would be my two cents, but it isn’t clear. I hope there isn’t real code that actually relies on this behavior.

**Comment by btracey** [comment]
LGTM with the decision on that issue. I think either case is reasonable, and hopefully anyone who encounters the case and cares will refactor their code to more adequately describe what they want to have happen.

**Comment by kortschak** [comment]
PTAL

Added zeroing and made the matrix not shrink in the orthogonal dimension - it will be zeroed there. We can revisit the issue of whether the elements are copied across at a later date (it's easy to to a copy that goes the whole way from 0,0 to capRows,capCols instead of the normal Copy behaviour.

File: mat64/dense.go [hunk 3]
```diff
@@ -164,6 +176,62 @@ func (m *Dense) View(i, j, r, c int) Matrix {
...
+		t.mat = blas64.General{
+			Rows:   r,
+			Cols:   c,
+			Stride: c,
+			Data:   use(m.mat.Data, r*c),
+		}
+		// We zero here because we don't know how the matrix will be used.
+		// In other places, the mat is immediately filled with a result;
+		// this is not the case here.
+		zero(t.mat.Data[:cap(t.mat.Data)])
```
btracey: Good catch.

btracey:
Okay. We can revisit if this behavior becomes a problem.

LGTM.

kortschak:
I will send a PR later to stop double zeroing when the data slice is newly make'd.

PR #84, opened 2015-01-14
Refactor Vector type

btracey:


btracey:
This commit is not in a pull-state, but I wanted to show my idea before adding all of the tests and putting the functions in the right place.

This changes Vector to be backed by a blas64.Vector rather than just a float64. This has a couple of advantages. First, it increases consistency in the package, as Dense is already backed by a blas64 type, and future Triangular, Symmetric, and Banded are all also likely to be backed by their blas64 equivalent. Secondly, it enables Matrix views of both rows and columns of a matrix. Naming and signatures to be decided, but one option is to have
````
RowView(i int) *Vector
ColView(i int) *Vector
RawRowView(i, data []float64) []float64
````
This matches the idea that "Raw" gets you the underlying data, while the non-raw View stays in safe managed mat64 space. An example of where this could be useful is easy plotting of a column of data.

**Comment by btracey** [comment]
If the idea of this change is acceptable, I'll make this commit-ready.

**Comment by kortschak** [comment]
It seems to me that this would be less invasive if you just made RawVector create the blas64.Vector on demand, and fix up a couple of the panics.

**Comment by btracey** [comment]
It would be less invasive, but aren’t the increased similarities with the other types and the increased interoperability with Matrix a benefit?

**Comment by kortschak** [comment]
You get the same benefits leaving Vec as a []float64. All you need to do is satisfy the interfaces. By hiding the []float inside the Vec struct, you lose direct interaction with the data at no (AFAICS) benefit since there is nothing that needs to be protected.

I don't understand why there is an n when it will just mirror len().

```go
func (m Vector) RawVector() blas64.Vector { return blas64.Vector{Inc:1, Data:m} }
```

```go
func (m Vector) ColView(i int) Vector { 
    if i != 0 {
        panic(ErrIndexOutOfRange)
    }
    return m
}
```

```go
func (m Vector) RowView(i int) Vector { 
    if i < 0 || i >= len(m) {
        panic(ErrIndexOutOfRange)
    }
    return m[i:i+1]
}
```

btracey:
Ah, sorry. The RowView and ColView comments I made earlier were in reference to Dense. 

func (m *Dense) ColView(i int) *Vector {
	return &Vector {
              mat: blas64.Vector {
                   Inc: m.mat.Stride,
                  Data: m.mat.Data[i:i*m.mat.Rows+1],
             },
             n: m.mat.Rows,
    }
}

Since it’s a strided vector, len(vector.mat.Data) is longer than the logical number of elements in the Vector.

kortschak:
Ahh, OK. SGTM.

btracey:
PTAL

mat64/index_bound_checks.go
```diff
@@ -35,3 +35,31 @@ func (m *Dense) set(r, c int, v float64) {
 	}
 	m.mat.Data[r*m.mat.Stride+c] = v
 }
+
+func (m *Vector) At(r, c int) float64 {
+	if c != 0 {
+		panic(ErrColIndexOutOfRange)
```
kortschak: Why are we differentiating these here? We don't do it aywhere else.

mat64/lq.go
```diff
@@ -35,11 +35,11 @@ func LQ(a *Dense) LQFactor {
...
 
 	lDiag := make([]float64, m)
-	projs := make(Vec, m)
+	projs := NewVector(m, nil)
 
 	// Main loop.
 	for k := 0; k < m; k++ {
-		hh := Vec(lq.RowView(k))[k:]
+		hh := lq.RowView(k)[k:]
 		norm := blas64.Nrm2(len(hh), blas64.Vector{Inc: 1, Data: hh})
```
kortschak: I think this is arguing that *Vector should have a Len() method.

mat64/index_bound_checks.go
btracey: If by differentiating, you mean the difference between row and column access, then we do differentiate in the Dense case just above. There isn't a named Error for it, but the plan is to add that as a named error. I had done so in Triangular, though looking now the error names are different, and it looks like the Triangular one is better (same message, more compact name).

mat64/lq.go
btracey: Yes, it probably should. I can add that to this PR if you'd like.

My goal was to change lq as little as possible and have it still be correct.

kortschak: Let's do that later when you add the view methods to *Dense you were suggesting.

btracey:
Changed the Error names. PTAL.

kortschak:
Will they be propagated through later?

LGTM

btracey:
Yes, I'll make that change next.

PR #85, opened 2015-01-14
Remove receiver zeroing in Pow test

kortschak:
blas/native now has the same behaviour as blas/cgo wrt beta == 0 and there are NaN elements, so we don't need to pre-zero where there are NaNs in the matrix.


btracey:
LGTM

PR #86, opened 2015-01-15
Add useZeroed partner to use

kortschak:
The choice of a separate function rather than a bool in use is highlight when the returned slice is expected to be zeroed; a true is likely to be ignored when reading.

Also added test of zeroing and moved zero into matrix.go.

### DIFF
--- a/mat64/matrix.go
+++ b/mat64/matrix.go
@@ -430,3 +430,23 @@ func use(f []float64, l int) []float64 {
 	}
 	return make([]float64, l)
 }
+
+// useZeroed returns a float64 slice with l elements, using f if it
+// has the necessary capacity, otherwise creating a new slice. The
+// elements of the returned slice are guaranteed to be zero.
+func useZeroed(f []float64, l int) []float64 {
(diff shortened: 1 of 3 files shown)


mat64/matrix.go
```diff
@@ -430,3 +430,23 @@ func use(f []float64, l int) []float64 {
...
+func useZeroed(f []float64, l int) []float64 {
+	if l <= cap(f) {
+		f = f[:l]
+		zero(f)
+		return f
+	}
+	return make([]float64, l)
+}
+
+// zero does a fast zeroing of the given slice's elements.
```
btracey: This might actually be slower in 1.5  https://groups.google.com/forum/#!topic/golang-codereviews/lx3hsZr4HZE

btracey: Sorry, this comment is in reference to the for loop.

kortschak: That's fine. It's a single location to change when the time comes.

btracey:
LGTM

PR #87, opened 2015-01-15
Fixup error docs and drop Must

kortschak:
Must is very cheap, we don't need to provide it.

### DIFF
--- a/mat64/matrix.go
+++ b/mat64/matrix.go
@@ -348,7 +348,7 @@ func Solve(a, b Matrix) (x *Dense, err error) {
 // A Panicker is a function that may panic.
 type Panicker func()
 
-// Maybe will recover a panic with a type matrix.Error from fn, and return this error.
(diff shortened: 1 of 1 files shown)


btracey:
LGTM

mat64/matrix.go
```diff
@@ -382,15 +382,7 @@ func MaybeFloat(fn FloatPanicker) (f float64, err error) {
...
-// Must can be used to wrap a function returning an error.
-// If the returned error is not nil, Must will panic.
-func Must(err error) {
-	if err != nil {
-		panic(err)
-	}
-}
-
-// Type Error represents matrix package errors. These errors can be recovered by Maybe wrappers.
+// Type Error represents matrix handling errors. These errors can be recovered by Maybe wrappers.
```
fhs: s/Type //

types are documented beginning with its name (with optional leading article).

fhs: Oops, didn't realized this was already merged. You can take care of this when you like.

PR #89, opened 2015-01-16
Changerowview

btracey:


mat64/matrix.go
```diff
@@ -63,16 +63,28 @@ type VectorSetter interface {
 	SetCol(i int, src []float64) int
 }
 
-// A RowViewer can return a slice of float64 reflecting a row that is backed by the matrix
-// data.
+// A RowViewer can return a Vector reflecting a row that is backed by the matrix
+// data. The Vector returned should have len(nCols).
```
kortschak: "Vector returned should have len(nCols)." This doesn't make sense; it is not possible to call len on a Vector. Lower case Vector and s/should/will/?

mat64/matrix.go
```diff
@@ -63,16 +63,28 @@ type VectorSetter interface {
...
 
-// A ColViewer can return a slice of float64 reflecting a column that is backed by the matrix
+// A RawRowViewer can return a slice of float64 reflecting a row that is backed by the matrix
 // data.
+type RawRowViewer interface {
+	RawRowView(r int) []float64
+}
+
+// A ColViewer can return a Vector reflecting a row that is backed by the matrix
+// data. The Vector returned should have len(nRows).
```
kortschak: As above.

kortschak:
LGTM after comments.

PR #90, opened 2015-01-16
Addsym

btracey:


mat64/symmetric.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/gonum/blas"
+	"github.com/gonum/blas/blas64"
+)
+
+// Symmetric represents a symmetric matrix.
+type Symmetric struct {
+	mat blas64.Symmetric
+}
+
+// NewTriangular constructs an n x n triangular matrix where the data is stored
```
vladimir-ch: NewSymmetric ... symmetric matrix ...

btracey: Oops. Thanks.

mat64/symmetric_test.go
```diff
@@ -0,0 +1,78 @@
+package mat64
+
+import (
+	"github.com/gonum/blas"
+	"github.com/gonum/blas/blas64"
+	"gopkg.in/check.v1"
+)
+
+func (s *S) TestNewSymmetric(c *check.C) {
```
vladimir-ch: Add a test for a panic with ErrShape?

btracey: Done.

btracey:
PTAL

mat64/symmetric_test.go
```diff
@@ -0,0 +1,80 @@
...
+	for i, test := range []struct {
+		data  []float64
+		N     int
+		upper bool
+		mat   *Symmetric
+	}{
+		{
```
kortschak: Add a check for matrix equality with a Dense?

mat64/index_bound_checks.go
```diff
@@ -63,3 +65,47 @@ func (m *Vector) set(r int, v float64) {
...
+		return t.mat.Data[r*t.mat.Stride+c]
+	}
+	if r < c {
+		r, c = c, r
+	}
+	return t.mat.Data[r*t.mat.Stride+c]
+}
+
+// Set sets the element at row r and column c. Set panics if the location is outside
+// the appropriate half of the matrix.
```
kortschak: Why panic rather than setting the appropriate element?

btracey: That is an alternate possible behavior. What worries me is that this satisfies Mutable, but it isn't as mutible as a true Mutable is (whatever that means). For a Dense matrix, 
````
for i := 0; i < rows; i++ {
     for j := 0; j < cols; j++ {
        m.Set(i,j, float64(i*cols + j))
    }
}
````
does exactly what you think it does. For a Symmetric matrix, it does not, and, in fact, if you loop over the rows first and then the columns, you'll get a different answer than if you do it the other way. This feels like a silent bug to me, especially when interfaces get involved. Furthermore, this felt more along the lines of the BLAS routines, where the opposite-triangle elements are never touched.

Comment (btracey) [inline mat64/index_bound_checks.go]: By  "feels like a silent bug", I mean that it seems like it's very easy to accidentally write buggy code that looks fine.

Comment (kortschak) [inline mat64/index_bound_checks.go]: I think this warrants further consideration, but I lean toward the permissive this time. Others' opinions?

Comment (btracey) [inline mat64/index_bound_checks.go]: A possible solution is to have it not satisfy Mutable. For example, it could have SetSym. This is probably a better description anyway, because for a Symmetric, Set does both i,j and j,i

Comment (kortschak) [inline mat64/index_bound_checks.go]: Yeah, maybe. That feels better. I think it still warrants discussion.

Comment (btracey) [inline mat64/index_bound_checks.go]: @jonlawlor @vladimir-ch thoughts? 

I'm leaning toward the SetSym. Sym doesn't really implement Mutable, and @kortschak is right that it doesn't feel like this should panic.

Comment (vladimir-ch) [inline mat64/index_bound_checks.go]: If SetSym replaces Set, then the fact that only a half of the matrix' memory is accessed arises only in NewSymmetric and through RawSymmetric. In that case it seems unnecessary to specify if it is the lower or upper part that is used, upper could be removed from NewSymmetric and one of the two options could be fixed for the underlying blas64.Symmetric. The users could could choose freely whichever part they want to access. On the other hand, we would not know if their random access to the matrix is a mistake or on purpose.

Comment (jonlawlor) [inline mat64/index_bound_checks.go]: I think it would be better for the matrix package as a whole if the matrix types tended to implement the same interfaces where possible.  In this case I think it should implement Set as @vladimir-ch suggests, and trust the users that they would access it correctly.  We can provide an additional SetSymer (or SymSetter)  that is only implemented by symmetric matrixes.  Then a type assertion could reveal to users that they only need to interact with half of a matrix, but if they aren't trying to be sophisticated they could still use Set and have it work correctly (even if it was inefficient).

Comment (btracey) [inline mat64/index_bound_checks.go]: As to your last point, if the function signature is SetSym, the user will be very aware that they're setting two things. Access will always work as expected.

Your first point is a very good one. It's not clear to me why blas allows both orientations of the matrix. My guess would be that there's some trickery one can do to save memory by treating half a Dense as symmetric. I'm not sure though -- it's not like you can fit two symmetric into one Dense. I do like the simplicity from having only Upper rather than needing to support both.

Comment (btracey) [inline mat64/index_bound_checks.go]: @jonlawlor I agree with you about "wherever possible", but I think this is a case where it's not possible. The documentation for Set is "// Set alters the matrix element at (r, c) to v. It will panic if r or c are out of bounds for the matrix.". For symmetric though, Set alters both the matrix element at (r, c) and the matrix element at (c, r). Seems to me like different behavior, different function name.

Comment (vladimir-ch) [inline mat64/index_bound_checks.go]: > As to your last point, if the function signature is SetSym, the user will be very aware that they're setting two things. Access will always work as expected.

True.

Comment (btracey) [inline mat64/index_bound_checks.go]: Also, while the specific case of applying Set will behave as expected, the general case of using Symmetric as a mutable will not.

Comment (vladimir-ch) [inline mat64/index_bound_checks.go]: Agreed, different behavior, different function name. I, too, am leaning toward a non-panicking SetSym (or SymSet). In that case, would NewSymmetric need `upper`? Does it really matter which half is used?

Comment (btracey) [inline mat64/index_bound_checks.go]: I can't think of a reason we would need to support both. The BLAS kernels all support both, and from an external view both are populated.

**Comment by btracey** [comment]
Updated to SetSym which removes the panics. Also made them all Upper, which removes that complexity.

File: mat64/index_no_bound_checks.go [hunk 5]
```diff
@@ -63,3 +65,45 @@ func (m *Vector) Set(r, c int, v float64) {
...
+		panic(ErrRowAccess)
+	}
+	if c >= t.mat.N || c < 0 {
+		panic(ErrColAccess)
+	}
+	return t.at(r, c)
+}
+
+func (t *Symmetric) at(r, c int) float64 {
+	if t.mat.Uplo == blas.Upper {
```
vladimir-ch: Just curious: why is this version of at() comparing Uplo and blas.Upper, while the other at() does not?

btracey: Fixed.

mat64/index_no_bound_checks.go
```diff
@@ -63,3 +63,39 @@ func (m *Vector) Set(r, c int, v float64) {
...
+
+// SetSym sets the elements at (r,c) and (c,r) to the value v.
+func (t *Symmetric) SetSym(r, c int, v float64) {
+	if r >= t.mat.N || r < 0 {
+		panic(ErrRowAccess)
+	}
+	if c >= t.mat.N || c < 0 {
+		panic(ErrColAccess)
+	}
+	if r > c {
```
vladimir-ch: Move this check to set()? Otherwise bound-checking and non-bound-checking set() behave differently.

mat64/symmetric.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/gonum/blas/blas64"
+)
+
+// Symmetric represents a symmetric matrix.
+type Symmetric struct {
+	mat blas64.Symmetric
+}
+
+// NewSymmetric constructs an n x n symmetric matrix where the data is stored
+// in the given orientation. If len(mat) == n * n, mat will be used to hold the
```
vladimir-ch: There is no given orientation anymore.

btracey: Done.

mat64/index_no_bound_checks.go
btracey: Done.

vladimir-ch:
LGTM

We could start using the rebase/squash workflow that is being discussed and see how it pans out. For example this PR could, or even should, be squash to a nice, single commit "Add Symmetric matrix". The subcommits are just noise.

btracey:
Agreed. Someone will have to walk me through that procedure (not a git master yet). 
Waiting for an okay by @kortschak .

mat64/index_bound_checks.go
```diff
@@ -63,3 +63,37 @@ func (m *Vector) set(r int, v float64) {
...
+		r, c = c, r
+	}
+	return t.mat.Data[r*t.mat.Stride+c]
+}
+
+// SetSym sets the elements at (r,c) and (c,r) to the value v.
+func (t *Symmetric) SetSym(r, c int, v float64) {
+	t.set(r, c, v)
+}
+func (t *Symmetric) set(r, c int, v float64) {
```
kortschak: Add new line between funcs.

btracey: Done.

kortschak:
LGTM

The instructions at the cockroachdb/cockroach CONTRIBUTING page are good:

Once ready to land your change, squash your commits. Where n is the number of commits in your branch, run `git rebase -i HEAD~n`

I would then check that you have the history you expect (one commit ahead of master). Note that you can checkout apparently lost commits, the last one in this PR is 7e92d51e1bea97213eb2ca6e6a21905ac3724895, so if everything goes pear shaped, you can `git checkout 7e92d5` and have a detached HEAD that is this state with all the subcommits.

and then when you are happy you haven't broken everything, `git checkout master`, `git merge addsym` and `git push` (implicit origin master).

vladimir-ch:
I like the instructions in https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request

kortschak:
Yes, that's nice. Highlights that I forgot about rebasing master.

btracey:
Any advice on error handling? I followed the instructions in vladimir's link and replaced all of the "pick" with "squash"

````
brendan:~/Documents/mygo/src/github.com/gonum/matrix/mat64$ git merge-base addsym master
aef57554f07f87d740f245baffb697baf8cfafa0
brendan:~/Documents/mygo/src/github.com/gonum/matrix/mat64$ git rebase --interactive aef57554f07f87d740f245baffb697baf8cfafa0
error: could not apply ba43817... Added symmetric At, Set, and added tests

When you have resolved this problem, run "git rebase --continue".
If you prefer to skip this patch, run "git rebase --skip" instead.
To check out the original branch and stop rebasing, run "git rebase --abort".

Could not apply ba4381705ce7b2bda57e6e652f8647d66b3c9884... Added symmetric At, Set, and added tests
````

**Comment by vladimir-ch** [comment]
I am not sure what exactly you did, so just a couple of questions:

* I see you merged master into your addsym branch, that is unnecessary, or may even be the cause of the problem.
* By "all of the pick" you mean "all except the first one pick", don't you?

**Comment by btracey** [comment]
- I had the problem before I merged from master. I thought not being up to date with master was the source of the problem. Apparently merging did not fix.

- Yes, I mean all except the first.

**Comment by vladimir-ch** [comment]
Get rid of the merge commit `git reset --hard HEAD~1` on your branch, run again `git merge-base` just to be sure, and try the interactive rebase again.

**Comment by kortschak** [comment]
I have a fixed addsym branch I can push to master if you want.

It was a bit messy - repeated remerges from master can cause conflicts.

**Comment by vladimir-ch** [comment]
I was just about to write that those problems were probably conflicts.

**Comment by btracey** [comment]
Yea, I saw that there were merge conflicts. Thanks. Working on making sure I actually have one commit now.

**Comment by kortschak** [comment]
Please update the commit message when you are ready at the moment the commit that would be merged has "Uncommented valid tests and commented invalid tests", which is probably not what we want. There just do a `git commit --amend` before you push.

**Comment by btracey** [comment]
Thanks for the help. I followed @kortschak  comment and merged locally and pushed to master, though I probably should have followed @vladimir-ch 's and pushed to here. It's merged now.

Thanks for the --amend reminder.

**Comment by kortschak** [comment]
It looks like we lose the link from the history to the PR.

**Comment by vladimir-ch** [comment]
Something went wrong, I guess. But this is git, everything is fixable.

**Comment by vladimir-ch** [comment]
This is what I would have done to merge this pr:
```bash
$ git pull
$ git checkout addsym
$ git merge-base addsym master
$ git rebase -i ${HASH}
(replace pick with squash from the second line on)
$ vim file_with_conflicts.go
$ git add file_with_conflicts.go
$ git rebase --continue
$ git rebase master
$ git push -f origin addsym
$ git checkout master
$ git merge addsym
$ git push
(delete addsym branch from github web page)
$ git fetch -p (to purge deleted remote branches)
$ git branch -d addsym

I hope I have made no mistake.

**Comment by vladimir-ch** [comment]
And next time to avoid squashing, I would just amend the last commit and push -f it:
```
$ git checkout addsym
(make edits)
$ git add changed_file.go
$ git commit --amend
$ git push -f origin addsym (for a new review)
```
And then for merging:
```
$ git pull
$ git checkout addsym
$ git rebase master (in case master has advanced. Fix conflicts using --continue as above)
$ git push -f origin addsym
$ git checkout master
$ git merge addsym
$ git push
(delete addsym branch from github web page)
$ git fetch -p (to purge deleted remote branches)
$ git branch -d addsym
```

vladimir-ch:
With the above approach github automatically detects that addsym has been merged into master and closes the branch automatically. @btracey , you had to close the branch manually, didn't you?

btracey:
Yes I did.

vladimir-ch:
And that is why this PR is not linked to the merged commit @kortschak . I am not sure, but maybe the key step is to push -f the squashed and rebased branch to github before pushing the master branch. push.default set to simple is very useful.

PR #91, opened 2015-01-18
Fixed incorrect check in Grow.

btracey:
Grow has a fast path if the expansion is zero. However, the fast-path is currently checked as an Or, but the fast path is only when both are zero. It is definitely possible to grow in one dimension and not the other.


btracey:
@kortschak

kortschak:
Oops. LGTM

btracey:
At least it was my code that wasn't working properly.

kortschak:
s/was/wasn't/

PR #95, opened 2015-01-30
Add CopySym and RankTwo methods to Symmetric

btracey:


btracey:
This isn't commit ready to submit, but I'm showing the 50% version before adding all the tests. There needs to be a way to communicate, in a general way, that a matrix is symmetric. Symmetric isn't the only symmetric matrix; Diagonal and BandedSymmetric are for example. I'm interested in hearing if there is a better interface definition than what I have currently.

btracey:
If anyone is curious why these functions, RankTwo is needed for fixing TODOs in optimize.

jonlawlor:
The name is a little confusing.  I'd expect SymmetricMatrix to be a concrete type instead of an interface.  I think that representing it with a marker function is ok; you could also define the SymmetricMatrix interface as matrices that implement SymAt.

btracey:
I'm open to a different name, but since Matrix is an interface, it seems natural that SymmetricMatrix would be as well. I thought about SymAt, but the implementation is identical to At, so it seems like a more complicated version of a marker.

jonlawlor:
Yes, you're right it does fit with `Matrix` - nevermind.  In SymAt, you may be able to avoid a check for c >= r (or vice versa), while in At you could switch their order depending on UpLo.

kortschak:
Couple of comments.

1. Make the interface Symmetric and rename Symmetric SymDense (or similar). The interface is what people should generally thinking about so it's nice to have a short, simple name for that.

2. Make the Symmetric method return the matrix's Uplo - then it's actually doing something for us (not much more, but at least it means that we can check that a blas64.Symmetric that has been passed in is the correct shape).

btracey:
1. SGTM

2. Should it be the blas Uplo, or should we introduce and Upper and Lower into mat64

I was wondering, throughout Dense methods there a check if the matrix is a RawMatrixer. Should this check instead be if the matrix is a Denser (can return a *Dense). If I had a custom Dense, I'd imagine it was either a wrapped version of Dense that had extra methods, or I'd use mat64.NewDense to construct it and return Dense.RawMatrix(). Is there a specific benefit that comes from making users implement RawMatrixer instead of Denser in order to get the faster calls?

btracey:
I ask here because the question is relevant for SymDenser and RawSymmetricer

jonlawlor:
If they wrap it, it will still implement RawMatrix().  Below I'm using Row but it could just as easily be RawMatrix:

```go
package pp

import (
	"bytes"
	"fmt"
	"text/tabwriter"

)

type PrettyPrinter interface {
	PrettyPrint() string
}

// DensePrettyPrint is a Mat64.Dense that implements PrettyPrinter
type DensePrettyPrint struct {
	*mat64.Dense
}

func (dpp DensePrettyPrint) PrettyPrint() string {
	// use a buffer to write to and later turn into a string
	s := new(bytes.Buffer)

	w := new(tabwriter.Writer)
	// \xff is used as an escape delim; see the tabwriter docs
	// align elements to the right as well
	w.Init(s, 1, 1, 1, ' ', tabwriter.StripEscape|tabwriter.AlignRight)

	c, r := dpp.Dims()
	row := make([]float64, c)
	for i := 0; i < r; i++ {
		row = dpp.Row(row, i)

		for _, x := range row {
			fmt.Fprintf(w, "\t %g ", x)
		}
		fmt.Fprintf(w, "\t\n")
	}

	w.Flush()

	return s.String()
}
```

Or, uh, maybe I'm not understanding.

kortschak:
Return blas.Uplo.

Satisfying RawMatrixer is simpler: 1. It explains why it is important - the data are needed for the BLAS call (I think we could do better here with the naming, but only at the cost of breaking a lot of people's code, so that won't happen) and 2. Dense would need a Dense method (or we type switch for the concrete type as well) which feels foolish to me.

btracey:
Thanks for the reiteration on RawMatrix.

btracey:
PTAL

Just curious -- why should Symmetric() return blas.Uplo? There are a couple things wrong with this conceptually. First, this is the interface definition for symmetric within mat64. BLAS has nothing to do with it. Your point about RawSymmetric is good, as there BLAS is the intended purpose, but that's not the case with the interface. Secondly, a symmetric matrix may not have an upper or lower storage format. For example

````
// RankTwo is an implicitly defined rank-two matrix
type RankTwo struct {
    x []float64
    y []float64
}

func (c CartesianProduct) At(i, j) {
    return x[i] * y[j] + x[j] * y[i]
}
````

Which storage format should RankTwo return? Neither is correct.

**Comment by kortschak** [comment]
Thinking aloud. We can define an blas.Uplo constant, InvalidUplo = 0. This should be documented as a value to return in places where an Uplo is expected, but does not make sense in the context of gonum/blas.=

**Comment by btracey** [comment]
Under what conditions would we use that return value? If it actually fits the BLAS model of storage, then it is probably a RawSymmetricer, in which case we don't need the return value. If it doesn't fit the BLAS model accessing from either side is valid, so we also don't care.

**Comment by kortschak** [comment]
OK. Let's think of something else that is actually needed or potentially useful. I have an aversion marker methods; I think they indicate a lack or API coherence.

**Comment by btracey** [comment]
How about:
````
// Symmetric indicates the matrix type is a symmetric matrix, and returns the dimension
// of the matrix (number of rows or columns)
Symmetric() int
````

This way one can say
````
n := a.Symmetric()
````
rather than 
````
n, _ := a.Dims()
````

**Comment by kortschak** [comment]
Nice.

**Comment by btracey** [comment]
PTAL

**Comment by jonlawlor** [comment]
Travis-ci just switched the go versioning that they use to gimme.  I'm working on fixing the build.

**Comment by jonlawlor** [comment]
There is no release tagged `release` in the git repo for go, so I have to change the .travis.yml file to use `1.4.1` instead of `release`.  On the plus side, `tip` works again.

**Comment by btracey** [comment]
Thanks for fixing that.

File: mat64/symmetric.go [hunk 1]
```diff
@@ -5,16 +5,41 @@ import (
...
 	mat blas64.Symmetric
 }
 
-// NewSymmetric constructs an n x n symmetric matrix. If len(mat) == n * n,
+// Symmetric represents a symmetric matrix (where the element at {i, j} equals
+// the element at {j, i}). Symmetric matrices are always square.
+type Symmetric interface {
+	Matrix
+	// Symmetric returns the number of rows/columns in the matrix.
+	Symmetric() int
```
kortschak: Sorry. Maybe `Size() int`?

kortschak: Or then again... maybe not. That would fail to distinguish Symmetric from Triangular.

btracey: It shouldn't be that general. The interface for Triangular will also heed a similar third function. Triangular should not implement Symmetric and vis-versa.

kortschak:
LGTM

mat64/symmetric.go
btracey: Agreed :)

jonlawlor:
The travis-ci issue is fixed in master, please disregard the error in this branch for the `release` version.

PR #100, opened 2015-02-03
Syminner

btracey:


mat64/inner.go
```diff
@@ -27,6 +27,17 @@ func Inner(x []float64, A Matrix, y []float64) float64 {
 	var sum float64
 
 	switch b := A.(type) {
+	case RawSymmmetricer:
```
fhs: There seems be an extra 'm' in the interface name.

Also, we have an interface named Symmetric, which seems inconsistent with RawSymmetricer. Interface names are usually `<verb>-er` or `<noun>`, so I would favor RawSymmetric.

btracey: Thanks for the catch

This particular case is weird. The main matrix-type interfaces are Matrix and Symmetric. All of the other interfaces follow the normal _er convention, including RawMatrixer. RawSymmetricer follows that precedent.

btracey:
PTAL

kortschak:
LGTM

PR #101, opened 2015-02-03
Added RawSymmetric case to MulVec

btracey:


kortschak:
At this stage it's worth replacing the series of conditional assertions with a type switch.

btracey:
Pulled from master to fix spelling.

PTAL.

kortschak:
LGTM

For future reference, use rebase rather than merge to bring master in. It'll save bother in the final merge.

PR #103, opened 2015-02-04
Fix symmetric Inner

btracey:
The previous implementation dotted along the row with x and the column with y. This is incorrect, because the data is only valid in the upper triangular half of the matrix. Instead, dot along the row for both x and y.


btracey:
Merging because it fixes the broken code. This is also tested locally with optimize and fixes the failure status over there.

kortschak:
Sorry.

PR #107, opened 2015-02-18
fix index out of range on lu decomp when m < n

rwcarlsen:
I'm not sure this is the fix you will want, but currently, LUGaussian panics when there are more columns than rows.  I haven't checked LU.

### DIFF
--- a/mat64/lu.go
+++ b/mat64/lu.go
@@ -137,7 +137,7 @@ func LUGaussian(a *Dense) LUFactors {
 		}
 
 		// Compute multipliers and eliminate k-th column.
-		if lu.at(k, k) != 0 {
+		if k >= m || lu.at(k, k) != 0 {
 			for i := k + 1; i < m; i++ {
 				lu.set(i, k, lu.at(i, k)/lu.at(k, k))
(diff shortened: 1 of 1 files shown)


rwcarlsen:
LU seems to work fine - just LUGaussian was broken.

kortschak:
I vaguely recall we were planning to drop LUGaussian, but having it correct before putting it to sleep would be nice. Do you have any memory of that @btracey?

@rwcarlson, please revert the change to the test imports it is not necessary, I we're going to keep LUGaussian it would be good to add it to the tests though.

kortschak:
... for this particular previously failing case.

rwcarlsen:
sorry - goimports hook got me.  This also fixes the same problem on the type LU in general - calling the U() method when m < n also resulted in an index out of range panic.

btracey:
I don't remember the previous conversation, but it seems like it should eventually go to sleep. Personally, I think a lot of these should be rethought with Lapack and SymDense etc. in mind. However, that requires time and effort, and so in the mean time the code should be correct. 

I agree that we should add tests. Thanks for the contribution @rwcarlsen

PR #110, opened 2015-03-06
Fixing bug in colView in Dense

evahlis:
Without the fix, code that uses colView crashes, unless the underlying []float64 has cap() > len().

### DIFF
--- a/mat64/dense.go
+++ b/mat64/dense.go
@@ -132,7 +132,7 @@ func (m *Dense) ColView(j int) *Vector {
 	return &Vector{
 		mat: blas64.Vector{
 			Inc:  m.mat.Stride,
-			Data: m.mat.Data[j : m.mat.Rows*m.mat.Stride+j],
+			Data: m.mat.Data[j : (m.mat.Rows-1)*m.mat.Stride+j+1],
 		},
 		n: m.mat.Rows,
 	}
--- a/mat64/matrix_test.go
+++ b/mat64/matrix_test.go
@@ -38,9 +38,12 @@ func panics(fn func()) (panicked bool, message string) {
 }
 
 func flatten(f [][]float64) (r, c int, d []float64) {
+	collected := make([]float64, 0)
 	for _, r := range f {
-		d = append(d, r...)
+		collected = append(collected, r...)
 	}
+	d = make([]float64, len(collected))
+	copy(d, collected[:len(d)])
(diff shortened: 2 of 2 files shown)


kortschak:
Please add a test that fails without the change.

mat64/dense_test.go
```diff
@@ -307,6 +307,79 @@ func (s *S) TestRowColView(c *check.C) {
...
+		{
+			mat: [][]float64{
+				{1, 2, 3, 4, 5},
+				{6, 7, 8, 9, 10},
+				{11, 12, 13, 14, 15},
+				{16, 17, 18, 19, 20},
+			},
+		},
+	} {
+		rows, cols, flattened := flatten(test.mat)
```
kortschak:
```
rows, cols, flat := flatten(test.mat)
m := NewDense(rows, cols, flat[:len(flat):len(flat)])
```

I think I'll also change flatten since that would have found this earlier. I'll send that.

kortschak:
LGTM after comment addressed.

Please send a PR against gonum/license adding yourself to A+C, and squash this into one commit.

I'll merge later today after that is done.

kortschak:
Out of curiosity how did you find this? I'm guessing you did NewDense(r, c, nil) to create the *Dense.

mat64/dense_test.go
kortschak: Please make this change and squash.

kortschak:
@evahlis, to clarify: I don't want this merged until the changes have been made to the [*Dense creation stanza](https://github.com/gonum/matrix/pull/110/files#diff-b99bb5e8ca364fa347701e7f763aa01bR341).

The three-index slice index is both more succinct and, significantly here, more clearly describes the intention - to ensure that that cap equals len. Note that this line will be redundant as far as the compiler is concerned after #111 is merged, but will be retained to ensure future readers understand the reason for the test and to decouple this test from possible future changes to flatten.

mat64/dense_test.go
btracey: Ping.

evahlis: Waiting for me?

btracey: Yes, to make the change Dan requested and to then squash the commits.

kortschak: This test should have remained. The change to flatten should not have happened.

PR #111, opened 2015-03-06
Make flatten stricter in data preparation

kortschak:
* Panic on ragged input or rowless data.
* Build data slice that has exact cap - this is what NewDense does, so  we should match that.

This change catches a bug in (*Dense).ColView that will be fixed with a PR to be merged later today (#110).

### DIFF
--- a/mat64/matrix_test.go
+++ b/mat64/matrix_test.go
@@ -38,13 +38,19 @@ func panics(fn func()) (panicked bool, message string) {
 }
 
 func flatten(f [][]float64) (r, c int, d []float64) {
-	collected := make([]float64, 0)
-	for _, r := range f {
-		collected = append(collected, r...)
+	r = len(f)
+	if r == 0 {
+		panic("bad test: no row")
 	}
-	d = make([]float64, len(collected))
-	copy(d, collected[:len(d)])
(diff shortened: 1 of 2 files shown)


mat64/matrix_test.go
```diff
@@ -38,10 +38,19 @@ func panics(fn func()) (panicked bool, message string) {
 }
 
 func flatten(f [][]float64) (r, c int, d []float64) {
-	for _, r := range f {
-		d = append(d, r...)
+	r = len(f)
+	if r == 0 {
+		panic("bad test: no row")
```
btracey: Maybe these should just be "flatten: no row".

btracey:
LGTM besides minor nit.

mat64/matrix_test.go
kortschak: It's not an issue with flatten though. I want the message to be to either the author of the test that they wrote a bad test (or to an unfortunate client that got code through a mistake in review that it isn't their fault).

btracey: SGTM

kortschak:
@btracey PTAL

btracey:
LGTM

PR #114, opened 2015-03-20
Make Cholesky take a SymDense

kortschak:
@vladimir-ch @btracey PTAL

Some times I wonder why you guys put up with my drivel.

### DIFF
--- a/mat64/cholesky.go
+++ b/mat64/cholesky.go
@@ -7,84 +7,124 @@ package mat64
 
 import (
 	"math"
+
+	"github.com/gonum/blas"
+	"github.com/gonum/blas/blas64"
+	"github.com/gonum/internal/asm"
 )
 
-type CholeskyFactor struct {
-	L   *Dense
-	SPD bool
-}
+// Cholesky calculates the Cholesky decomposition of the matrix A and returns
+// whether the matrix is positive definite. The returned matrix is either a
+// lower triangular matrix such that A = L * L^T or an upper triangular matrix
+// such that A = U^T * U depending on the upper parameter.
+func (t *Triangular) Cholesky(a *SymDense, upper bool) (ok bool) {
+	n := a.Symmetric()
+	if t.isZero() {
+		t.mat = blas64.Triangular{
+			N:      n,
+			Stride: n,
+			Diag:   blas.NonUnit,
+			Data:   use(t.mat.Data, n*n),
+		}
+	} else if n != t.mat.N {
+		panic(ErrShape)
+	}
+	mat := t.mat.Data
+	stride := t.mat.Stride
 
-// CholeskyL returns the left Cholesky decomposition of the matrix a and whether
-// the matrix is symmetric and positive definite. The returned matrix l is a lower
-// triangular matrix such that a = l.l'.
-func Cholesky(a *Dense) CholeskyFactor {
-	// Initialize.
-	m, n := a.Dims()
-	spd := m == n
-	l := NewDense(n, n, nil)
-
-	// Main loop.
-	for j := 0; j < n; j++ {
-		lRowj := l.RawRowView(j)
-		var d float64
-		for k := 0; k < j; k++ {
-			var s float64
-			for i, v := range l.RawRowView(k)[:k] {
-				s += v * lRowj[i]
+	if upper {
+		t.mat.Uplo = blas.Upper
+		for j := 0; j < n; j++ {
+			var d float64
+			for k := 0; k < j; k++ {
+				s := asm.DdotInc(
+					mat, mat,
+					uintptr(k),
+					uintptr(stride), uintptr(stride),
+					uintptr(k), uintptr(j),
+				)
+				s = (a.at(j, k) - s) / t.at(k, k)
+				t.set(k, j, s)
+				d += s * s
 			}
-			s = (a.at(j, k) - s) / l.at(k, k)
-			lRowj[k] = s
-			d += s * s
-			spd = spd && a.at(k, j) == a.at(j, k)
+			d = a.at(j, j) - d
+			if d <= 0 {
+				t.Reset()
+				return false
+			}
+			t.set(j, j, math.Sqrt(math.Max(d, 0)))
 		}
-		d = a.at(j, j) - d
-		spd = spd && d > 0
-		l.set(j, j, math.Sqrt(math.Max(d, 0)))
-		for k := j + 1; k < n; k++ {
-			l.set(j, k, 0)
+	} else {
+		t.mat.Uplo = blas.Lower
+		for j := 0; j < n; j++ {
+			var d float64
+			for k := 0; k < j; k++ {
+				s := asm.DdotUnitary(mat[k*stride:k*stride+(n-k)], mat[j*stride:j*stride+(n-k)])
+				s = (a.at(j, k) - s) / t.at(k, k)
+				t.set(j, k, s)
+				d += s * s
+			}
+			d = a.at(j, j) - d
+			if d <= 0 {
+				t.Reset()
+				return false
+			}
+			t.set(j, j, math.Sqrt(math.Max(d, 0)))
 		}
 	}
 
-	return CholeskyFactor{L: l, SPD: spd}
+	return true
 }
 
-// CholeskySolve returns a matrix x that solves a.x = b where a = l.l'. The matrix b must
-// have the same number of rows as a, and a must be symmetric and positive definite. The
-// matrix b is overwritten by the operation.
-func (f CholeskyFactor) Solve(b *Dense) (x *Dense) {
-	if !f.SPD {
-		panic("mat64: matrix not symmetric positive definite")
-	}
-	l := f.L
-
-	m, n := l.Dims()
+// SolveCholesky finds the matrix x that solves A * X = B where A = L * L^T or
+// A = U^T * U, and U or L are represented by t. The matrix A must be symmetric
+// and positive definite.
+func (m *Dense) SolveCholesky(t *Triangular, b Matrix) {
+	_, n := t.Dims()
 	bm, bn := b.Dims()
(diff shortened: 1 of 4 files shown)


mat64/cholesky_test.go
```diff
@@ -10,21 +10,28 @@ import (
...
+			a: NewSymDense(3, []float64{
 				4, 1, 1,
-				1, 2, 3,
-				1, 3, 6,
+				0, 2, 3,
+				0, 0, 6,
 			}),
 
-			spd: true,
+			l: NewDense(3, 3, []float64{
```
kortschak: This is from octave - though transposed.

btracey:
It seems to me like the signature should be 
````
func Cholesky(a *SymDense) (*Triangular,error)
````
The Cholesky decomposition doesn't work if the matrix is not symmetric positive definite (dpotf2, for example, stops early if it is not). Therefore, it seems like it should be an error instead of a boolean in the struct. 

Secondly, instead of having the CholeskyFactor and the Solve method, it would seem better to just add a Solve method (with Matrix input) to Triangular that calls Dtrsm, Dtrsv, or uses the current code. We can then add a comment to the Cholesky function noting that this can be done.

**Comment by kortschak** [comment]
Error or success boolean? If there's only one failure mode, boolean seems better.

SGTM generally.

**Comment by btracey** [comment]
I think either is fine. If it's a boolean, it should be a named return so it's clear what the boolean means from the signature.

**Comment by kortschak** [comment]
I was planning on using "ok". The return of nil, false will lead people to read the docs if they haven't otherwise.

**Comment by btracey** [comment]
SGTM

**Comment by kortschak** [comment]
We now also have the possibility of the user choosing whether to get an upper or a lower triangle. 
Should I add that?

**Comment by btracey** [comment]
When uncertain, leaning toward Lapack seems like good policy. Lapack does
````
The factorization has the form
     A = U**T * U , if UPLO = 'U', or
     A = L * L**T, if UPLO = 'L',
````
which seems reasonable to me.

**Comment by btracey** [comment]
Oh, also, we should think about our normal style and allowing a matrix to be used in-place.
Cholesky(*Triangular, SymDense)

**Comment by kortschak** [comment]
@btracey PTAL

I had a stab at using blas64.Trsm and I obviously don't know how to use it. When I called blas64.Trsm(blas.Left, blas.NoTrans, 1, t.mat, b.RawMatrix()) (with some other relevant changes needed for shadowing etc) I don't get the result we expect. Any ideas?

File: mat64/cholesky.go [hunk 2]
```diff
@@ -7,84 +7,180 @@ package mat64
...
 )
 
-type CholeskyFactor struct {
-	L   *Dense
-	SPD bool
-}
+// Cholesky calculates the Cholesky decomposition of the matrix A and returns
+// whether the matrix is positive definite. The returned matrix is either a
+// lower triangular matrix such that A = L * L^T or an upper triangular matrix
+// such that A = U^T * U depending on the upper parameter.
```
btracey: I think there should be a "panics if triangular is of non-zero size and orientation does not match upper"

mat64/cholesky.go
```diff
@@ -7,84 +7,180 @@ package mat64
...
-// have the same number of rows as a, and a must be symmetric and positive definite. The
-// matrix b is overwritten by the operation.
-func (f CholeskyFactor) Solve(b *Dense) (x *Dense) {
-	if !f.SPD {
-		panic("mat64: matrix not symmetric positive definite")
-	}
-	l := f.L
-
-	m, n := l.Dims()
+// Solve returns a matrix x that solves a.x = b where a = t.t'. The matrix a
```
btracey: This comment is not correct. It should be that solve finds x where a is triangular.

mat64/cholesky.go
kortschak: We don't say that on method comments for anything else. Why here?

mat64/cholesky.go
kortschak: A isn't triangular here. The factors are - A is symmetric. I've fixed the comment.

btracey:
Do you have a plan in mind for the Solve function? There are a number of inconsistencies between this Solve and the solve function. Maybe these should both be methods on dense:
````
// Solve finds the matrix x where a * x = b and puts the result in the receiver
(m *Dense) Solve(a, b Matrix) error
// SolveTri finds the matrix x where a * x = b and a is triangular and puts the result in the receiver
(m *Dense) SolveTri(a *Triangular, b Matrix) error
````
I think I'd also prefer we not use the Mutable thing. That way we maintain that it's the receivers that change and not the inptus

**Comment by kortschak** [comment]
I'm happy to have it on Dense. I was just putting it on Triangular from your suggestion.

**Comment by btracey** [comment]
I hadn't considered Cholesky as a method on Triangular, but that does seem to be correct with the package design. With that in mind, it seems like the Solves should be methods on Dense, as that is where the receiver is. Does that make sense?

**Comment by btracey** [comment]
Your Dtrsm call seems fine to me. I'll try to look into it tomorrow.

**Comment by kortschak** [comment]
Yeah, I just over interpretted your comment above. I've done the change to make it a method on Dense and I think now I just need to spend some time away from the screen.

**Comment by kortschak** [comment]
OK. Figured out why blas64.Trsm wasn't working and also why there is a confusion around the triangularity of A.

The solve as it was, was not a triangular solve, but rather a general solve that makes use of a pair of triangular solves where the triangles are given by the chol. I've retained the old behaviour, but that is up for discussion.

Now I'm going to walk my dog.

File: mat64/cholesky.go [hunk 4]
```diff
@@ -7,84 +7,109 @@ package mat64
...
-	// Initialize.
-	m, n := a.Dims()
-	spd := m == n
-	l := NewDense(n, n, nil)
+// Cholesky calculates the Cholesky decomposition of the matrix A and returns
+// whether the matrix is positive definite. The returned matrix is either a
+// lower triangular matrix such that A = L * L^T or an upper triangular matrix
+// such that A = U^T * U depending on the upper parameter.
+func (t *Triangular) Cholesky(a *SymDense, upper bool) (ok bool) {
+	_, n := a.Dims()
```
btracey: Should this be a call to Symmetric() instead?

mat64/cholesky.go
```diff
@@ -7,84 +7,109 @@ package mat64
...
-				s += v * lRowj[i]
+	if upper {
+		t.mat.Uplo = blas.Upper
+		for j := 0; j < n; j++ {
+			var d float64
+			for k := 0; k < j; k++ {
+				s := asm.DdotInc(
+					mat, mat,
+					uintptr(k),
+					uintptr(n), uintptr(n),
```
btracey: Shouldn't "n" be the stride instead?

mat64/cholesky.go
```diff
@@ -7,84 +7,109 @@ package mat64
...
-		l.set(j, j, math.Sqrt(math.Max(d, 0)))
-		for k := j + 1; k < n; k++ {
-			l.set(j, k, 0)
+	} else {
+		t.mat.Uplo = blas.Lower
+		for j := 0; j < n; j++ {
+			var d float64
+			for k := 0; k < j; k++ {
+				// Short slice must go first.
+				s := asm.DdotUnitary(mat[j*n:], mat[k*n:])
```
btracey: Similarly, I feel like there should be a stride here. Also, shouldn't there be some form of cap? If k == j == 0, then it's a dot product over all the elements of the matrix? The above code looks like it should just be the rows.

mat64/cholesky.go
```diff
@@ -7,84 +7,109 @@ package mat64
...
-	if !f.SPD {
-		panic("mat64: matrix not symmetric positive definite")
-	}
-	l := f.L
-
-	m, n := l.Dims()
+// SolveTri finds the matrix x that solves A * X = B where A = L * L^T or
+// A = U^T * U, and U or L are represented by t. The matrix A must be symmetric
+// and positive definite.
+func (m *Dense) SolveTri(t *Triangular, b Matrix) error {
```
btracey: This should probably be SolveChol, so that SolveTri can be the direct wrapping around Dtrsm. Sorry for my lack of understanding previously.

mat64/cholesky.go
```diff
@@ -7,84 +7,109 @@ package mat64
...
+	_, n := t.Dims()
```
btracey: Symmetric()?

mat64/cholesky.go
```diff
@@ -7,84 +7,109 @@ package mat64
...
 	bm, bn := b.Dims()
-	if m != bm {
-		panic(ErrShape)
+	if n != bm {
+		return ErrShape
```
btracey: Why is this return? We panic elsewhere on ErrShape.

mat64/cholesky.go
kortschak: I didn't know that existed.

mat64/cholesky.go
kortschak: t is not SymDense.

mat64/cholesky.go
kortschak: The comment about cap is significant.

I won't fix it here, but I think we don't treat capCols and capRows correctly in any of the cases where m.isZero() == true. In short, when we get a use result that allocated a new slice, the cap values are now incorrect. I think the best way to fix this is factor the isZero == true  consequences out into a separate function - this will reduce the line count significantly and centralise that behaviour.

kortschak:
@PTAL

kortschak:
PTAL

btracey:
LGTM. 

Thanks for your patience.

kortschak:
Added minor change, but I'm going to merge now (It will be changed in a coming PR and is essentially just a place holder). If you are interested, the change is here 088ee8b03d94b20a2edfdd8435b4882a76e2b862.

kortschak:
Thank you for the review. It was very helpful.

Issue #115, opened 2015-03-22
mat64: capRows and capCols are not correctly handled in arithmetic reuse cases


kortschak:
When m.isZero() == true, a call to use is made. This call may result in the allocation of a new data slice, invalidating the previous cap values. We don't handle this at all. The simplest case of this is when a Dense{} is used; since the data slice is nil, a new data slice is allocated, but (AFAICS) in no case is capRows or capCols set to reflect that this matrix now has a capacity.

The best approach I think is to factor the m.isZero == true consequences out into a separate method, func (m *Dense) make(r, c int, zero bool).

This arose in discussion in #114.

PR #116, opened 2015-03-23
mat64: fix Dense.{capRows,capCols} handling

kortschak:
Dense.{capRows,capCols} handling is currently broken. This fixes that.

See #115.

@btracey

### DIFF
--- a/mat64/dense.go
+++ b/mat64/dense.go
@@ -90,6 +90,29 @@ func NewDense(r, c int, mat []float64) *Dense {
 	}
 }
 
+// reUseAs resizes an empty matrix to a r×c matrix,
+// or checks that a non-empty matrix is r×c.
+func (m *Dense) reUseAs(r, c int) {
+	if m.isZero() {
+		m.mat = blas64.General{
+			Rows:   r,
+			Cols:   c,
+			Stride: c,
+			Data:   use(m.mat.Data, r*c),
+		}
+		return
+	}
+	if r != m.mat.Rows || c != m.mat.Cols {
+		panic(ErrShape)
+	}
+}
+
+func (m *Dense) isZero() bool {
+	// It must be the case that m.Dims() returns
+	// zeros in this case. See comment in Reset().
+	return m.mat.Stride == 0
+}
+
 // DenseCopyOf returns a newly allocated copy of the elements of a.
 func DenseCopyOf(a Matrix) *Dense {
 	d := &Dense{}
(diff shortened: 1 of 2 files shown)


kortschak:
My current thinking about how to ensure this is correct is to change index_bounds_checks.go to include the following:

```
func (m *Dense) At(r, c int) float64 {
	return m.at(r, c)
}

func (m *Dense) at(r, c int) float64 {
	if r >= m.mat.Rows || r < 0 {
		panic(ErrRowAccess)
	}
	if c >= m.mat.Cols || c < 0 {
		panic(ErrColAccess)
	}
	if m.capRows == 0 || m.capCols == 0 {
		panic("caps not set")
	}
	return m.mat.Data[r*m.mat.Stride+c]
}

func (m *Dense) Set(r, c int, v float64) {
	m.set(r, c, v)
}

func (m *Dense) set(r, c int, v float64) {
	if r >= m.mat.Rows || r < 0 {
		panic(ErrRowAccess)
	}
	if c >= m.mat.Cols || c < 0 {
		panic(ErrColAccess)
	}
	if m.capRows == 0 || m.capCols == 0 {
		panic("caps not set")
	}
}
```

This breaks symmetry with no bounds checks, but I think that is OK here. Waiting on other opinions.

mat64/dense.go
```diff
@@ -90,6 +90,31 @@ func NewDense(r, c int, mat []float64) *Dense {
 	}
 }
 
+// reUseAs resizes an empty matrix to a r×c matrix,
+// or checks that a non-empty matrix is r×c.
+func (m *Dense) reUseAs(r, c int) {
```
btracey: Sorry for the bikeshedding, but isn't reuse one word?

mat64/dense_arithmetic.go
```diff
@@ -674,9 +620,10 @@ func (m *Dense) Exp(a Matrix) {
 			Rows:   r,
```
btracey: I don't see what's different here. Why doesn't this use reuseAs?

mat64/dense.go
kortschak: Could be hyphenated - but I was thinking similar things. I'll change it.

btracey:
I'd be okay with both of your proposed changes. Doesn't need to happen in this PR.

mat64/dense_arithmetic.go
kortschak: It's a special case because Exp requires the matrix to be eye. I could rewrite it to use this code, but I would rather not.

kortschak:
I'd like to have the test change in the PR as well.

mat64/dense_arithmetic.go
btracey: Oh, I see. SGTM.

btracey:
"The test change" is the bounds_checks change? If so, SGTM.

kortschak:
The reuseAs preamble is my preferred option.

btracey:
Okay.

kortschak:
PTAL

btracey:
LGTM

PR #117, opened 2015-03-23
mat64: improve error handling

kortschak:
Previously, we used Error string as potentially recoverable mat64 failures (clients can obviously choose to recover and panics we call, but these are the one we think are not too serious). That approach allowed clients to create new mat64.Errors.

This change prevents clients from creating new mat64.Error values.

### DIFF
--- a/mat64/matrix.go
+++ b/mat64/matrix.go
@@ -365,8 +365,11 @@ type Panicker func()
 func Maybe(fn Panicker) (err error) {
 	defer func() {
 		if r := recover(); r != nil {
-			var ok bool
-			if err, ok = r.(Error); ok {
+			if e, ok := r.(Error); ok {
+				if e.string == "" {
+					panic("mat64: invalid error")
+				}
+				err = e
 				return
 			}
 			panic(r)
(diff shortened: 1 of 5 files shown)


mat64/matrix.go
```diff
@@ -395,23 +401,23 @@ func MaybeFloat(fn FloatPanicker) (f float64, err error) {
...
-	ErrZeroLength      = Error("mat64: zero length in matrix definition")
-	ErrRowLength       = Error("mat64: row length mismatch")
-	ErrColLength       = Error("mat64: col length mismatch")
-	ErrSquare          = Error("mat64: expect square matrix")
-	ErrNormOrder       = Error("mat64: invalid norm order for matrix")
-	ErrSingular        = Error("mat64: matrix is singular")
-	ErrShape           = Error("mat64: dimension mismatch")
-	ErrIllegalStride   = Error("mat64: illegal stride")
-	ErrPivot           = Error("mat64: malformed pivot list")
+type Error struct{ string }
```
btracey: What's the rationale for being a struct instead of just a string type?

kortschak: It allows us to own the mat64.Error type. With this change, it is not possible to another package to create a new non-zero Error.

btracey: I see. I didn't realize one could do this. Interesting!

btracey:
LGTM

mat64/matrix.go
kortschak: It's why errors.errorString is not a string.

http://golang.org/src/errors/errors.go

PR #118, opened 2015-03-26
Change Triangular to TriDense and add Triangular interface.

btracey:
This brings the Triangular type in line with Dense and Symmetric

### DIFF
--- a/mat64/cholesky.go
+++ b/mat64/cholesky.go
@@ -19,7 +19,7 @@ const badTriangle = "mat64: invalid triangle"
 // whether the matrix is positive definite. The returned matrix is either a
 // lower triangular matrix such that A = L * L^T or an upper triangular matrix
 // such that A = U^T * U depending on the upper parameter.
-func (t *Triangular) Cholesky(a *SymDense, upper bool) (ok bool) {
+func (t *TriDense) Cholesky(a *SymDense, upper bool) (ok bool) {
 	n := a.Symmetric()
 	if t.isZero() {
 		t.mat = blas64.Triangular{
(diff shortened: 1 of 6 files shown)


mat64/cholesky.go
```diff
@@ -81,7 +81,7 @@ func (t *Triangular) Cholesky(a *SymDense, upper bool) (ok bool) {
 // SolveCholesky finds the matrix x that solves A * X = B where A = L * L^T or
 // A = U^T * U, and U or L are represented by t. The matrix A must be symmetric
 // and positive definite.
-func (m *Dense) SolveCholesky(t *Triangular, b Matrix) {
+func (m *Dense) SolveCholesky(t *TriDense, b Matrix) {
```
kortschak: Is there a reason this doesn't also do the RawTriangular dance you have below?

btracey: If you'd like it can, but the thought was that it would always be used after calling TriDense.Cholesky.

btracey: But there's no good reason not to allow an arbitrary Triangular.

btracey:
PTAL

mat64/cholesky.go
kortschak: Not fussed really, just wondering.

mat64/cholesky.go
```diff
@@ -107,8 +111,8 @@ func (m *Dense) SolveCholesky(t *Triangular, b Matrix) {
 
 // SolveTri finds the matrix x that solves op(A) * X = B where A is a triangular
 // matrix and op is specified by trans.
-func (m *Dense) SolveTri(a *Triangular, trans bool, b Matrix) {
-	_, n := a.Dims()
+func (m *Dense) SolveTri(a Triangular, trans bool, b Matrix) {
+	n, _ := a.Triangular()
```
kortschak: Triangle?

mat64/triangular.go
```diff
@@ -6,15 +6,29 @@ import (
...
+// format.
+type TriDense struct {
 	mat blas64.Triangular
 }
 
+type Triangular interface {
+	Matrix
+	// Triangular returns the number of rows/columns in the matrix and if it is
+	// an upper triangular matrix.
+	Triangular() (n int, upper bool)
```
kortschak: Triangle?

kortschak:
It seems to me that a better name for the triangular dims equivalent is Triangle. Triangular is an adjective, to if I say it without its return values, I would guess that it was a badly named IsTriangular() bool. Changing to Triangle() int, bool also has the advantage that the interface name is now different to the interface's method: Triangler.

btracey:
Done

kortschak:
LGTM

PR #120, opened 2015-04-03
Fix slice size in TriDense.Cholesky()

### LINKED ISSUE #119: Incorrect slice size in TriDense.Cholesky()
When `upper == false`, slice size for `DdotUnitary()` should be `k`, not `(n-k)`.

vladimir-ch:
What would be the best way to update the tests? Leave the existing tests and add a new function `TestIssue119()`, or change the tests to use non-zero TriDense receiver?

Fixes #119


kortschak:
LGTM after tests added. Thanks.

I would just add extra tests into the test slice.

(I have no idea what I was thinking when I wrote that).

kortschak:
When you add the test to the slice, poison all the elements with NaN.

PR #124, opened 2015-05-05
mat64: add element-wise arithmetic to Vector

kortschak:
Also fix doc comment on Dense DivElem.

@btracey

### DIFF
--- a/mat64/vector.go
+++ b/mat64/vector.go
@@ -103,6 +103,76 @@ func (m *Vector) RawVector() blas64.Vector {
 	return m.mat
 }
 
+// AddVec adds a and b element-wise, placing the result in the receiver.
+func (v *Vector) AddVec(a, b *Vector) {
+	ar := a.Len()
+	br := b.Len()
+
+	if ar != br {
+		panic(ErrShape)
+	}
+
+	v.reuseAs(ar)
+
+	amat, bmat := a.RawVector(), b.RawVector()
+	for i := 0; i < v.n; i++ {
+		v.mat.Data[i*v.mat.Inc] = amat.Data[i*amat.Inc] + bmat.Data[i*bmat.Inc]
+	}
+}
+
+// SubVec subtracts the vector b from a, placing the result in the receiver.
+func (v *Vector) SubVec(a, b *Vector) {
+	ar := a.Len()
+	br := b.Len()
+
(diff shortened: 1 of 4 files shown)


btracey:
I don't think this is a good idea. As much as reasonable, we should try to only provide "type safe" methods. For example, here we should see 
````
func ( v Vector) AddVec(a, b Vector)
````
As a vector is the result of adding a vector and a vector, not the result of adding a Matrix and a Matrix.

**Comment by kortschak** [comment]
PTAL

**Comment by kortschak** [comment]
Ignore that. Fixing now.

**Comment by kortschak** [comment]
PTAL

File: mat64/vector.go [hunk 1]
```diff
@@ -103,6 +103,74 @@ func (m *Vector) RawVector() blas64.Vector {
...
+
+	v.reuseAs(ar)
+
+	amat, bmat := a.RawVector(), b.RawVector()
+	for i := 0; i < v.n; i++ {
+		v.mat.Data[i*v.mat.Inc] = amat.Data[i*amat.Inc] - bmat.Data[i*bmat.Inc]
+	}
+}
+
+// MulElemVec performs element-wise multiplication of a and b, placing the result
```
btracey: Comment missing.

mat64/vector.go
```diff
@@ -103,6 +103,74 @@ func (m *Vector) RawVector() blas64.Vector {
...
+
+	v.reuseAs(ar)
+
+	amat, bmat := a.RawVector(), b.RawVector()
+	for i := 0; i < v.n; i++ {
+		v.mat.Data[i*v.mat.Inc] = amat.Data[i*amat.Inc] * bmat.Data[i*bmat.Inc]
+	}
+}
+
+// DivElemVec performs element-wise division of a by b, placing the result
```
btracey: Same here.

btracey:
LGTM with comments fixed.

mat64/vector.go
kortschak: Fixed.

mat64/vector.go
kortschak: Fixed.

Thanks for your patience.

kortschak:
travis pr tests failed because it hadn't pulled the pr refs before the merge and github deletes them at merge.

PR #127, opened 2015-05-07
Add CholeskySolveVec method and auxiliary routines

btracey:


kortschak:
LGTM

mat64/cholesky.go
```diff
@@ -79,8 +79,7 @@ func (t *TriDense) Cholesky(a *SymDense, upper bool) (ok bool) {
 }
 
 // SolveCholesky finds the matrix x that solves A * X = B where A = L * L^T or
-// A = U^T * U, and U or L are represented by t. The matrix A must be symmetric
-// and positive definite.
+// A = U^T * U, and U or L are represented by t.
 func (m *Dense) SolveCholesky(t Triangular, b Matrix) {
```
vladimir-ch: Dense receivers are almost consistently called 'm' but for example in this case wouldn't it be better to relate better the function signature with the comment? The comment talks about x, X, A, B, L, U, while the function takes m, t, b.

mat64/cholesky.go
```diff
@@ -109,6 +108,31 @@ func (m *Dense) SolveCholesky(t Triangular, b Matrix) {
 	}
 }
 
+// SolveCholesky finds the vector x that solves A * x = b where A = L * L^T or
```
vladimir-ch: SolveCholeskyVec

mat64/vector.go
```diff
@@ -103,6 +105,15 @@ func (m *Vector) RawVector() blas64.Vector {
 	return m.mat
 }
 
+// Copy makes a copy of elements of a into the receiver. It is similar to the
```
vladimir-ch: CopyVec

mat64/cholesky.go
kortschak: Add ", placing the result in the receiver."

vladimir-ch: ... and make both x in the comment the same case.

mat64/cholesky.go
```diff
@@ -109,6 +108,31 @@ func (m *Dense) SolveCholesky(t Triangular, b Matrix) {
...
+// A = U^T * U, and U or L are represented by t.
```
vladimir-ch: ", placing the result in the receiver." also here?

kortschak: Yes, I think so.

mat64/vector.go
btracey: Done

mat64/cholesky.go
btracey: Done.

mat64/cholesky.go
btracey: Done.

mat64/cholesky.go
btracey: Done.

PR #128, opened 2015-05-07
Check Triangular match in Cholesky

btracey:
If the TriDense receiver in Cholesky is non-zero, the triangular storage should match the upper input argument. This change ensures this; previously the shape of the receiver was silently overwritten.

Fixes issue 126.


kortschak:
LGTM

PR #129, opened 2015-05-14
Use Lapack implementation of Cholesky implementation and add benchmarks

btracey:


btracey:
Sorry, the benchmark code is wrong. Working on fixing.

btracey:
This code is almost an order of magnitude faster that the previous implementation for large matrices
````
benchcmp bench_nolapack.bench bench_go_lapack.bench 
benchmark                   old ns/op      new ns/op     delta        
BenchmarkCholeskySmall      47.2           179           +279.24%     
BenchmarkCholeskyMedium     247397         180221        -27.15%      
BenchmarkCholeskyLarge      1287486048     98749708      -92.33%  
````

And if you're using cgo, about two orders of magnitude faster
````
benchcmp bench_nolapack.bench bench_cgo_lapack.bench 
benchmark                   old ns/op      new ns/op     delta         
BenchmarkCholeskySmall      47.2           853           +1707.20%     
BenchmarkCholeskyMedium     247397         126584        -48.83%       
BenchmarkCholeskyLarge      1287486048     7992787       -99.38%    
````

kortschak:
Is there any reason not to just use the s.mat.Datat for the t.mat.Data if the given t's data slice is not yet allocated?

Otherwise LGTM

Nice work.

btracey:
Potrf modifies the data in place, so using s.data in the call would change the input argument. Our functions leave the input arguments alone and only modify the receiver.

kortschak:
Fair enough.

PR #135, opened 2015-05-23
mat64: update outer product methods

kortschak:
@btracey

### DIFF
--- a/mat64/dense_arithmetic.go
+++ b/mat64/dense_arithmetic.go
@@ -895,11 +895,17 @@ func (m *Dense) EqualsApprox(b Matrix, epsilon float64) bool {
 	return true
 }
 
-// RankOne performs a rank-one update to the matrix b and stores the result
+// RankOne performs a rank-one update to the matrix a and stores the result
 // in the receiver
 //  m = a + alpha * x * y'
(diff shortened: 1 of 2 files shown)


btracey:
Referencing an offline discussion:

I think there is merit to having both outer and RankOne. Outer stores a new result, while RankOne updates a matrix. Even when RankOne is changed to not allocate, Outer could still be used to generate the original matrix (starting from a zero matrix), while RankOne has to have a non-zero matrix argument. 

I think cross-comments between the methods are useful.

btracey:
LGTM

kortschak:
Would you like these added? I'm not sure I understand.

btracey:
Off of list, you said 

" If we do that, we should add a sentence to the docs saying RankOne is the equivalent to outer explcitly, numpy e.g. calls it outer - so it would be easier for people to find if it at least mentioned that name."

I see that I misread it, and you meant if we didn't have Outer. I do think that it would be useful to have on the Outer method 
````
In order to update to an existing matrix, see RankOne
````
and for RankOne
````
If a is zero, see Outer.
````

kortschak:
I'll add those.

PR #137, opened 2015-05-27
mat64: make Inner take Vector type

kortschak:
This needs tests for Inc != 1. I'm thinking try all four of `[]struct{ incX, incY int }{{1, 1}, {1,2}, {2, 1}, {2, 2}}` in a loop with the given data. This needs then a helper to construct those vectors from the []float that are in the test struct.

### DIFF
--- a/mat64/inner.go
+++ b/mat64/inner.go
@@ -4,20 +4,23 @@
 
 package mat64
 
-import "github.com/gonum/internal/asm"
+import (
+	"github.com/gonum/blas"
+	"github.com/gonum/internal/asm"
+)
 
 // Inner computes the generalized inner product
 //   x^T A y
 // between vectors x and y with matrix A. This is only a true inner product if
 // A is symmetric positive definite, though the operation works for any matrix A.
 //
-// Inner panics if len(x) != m or len(y) != n when A is an m x n matrix.
-func Inner(x []float64, A Matrix, y []float64) float64 {
+// Inner panics if x.Len != m or y.Len != n when A is an m x n matrix.
+func Inner(x *Vector, A Matrix, y *Vector) float64 {
 	m, n := A.Dims()
-	if len(x) != m {
+	if x.Len() != m {
 		panic(ErrShape)
 	}
-	if len(y) != n {
+	if y.Len() != n {
 		panic(ErrShape)
 	}
 	if m == 0 || n == 0 {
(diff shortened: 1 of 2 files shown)


kortschak:
@btracey

mat64/inner.go
```diff
@@ -29,26 +29,65 @@ func Inner(x []float64, A Matrix, y []float64) float64 {
 	switch b := A.(type) {
 	case RawSymmetricer:
 		bmat := b.RawSymmetric()
```
btracey: This should probably be "case RawSymmetricer, RawTriangular", and then get the data slice and stride as variables. Also, we should check that the Uplo is correct.

btracey:
@kortschak , I think the functions I implemented in BLAS are nice for this. It puts weird unique values in the off-elements, which makes debugging easier when the code is incorrect.

mat64/inner.go
btracey: I guess it's more complicated with Triangular, because we support both shapes.

kortschak: Can't do `case RawSymmetricer, RawTriangular:`. We can do separate cases and send the results to a func.

kortschak:
I'll poison the off index elements with NaN.

mat64/inner.go
btracey: Yes, sorry, I meant that informally. Sorry for not being specific.

mat64/inner.go
```diff
@@ -11,13 +11,13 @@ import "github.com/gonum/internal/asm"
 // between vectors x and y with matrix A. This is only a true inner product if
 // A is symmetric positive definite, though the operation works for any matrix A.
 //
-// Inner panics if len(x) != m or len(y) != n when A is an m x n matrix.
-func Inner(x []float64, A Matrix, y []float64) float64 {
+// Inner panics if x.Len != m or y.Len != n when A is an m x n matrix.
```
btracey: Should these be "x.Len()" since it's a function and not a field?

kortschak:
Just read your comments there, I'll do that.

mat64/inner.go
kortschak: Conventionally no.

mat64/inner.go
kortschak: I think the Triangular handling is for another PR. I'll add the Uplo check though.

btracey:
The code in Inner looks correct to me

mat64/inner.go
btracey: SGTM

kortschak:
OK, PTAL

btracey:
LGTM

btracey:
Thanks

PR #140, opened 2015-05-27
mat64: make Exp mul calls work in workspace not receiver

kortschak:
@btracey 

While I was here I made things faster

```
name     old mean              new mean              delta
Exp10     102µs × (0.98,1.01)    70µs × (0.96,1.07)  -31.27% (p=0.000 n=8+10)
Exp100   14.0ms × (0.95,1.08)  13.1ms × (0.94,1.07)  -6.41% (p=0.019 n=10+10)
Exp1000   11.3s × (0.97,1.04)   11.4s × (0.97,1.03)     ~    (p=0.243 n=9+10)
```

and reduced allocations

```
benchmark            old allocs     new allocs     delta
BenchmarkExp10       138            128            -7.25%
BenchmarkExp100      204            189            -7.35%
BenchmarkExp1000     209            191            -8.61%

benchmark            old bytes     new bytes     delta
BenchmarkExp10       16608         6816          -58.96%
BenchmarkExp100      1236819       334686        -72.94%
BenchmarkExp1000     128064800     32021408      -75.00%
```

### DIFF
--- a/mat64/dense_arithmetic.go
+++ b/mat64/dense_arithmetic.go
@@ -646,6 +646,8 @@ func (m *Dense) Exp(a Matrix) {
 	if r != c {
 		panic(ErrShape)
 	}
+
+	var w *Dense
 	switch {
 	case m.isZero():
 		m.mat = blas64.General{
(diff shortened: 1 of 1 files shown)


mat64/dense_arithmetic.go
```diff
@@ -641,9 +644,10 @@ func (m *Dense) Exp(a Matrix) {
 		scaling = 4
 	)
 
-	var small, power Dense
+	small := &Dense{}
 	small.Scale(math.Pow(2, -scaling), a)
-	power.Clone(&small)
+	power := &Dense{}
```
btracey: The code is fine, but do you have guidelines as to when one should use
````
var a &Dense{}
````
as opposed to 
````
var a Dense
````
I don't feel like I've gotten the hang of it.

btracey:
LGTM

mat64/dense_arithmetic.go
kortschak: My preference is concrete over pointer for internal things unless it stos you doing what you need to,do. That's what happened here.

PR #141, opened 2015-05-27
mat64: fix {Mul,MulTrans} behaviour when receiver is also parameter

kortschak:
@btracey 

This was good. Found bugs with it.

### DIFF
--- a/mat64/dense_arithmetic.go
+++ b/mat64/dense_arithmetic.go
@@ -351,6 +351,7 @@ func (m *Dense) Dot(b Matrix) float64 {
 }
 
 // Mul takes the matrix product of a and b, placing the result in the receiver.
+// If a or b are also the receiver, temporary workspace memory will be allocated.
 //
 // See the Muler interface for more information.
 func (m *Dense) Mul(a, b Matrix) {
@@ -361,9 +362,12 @@ func (m *Dense) Mul(a, b Matrix) {
 		panic(ErrShape)
 	}
 
-	var w Dense
+	var w *Dense
 	if m != a && m != b {
-		w = *m
+		w = m
+	} else {
+		m.reuseAs(ar, bc)
+		w = &Dense{}
 	}
 	w.reuseAs(ar, bc)
 
@@ -371,7 +375,9 @@ func (m *Dense) Mul(a, b Matrix) {
 		if b, ok := b.(RawMatrixer); ok {
 			amat, bmat := a.RawMatrix(), b.RawMatrix()
 			blas64.Gemm(blas.NoTrans, blas.NoTrans, 1, amat, bmat, 0, w.mat)
-			*m = w
+			if w != m {
+				m.Copy(w)
+			}
 			return
 		}
 	}
@@ -389,7 +395,9 @@ func (m *Dense) Mul(a, b Matrix) {
 					)
 				}
 			}
-			*m = w
+			if w != m {
+				m.Copy(w)
+			}
 			return
 		}
 	}
@@ -407,11 +415,14 @@ func (m *Dense) Mul(a, b Matrix) {
 			w.mat.Data[r*w.mat.Stride+c] = v
 		}
 	}
-	*m = w
+	if w != m {
+		m.Copy(w)
+	}
 }
 
 // MulTrans takes the matrix product of a and b, optionally transposing each,
-// and placing the result in the receiver.
+// and placing the result in the receiver. If a or b are also the receiver,
+// temporary workspace memory will be allocated.
 //
 // See the MulTranser interface for more information.
 func (m *Dense) MulTrans(a Matrix, aTrans bool, b Matrix, bTrans bool) {
@@ -429,9 +440,12 @@ func (m *Dense) MulTrans(a Matrix, aTrans bool, b Matrix, bTrans bool) {
 		panic(ErrShape)
 	}
 
-	var w Dense
+	var w *Dense
 	if m != a && m != b {
-		w = *m
+		w = m
+	} else {
+		m.reuseAs(ar, bc)
+		w = &Dense{}
 	}
 	w.reuseAs(ar, bc)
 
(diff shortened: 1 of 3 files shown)


mat64/dense_arithmetic.go
```diff
@@ -351,6 +351,7 @@ func (m *Dense) Dot(b Matrix) float64 {
 }
 
 // Mul takes the matrix product of a and b, placing the result in the receiver.
+// If a or b are also the receiver, a workspace will be allocated.
 //
 // See the Muler interface for more information.
 func (m *Dense) Mul(a, b Matrix) {
```
btracey: It looks like this is missing a check that mr == ar and mc == bc

mat64/dense_arithmetic.go
```diff
@@ -361,9 +362,12 @@ func (m *Dense) Mul(a, b Matrix) {
 		panic(ErrShape)
 	}
 
-	var w Dense
+	var w *Dense
 	if m != a && m != b {
-		w = *m
+		w = m
+	} else {
+		m.reuseAs(ar, bc)
```
btracey: This shouldn't be necessary. Since m is the same as an input argument, it can only be ar x ac, or br x bc. Either the multiplication ends with that size, or the result should panic because the receiver is non-zero and the dimensions do not match. While the code you added in the inner test looks deceptively nice, I think it's more surprising to overwrite a mis-sized receiver.

mat64/dense_arithmetic.go
```diff
@@ -351,6 +351,7 @@ func (m *Dense) Dot(b Matrix) float64 {
 }
 
 // Mul takes the matrix product of a and b, placing the result in the receiver.
+// If a or b are also the receiver, a workspace will be allocated.
```
btracey: I don't like this phrasing of "a workspace will be allocated". I would prefer "temporary memory will be allocated to store the result" or something along those lines.

mat64/dense_arithmetic.go
```diff
@@ -407,11 +415,14 @@ func (m *Dense) Mul(a, b Matrix) {
...
+	}
 }
 
 // MulTrans takes the matrix product of a and b, optionally transposing each,
-// and placing the result in the receiver.
+// and placing the result in the receiver. If a or b are also the receiver, a
+// workspace will be allocated.
 //
 // See the MulTranser interface for more information.
 func (m *Dense) MulTrans(a Matrix, aTrans bool, b Matrix, bTrans bool) {
```
btracey: Same here, I think instead we should check that m has the right size if it's non-zero.

mat64/dense_arithmetic.go
kortschak: I think you've misunderstood what this is doing.

The call the reuseAs here is not to allocate or resize, but to check if it can be reused. We know that m is non-zero if the operation is at all valid (a and b are non-zero), so the only thing that can be done in the reuseAs call is to pass the size checks or not. We panic in the second case and do nothing in the first case.

I'm confused about what you mean wrt the change in the inner test. The inner test (and the svd test) were previously incorrect; we had a nxm receiver failing to panic when calling an operation that produced a nx1 and nx1 not panicing when calling a 1x1 operation. Without the change I made in the tests, those tests would now panic. That actually threw me for a bit, because I was not expecting this fix to cause shape panics.

btracey: You are right, I did misunderstand. Ignore the comment about the inner test. Sorry about that.

mat64/dense_arithmetic.go
kortschak: I don't think any mention of storing the result should be made, that confuses the matter, I will s/workspace/temporary workspace memory/. The text is purely there to prevent people from doing silly things like a.Mul(a, a) in a tight loop and wondering why things are slow.

Here might actually be a good place for pools though. I can dig that code out and make it work again.

btracey: We can just say "temporary memory will be allocated". The word "workspace" is confusing to me in this context. 

If/when you revive the pools, could you add more comments? I'm better at Go than when it was first proposed, but I remember not understanding what was going on.

btracey:
LGTM with the comment fix. Sorry for causing unnecessary confusion.

mat64/dense_arithmetic.go
kortschak: Can you explain why "workspace" is confusing? The reason I would like to include it is that it signifies a large allocation rather than just some memory being allocated. The term is used in other matrix code that is allocating (e.g csparse and sparseBLAS).

btracey: I haven't read the csparse or sparseBLAS documentation, so that may be the problem. When someone says "workspace", I think of something of bigger importance than allocating a []float64. For example, a workspace in Matlab is all of the variables that you have in memory. I've heard it used in python as a synonym for "environment", where it refers to the importable packages and where they live. Perhaps falsely, but I also associate the term with what's talked about in fftw and the CUDA documentation, where it is a memory allocation issue, but there are also a whole bunch of parameters you can tweak to get the behavior you want. In go, memory allocation is a relatively small deal with GC, and the important parts beyond the memory (size, shape, etc.) are handled within the function, not by the user. It matters for performance (I agree that the documentation is good), but 'workspace' to me implies something I need to care after and configure, while here we're just letting the user know that a significant memory allocation may happen.

btracey: and by "associate the term", I mean associating the term workspace with, for example 'plan' in fftw

kortschak: I think that in conjunction with "temporary" and the absence of any handle to a workspace clients should not over-read the word "workspace".

If we implement pools, this all becomes moot - which to me is a good reason for doing that.

btracey: Okay.

kortschak:
Travis is an absolute dog today.

PR #144, opened 2015-05-30
travis: always rebuild all in tests

kortschak:
This fixes the noasm weirdness. The proof that this works is at https://travis-ci.org/gonum/matrix/builds/64648387 and the explanation is at https://groups.google.com/d/topic/golang-nuts/y2mTZ3PAPYs/discussion

@btracey


btracey:
LGTM

PR #145, opened 2015-05-30
mat64: drop LUGaussian

kortschak:
@btracey 

Magic values checked with octave.

I'll add the # 107 when this is approved.


btracey:
LGTM.

PR #151, opened 2015-06-08
mat64: add slicing bounds checks to Dense

kortschak:
This is a sketch for how we might do slicing bounds checks. At this stage for discussion.

Updates #150.

@btracey Please take a look.

### DIFF
--- a/mat64/dense_arithmetic.go
+++ b/mat64/dense_arithmetic.go
@@ -100,7 +100,7 @@ func (m *Dense) Norm(ord float64) float64 {
 		}
 	case ord == 0:
 		for i := 0; i < len(m.mat.Data); i += m.mat.Stride {
-			for _, v := range m.mat.Data[i : i+m.mat.Cols] {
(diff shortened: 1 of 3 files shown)


btracey:
I like the idea. Do we think it would be useful to also have helper routines "sliceRow, sliceRowTri" and the like? It may help write less buggy code.

kortschak:
The intention was to have it for all, yes.

btracey:
Sorry, we may have communicated properly, but I'm going to restate what we said in case we did not.

I like the code that exists there, but I'm suggesting in addition we may want to have helper scripts. The one there now is for getting a general subset of a row. We could also provide 
````
sliceDenseRow(m blas64.General, i) []float64{
     return m.Data[i*m.Stride: i*m.Stride + m.Cols]
}
````
which helps some common cases.

kortschak:
Ahh, I missed that. Sure.

kortschak:
Closing due to staleness. Will send another in due course.

PR #152, opened 2015-06-11
cleanup .travis.yml and associated files

jonlawlor:
As explained in
https://groups.google.com/d/msg/gonum-dev/GKJjAtDxR9I/n0FD7WvvN5AJ

Removed notifications and go tip, simplified code, added some comments,
put all travis related code (except for .travis.yml) into .travis/

Currently matrix does not build against blas libs.

It would probably be nice to switch check.v1 to std lib testing?


kortschak:
LGTM

kortschak:
Yes, we will switch the testing back to not use check. The main issue is that it's not a small amount of work. There is probably a minimal set of eg transformations that will do it, but I haven't looked into that yet.

jonlawlor:
I'll take a look at it, I've wanted to do some code gen stuff.

kortschak:
Wonderful. Thank you.

jonlawlor:
I don't think it is possible with eg, unless we wanted to write our own gocheck anyway.  eg doesn't allow substitutions to have multiple statements, and things like `c.Check(left, check.Equals, right)` have to be replaced with a conditional.

kortschak:
Oh well. It would require a gofix module then - or just manual re-write.

Issue #154, opened 2015-07-15
Vec should have SetVec and should not have Set


btracey:
The set-like method for vector should be
````
func (vec *Vector) SetVec(i int, v float64) 
````
The general Set method that is there now must have 0 in the jth element. This enforces that to always be true.

PR #155, opened 2015-07-15
Change the Set method of Vector to SetVec. This enforces at compile t…

btracey:
…ime that the 'column' is zero

Fixes Issue 154.

### DIFF
--- a/mat64/index_bound_checks.go
+++ b/mat64/index_bound_checks.go
@@ -55,19 +55,17 @@ func (v *Vector) at(r int) float64 {
 	return v.mat.Data[r*v.mat.Inc]
 }
 
-// Set sets the element at (r,c) to the value val. It panics if c is not zero.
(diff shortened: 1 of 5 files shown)


btracey:
Fixes issue #154

btracey:
Good catch Travis.

kortschak:
Make the index label i instead of r?

btracey:
I almost changed it, but I thought this might help reinforce the idea that Vector is a column vector in other operations. Happy to go either way.

kortschak:
OK

btracey:
Is that the same as a LGTM?

kortschak:
Sure.

kortschak:
Sorry, I meant I accepted your explanation - leaving it as r is fine, but i is good too.

LGTM

btracey:
Changed to i. You're right, it's better.

Issue #123, opened 2015-04-19
mat64: add Tranpose type


kortschak:
In #121 [I proposed a new type Transpose](https://github.com/gonum/matrix/issues/121#issuecomment-94210532).

I suggest we add a `Transpose struct { Matrix }` with a Dims and At method that transposes the embedded matrix. In itself, this is not wildly useful, but it can be used internally to signal the transpose state to the Mul routine. This allows us to retire the MulTrans API, keeping its optimisations (we could keep it, but this is worth thinking about).

Identifying the transpose state would be done by a helper that walks the T chain until the underlying Dense (or other optimisable type) is found, keeping track of the state. Additionally, I think a helper T(Matrix) Matrix would be worthwhile to optimise this process:

```
func T(m Matrix) Matrix {
    if t, ok := m.(Transpose); ok {
        return t.Matrix
    }
    return Transpose{m}
}
```

btracey:
Reply is on the long side, just to set the stage of my argument.

This type seeks to solve a real problem. We have to deal with the fact that transpose is a common operation, and for many matrix methods the transpose can be dealt with implicitly instead of allocating and copying the transposed matrix. Our present solution to the problem makes an unnecessarily large API surface, and we can probably do better. 

In my opinion, any effective solution must:
1) Be easy and legible in the context of matrix operations
2) Be the same mechanism with all the matrix types
3) Cause minimal interference 

I'm worried about the complications in state that come from this. Take, for example,
````
a := mat64.NewDense(2,3, []float64{1,2,3,4,5,6})
aTrans := mat64.Transpose{a}
aTrans.Set(0, 1, 16)
````
This sets the {0,1} entry, which is definitely confusing to a reader (sure, it's terrible code, but we still need to think about what we enable). 

My counterproposal is to make the type solely for communicating the matrix should be transposed for the operation, and worthless otherwise. I would propose that the type be ````type T struct{Matrix}````. This would allow, for example, ````m.Mul(mat64.T{a}, mat64.T{b})````. I think this is more legible than ````m.Mul(mat64.Transpose{a}, mat64.Transpose{b})````, and since it's fewer characters it encourages use within the expression. 

Secondly, I would suggest that T does NOT have any overloaded methods, and instead includes some documentation saying as much and that its only purpose is to communicate the transposedness of the function argument. Yes, if someone does 
````
aTrans := mat64.T{a}
aTrans.At(0,1)
````
[...]

PR #156, opened 2015-07-16
Change Matrix definition to support transposes. Add Transpose type in…

btracey:
… support.

There is a need to support matrix transposes, for example the ability to perform A^T * B in addition to A * B. One way to support this is to add input booleans specifying if the matrix is transposed, as is done now in MulTrans. This does not scale well for many arguments and functions. This change adds a T() method to Matrix to support implicit transposing of matrices. This allows user to call, for example, Mul(A.T(), B) to perform A^T * B.

Please see the discussion in 123 for more information.

Changes to actual methods in order to efficiently support this new method type will come in future commits.

Fixes issue #123.

### DIFF
--- a/mat64/matrix.go
+++ b/mat64/matrix.go
@@ -22,6 +22,59 @@ type Matrix interface {
 	// At returns the value of a matrix element at (r, c). It will panic if r or c are
 	// out of bounds for the matrix.
 	At(r, c int) float64
+
+	// T returns the transpose of the Matrix. Whether T returns a copy of the
+	// underlying data is implementation dependent.
+	// This method may be implemented using the Transpose type, which
+	// provides an implicit matrix transpose.
+	T() Matrix
+}
+
+var (
+	_ Matrix       = Transpose{}
+	_ Untransposer = Transpose{}
+)
+
+// Transpose is a type for performing an implicit matrix transpose. It implements
+// the Matrix interface, returning values from the transpose of the matrix within.
+type Transpose struct {
+	Matrix Matrix
+}
+
+// At returns the value of the element at row i and column j of the transposed
+// matrix, that is, row j and column i of the Matrix field.
+func (t Transpose) At(i, j int) float64 {
+	return t.Matrix.At(j, i)
+}
+
+// Dims returns the dimensions of the transposed matrix. The number of rows returned
+// is the number of columns in the Matrix field, and the number of columns is
+// the number of rows in the Matrix field.
+func (t Transpose) Dims() (r, c int) {
+	c, r = t.Matrix.Dims()
+	return r, c
+}
+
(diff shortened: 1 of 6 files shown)


kortschak:
Remove the Transposer interface since it's no longer needed - and is now misinformative.

Add a comment to the T method in Matrix that the in-place behaviour is implementation dependent and add that the T and TTri methods return an in-place transpose of the receiver.

Also add method comments for the Matrix methods of the Transpose and TransposeTri.

Then LGTM.

btracey:
I originally used the words "in-place", but I thought that was confusing. In place could mean that the data is shuffled around in the same way. It seems like implicit is better than in place. Comments?

As far as the Matrix interface goes, I think all we want to require is that Matrix methods work properly. If it wants to be the same type (like Symmetric) that's fine. If it wants to be a new type where changes could be a problem (like Dense and TriDense) that's fine. If an external implementation wants to generate a whole new matrix and return that, that should be fine too. I wasn't sure how to word this lack of requirement in the documentation

kortschak:
"Whether T returns a copy of the underlying data is implementation dependent."

btracey:
PTAL

mat64/matrix.go
```diff
@@ -22,6 +22,59 @@ type Matrix interface {
 	// At returns the value of a matrix element at (r, c). It will panic if r or c are
 	// out of bounds for the matrix.
 	At(r, c int) float64
+
+	// T returns the transpose of the Matrix. The manner in which this is implemented
+	// is implementation dependent, including if T returns a copy of the underlying
+	// data. This function may be implemented using the Transpose type, which
```
kortschak: s/function/method/

mat64/matrix.go
```diff
@@ -22,6 +22,59 @@ type Matrix interface {
 	// At returns the value of a matrix element at (r, c). It will panic if r or c are
 	// out of bounds for the matrix.
 	At(r, c int) float64
+
+	// T returns the transpose of the Matrix. The manner in which this is implemented
+	// is implementation dependent, including if T returns a copy of the underlying
```
kortschak: I don't see the need for "The manner in which this is implemented is implementation dependent"; it is tautological. The only thing that needs to be conveyed is whether the transpose will reflect changes to the original matrix after the call to T. I think the original wording I have above does this. I also thing "if" reads less well than "whether".

mat64/matrix.go

kortschak:
LGTM

btracey:
Thanks for leading the charge on the change. A bunch of work to implement, but it's going to make interacting with the package much nicer.

kortschak:
The easiest thing to do will be Mul, since it involves just dropping Mul and slightly modifying MulTrans. The rest, not so much.

PR #157, opened 2015-07-17
Fixtestadd

btracey:


btracey:
This is working toward improving the mat64 testing capability.

The problem with the package is that we have a lot of types that could/should have special casing for efficiency. Testing all of these special cases in every package would be a big burden. This adds a single function for testing ANY method of the form Matrix.Operation(Matrix, Matrix) (or that can be wrapped in that form). To demonstrate this, I added this check to the Add methods for SymDense and Dense (and they pass, woo!).

This testing function is a bit of a doozy, but it is very powerful. It requires the ability to emulate what the method does for Dense matrices. Handwritten tests will be needed for Dense.Op(Dense, Dense). For any other case (including all non-Dense matrix types), all of the tests are effectively automatically generated. In particular, note that testTwoInputs
1) Tests combinations of Matrix types
2) Certain methods require particular input classes. If the matrix doesn't work properly (for example Dense doesn't implement Symmetric), that case is skipped.
3) Tests the transpose cases
4) Tests the cases where the underlying pointer for the receiver and the inputs match. 
5) Tests a variety of sizes.

btracey:
This isn't done. Lots of comments to be fixed. Also need to add in the tests for some viewed matrices.

btracey:
Note to self: The vector/transpose issue can be fixed by having makeRandOf wrap in a Transpose for row matrices, and then Transpose can do the unwrapping or wrapping.

btracey:
Added all of the "basic" types to the testing matrix and implemented the vector/transpose fix.

btracey:
PTAL.

btracey:
I'm not sure how to do "sameType" correctly. I would have thought the correct code is 
````
if reflect.ValueOf(a).Type().Name() == reflect.ValueOf(b).Type().Name() {
	return true
}
return false
````
But the Name is the empty string always, and I don't understand why.

**Comment by kortschak** [comment]
Why not compare the reflect.Types directly?

**Comment by kortschak** [comment]
Which types are you seeing "" as the name?

**Comment by btracey** [comment]
All of them. I added fmt.Println(reflect.ValueOf(a).Type().Name()) to the top of the file, and it prints a bunch of newlines

**Comment by btracey** [comment]
Although using String() works just fine

**Comment by kortschak** [comment]
If you printf with %T and print typ.Name() what do you see?

File: mat64/list_test.go [hunk 1]
```diff
@@ -0,0 +1,706 @@
...
+	case Transpose:
+		v := t.Matrix.(*Vector)
+		m := makeCopyOf(v)
+		return m
+	}
+}
+
+// sameType returns true if a and b have the same underlying type.
+func sameType(a, b Matrix) bool {
+	if reflect.ValueOf(a).Type().String() == reflect.ValueOf(b).Type().String() {
```
kortschak: Don't do this with a string comparison; you can compare reflect.Types returned by Type() (it's a pointer).

btracey: Done. Thanks.

btracey:
Ping.

mat64/list_test.go
```diff
@@ -0,0 +1,706 @@
...
+			if a.At(i, j) != b.At(i, j) {
+				return false
+			}
+		}
+	}
+	return true
+}
+
+// testTwoInput tests a method that has two input arguments.
+// receiver is a value of the receiver type.
```
kortschak: The docs here are confusing; there is a claim that method returns false for some condition, but this is clearly not true.

I'm going to suggest something that I've never seen before, but I think will work here: make parameter doc comments look like a public struct definition, e.g.

```
// testTwoInput tests a method that has two input arguments.
func testTwoInput(c *check.C,
	// receiver is a value of the receiver type.
	receiver Matrix,

	// method is the generalied receiver.Method(a,b).
	method func(receiver, a, b Matrix),

	// denseComparison performs the same operation
	// as method, but with dense matrices for
	// comparison with the result.
	denseComparison func(receiver, a, b *Dense),

	// legalTypes returns whether the concrete types
	// in Matrix are valid for the method.
	legalTypes func(a, b Matrix) bool,

	// legalSize returns whether the matrix sizes are
	// valid for the method.
	legalSize func(ar, ac, br, bc int) bool,
) {
    ...
```

mat64/list_test.go
```diff
@@ -0,0 +1,706 @@
...
+	legalTypes func(a, b Matrix) bool,
+	legalSize func(ar, ac, br, bc int) bool,
+) {
+	strideVec := &Vector{
+		mat: blas64.Vector{
+			Inc: 10,
+		},
+	}
+	// Loop over all of the matrix types.
+	matrices := []Matrix{
```
kortschak: s/matrices/types/g

mat64/list_test.go
```diff
@@ -0,0 +1,706 @@
...
+				{
+					br, bc := b.Dims()
+					if br != test.br || bc != test.bc {
+						panic("makeRandOf returns wrong size" + fmt.Sprintf("%T\n", bMat))
+						continue
+					}
+				}
+
+				// Compute the true answer if the sizes are legal.
+				legal := legalSize(test.ar, test.ac, test.br, test.bc)
```
kortschak: s/legal/dimsOK/g

mat64/list_test.go
```diff
@@ -0,0 +1,706 @@
...
+				{17, 6, 6, 11},
+				{11, 6, 17, 6},
+				{17, 6, 11, 6},
+				{11, 17, 6, 6},
+				{17, 11, 6, 6},
+			} {
+				// Skip the test if the arguments are compile-time provably incorrect
+				// This is either because the matrix types are wrong at compile time
+				// or because the sizes are impossible
+				ok := legalTypes(aMat, bMat)
```
kortschak: In these cases just use the construction `if !legalX(...` since ok is not used again.

mat64/list_test.go
```diff
@@ -0,0 +1,706 @@
...
+				{6, 11, 17, 6},
+				{6, 17, 11, 6},
+				{11, 6, 6, 17},
+				{17, 6, 6, 11},
+				{11, 6, 17, 6},
+				{17, 6, 11, 6},
+				{11, 17, 6, 6},
+				{17, 11, 6, 6},
+			} {
+				// Skip the test if the arguments are compile-time provably incorrect
```
kortschak: I don't understand why you are using "compile-time provably". The assertions are runtime checks.

mat64/list_test.go
```diff
@@ -0,0 +1,706 @@
...
+				// This is either because the matrix types are wrong at compile time
+				// or because the sizes are impossible
+				ok := legalTypes(aMat, bMat)
+				if !ok {
+					continue
+				}
+				if ok := dimsLegal(aMat, test.ar, test.ac); !ok {
```
kortschak: s/dimsLegal/legalDims/g

mat64/list_test.go
```diff
@@ -0,0 +1,706 @@
...
+					continue
+				}
+				if ok := dimsLegal(bMat, test.br, test.bc); !ok {
+					continue
+				}
+				a := makeRandOf(aMat, test.ar, test.ac)
+				{
+					ar, ac := a.Dims()
```
kortschak: Here and below, put the Dims call into the if statement.

mat64/list_test.go
```diff
@@ -0,0 +1,706 @@
...
+					if ar != test.ar || ac != test.ac {
+						panic("makeRandOf returns wrong size" + fmt.Sprintf("%T\n", aMat))
```
kortschak: One of these is wrong - do you want to panic or continue. You can't do both.

kortschak: You can `panic(fmt.Sprintf("makeRandOf for %T returns wrong size: %dx%d != %dx%d", aMat, ar, ac, test.ar, test.ac))`. This is more informative and cleaner than appending a string to the output of fmt.Sprintf.

mat64/list_test.go
```diff
@@ -0,0 +1,706 @@
...
+					br, bc := b.Dims()
+					if br != test.br || bc != test.bc {
+						panic("makeRandOf returns wrong size" + fmt.Sprintf("%T\n", bMat))
+						continue
+					}
+				}
+
+				// Compute the true answer if the sizes are legal.
+				legal := legalSize(test.ar, test.ac, test.br, test.bc)
+				var truth Dense
```
kortschak: s/truth/want/g

mat64/list_test.go
```diff
@@ -0,0 +1,706 @@
...
+					bDense.Clone(b)
+					denseComparison(&truth, &aDense, &bDense)
+				}
+				// Call the method and check the result with all transpose variations.
+				// Do this by explicitly create a transpose, and then calling T().
+				for _, aTrans := range []bool{true, false} {
+					for _, bTrans := range []bool{true, false} {
+						aCopy := makeCopyOf(a)
+						bCopy := makeCopyOf(b)
+						// Set the proper function call
```
kortschak: Period.

mat64/list_test.go
```diff
@@ -0,0 +1,706 @@
...
+				{11, 6, 6, 17},
+				{17, 6, 6, 11},
+				{11, 6, 17, 6},
+				{17, 6, 11, 6},
+				{11, 17, 6, 6},
+				{17, 11, 6, 6},
+			} {
+				// Skip the test if the arguments are compile-time provably incorrect
+				// This is either because the matrix types are wrong at compile time
+				// or because the sizes are impossible
```
kortschak: Period.

mat64/list_test.go
```diff
@@ -0,0 +1,706 @@
...
+		strideVec,
+		&basicMatrix{},
+		&basicVectorer{},
+		&basicSymmetric{},
+		&basicTriangular{},
+	}
+
+	for _, aMat := range matrices {
+		for _, bMat := range matrices {
+			// Loop over all of the size combinations (bigger, smaller, etc.)
```
kortschak: Period.

mat64/list_test.go
```diff
@@ -0,0 +1,706 @@
...
+						// Set the proper function call
+						var f func(receiver, a, b Matrix)
+						if aTrans {
+							if bTrans {
+								f = func(receiver, a, b Matrix) {
+									aT := makeTransposeCopyOf(a)
+									r, c := a.Dims()
+									tr, tc := aT.Dims()
+									if r != tc || c != tr {
+										panic("makeTransposeCopyOf returns wrong size" + fmt.Sprintf("%T", a))
```
kortschak: Here and beyond, see comment above about the panic formating.

mat64/list_test.go
```diff
@@ -0,0 +1,706 @@
...
+				}
+				// Call the method and check the result with all transpose variations.
+				// Do this by explicitly create a transpose, and then calling T().
+				for _, aTrans := range []bool{true, false} {
+					for _, bTrans := range []bool{true, false} {
+						aCopy := makeCopyOf(a)
+						bCopy := makeCopyOf(b)
+						// Set the proper function call
+						var f func(receiver, a, b Matrix)
+						if aTrans {
```
kortschak: It seems to me that this can be flattened into two conditions and a final call to method using a transient pair of Matrix values. This abolishes the need for 2 nested conditionals.

mat64/list_test.go
```diff
@@ -0,0 +1,706 @@
...
+									r, c = b.Dims()
+									tr, tc = bT.Dims()
+									if r != tc || c != tr {
+										panic("makeTransposeCopyOf returns wrong size" + fmt.Sprintf("%T", b))
+									}
+									method(receiver, aT.T(), bT.T())
+								}
+							} else {
+								f = func(receiver, a, b Matrix) {
+									aT := makeTransposeCopyOf(a)
```
kortschak: Why are the same checks for a transposition done above not done here?

mat64/list_test.go
```diff
@@ -0,0 +1,706 @@
...
+									method(receiver, aT.T(), b)
+								}
+							}
+						} else {
+							if bTrans {
+								f = func(receiver, a, b Matrix) {
+									bT := makeTransposeCopyOf(b)
```
kortschak: Same here?

mat64/list_test.go
```diff
@@ -0,0 +1,706 @@
...
+
+						if !legal {
+							continue
+						}
+
+						// Test the method with a non-zero-value of the receiver.
+						// The receiver has been overwritten in place so we can
+						// take its new size.
+						rr, rc := zero.Dims()
+						nonzero := makeRandOf(receiver, rr, rc)
```
kortschak: s/nonzero/nonZero/g

CAMELS!

mat64/list_test.go
```diff
@@ -0,0 +1,706 @@
...
+									method(receiver, a, bT.T())
+								}
+							} else {
+								f = func(receiver, a, b Matrix) {
+									method(receiver, a, b)
+								}
+							}
+						}
+						// Test the method for a zero-value of the receiver.
+						errStr := fmt.Sprintf("%T(%T, %T), sizes: %#v, atrans %v, btrans %v", receiver, a, b, test, aTrans, bTrans)
```
kortschak: Add the method name into this: "%T.%s(%T, %T)...". This needs to be threaded in as another parameter to testTwoInput, since it's not possible to determine this by reflection AFAICS.

mat64/list_test.go
```diff
@@ -0,0 +1,706 @@
...
+						// take its new size.
+						rr, rc := zero.Dims()
+						nonzero := makeRandOf(receiver, rr, rc)
+						p, err = panics(func() { f(nonzero, a, b) })
+						if p {
+							c.Errorf("Panicked with non-zero receiver: " + errStr)
+						}
+
+						if !equalApprox(nonzero, &truth, 1e-14) {
+
```
kortschak: Delete empty line.

mat64/list_test.go
```diff
@@ -0,0 +1,706 @@
...
+						if !legal {
+							continue
+						}
+
+						// Test the method with a non-zero-value of the receiver.
+						// The receiver has been overwritten in place so we can
+						// take its new size.
+						rr, rc := zero.Dims()
+						nonzero := makeRandOf(receiver, rr, rc)
+						p, err = panics(func() { f(nonzero, a, b) })
```
kortschak: s/p/panicked/g

mat64/list_test.go
```diff
@@ -0,0 +1,706 @@
...
+
+						if !equalApprox(nonzero, &truth, 1e-14) {
+
+							c.Errorf("Answer mismatch non-zero receiver: " + errStr)
+						}
+
+						// Test with an incorrectly sized matrix
+						switch receiver.(type) {
+						default:
+							panic("matrix type not coded")
```
kortschak: Can this be a little more informative?

mat64/symmetric_test.go
```diff
@@ -119,6 +119,15 @@ func (s *S) TestSymAdd(c *check.C) {
 			}
 		}
 	}
+
+	method := func(receiver, a, b Matrix) {
+		rd := receiver.(*SymDense)
```
kortschak: Add `type symAdder interface { AddSym(a, b Symmetric) }` then s/.(*SymDense)/.(symAdder)/.

This is superficial, but goes to indicating that this is a generic case.

mat64/list_test.go
```diff
@@ -0,0 +1,706 @@
...
+func sameType(a, b Matrix) bool {
+	if reflect.ValueOf(a).Type() == reflect.ValueOf(b).Type() {
+		return true
+	}
+	return false
+}
+
+// canShadow returns true if the two matrices could be represented by the same
+// pointer.
+func canShadow(receiver, a Matrix, trans bool) bool {
```
kortschak: It would be nice to know what the logic is behind this test. It seems pretty loose to me; presumably the domain being tested has some invariants that make the false positives and negatives that are here not be a problem.

mat64/list_test.go
```diff
@@ -0,0 +1,706 @@
...
+		return true
+	}
+)
+
+// dimsLegal returns true if {m,n} is a valid dimension of the given matrix type.
+func dimsLegal(a Matrix, m, n int) bool {
+	switch a.(type) {
+	default:
+		panic("not coded")
+	case *Dense, *basicMatrix, *basicVectorer:
```
kortschak: I had always assumed this wasn't legal - good to know about. Thanks.

mat64/list_test.go
```diff
@@ -0,0 +1,706 @@
...
+	"fmt"
+	"math/rand"
+	"reflect"
+
+	"github.com/gonum/blas/blas64"
+	"github.com/gonum/floats"
+	"gopkg.in/check.v1"
+)
+
+var (
```
kortschak: Make these vars funcs.

mat64/list_test.go
btracey: Done.

mat64/list_test.go
btracey: Done.

mat64/list_test.go
btracey: You can't do anything with the type (in t := a.(type)), but it does work.

mat64/list_test.go
btracey: There have been some bugs discovered when performing operations where the receiver is the same as one of the input arguments, i.e. a.Mul(a,b). "canShadow" determines if it's possible with this combination of underlying Matrix types and sizes for that to be true. In other words, for multiply, if all the types are dense, and it's a 4x4 times a 4x4, it could be of the form
````
a.Mul(a,a)
a.Mul(a,b)
a.Mul(b,a)
c.Mul(a,b)
````
If instead b is a Symmetric, then only
````
a.Mul(a,b)
c.Mul(a,b)
````
are possible.

File: mat64/list_test.go (hunk 3 above)
Comment (btracey) [inline mat64/list_test.go]: Done.

File: mat64/list_test.go (hunk 13 above)
Comment (btracey) [inline mat64/list_test.go]: Done.

File: mat64/list_test.go (hunk 6 above)
Comment (btracey) [inline mat64/list_test.go]: The comment was talking about the call
````
sym.Mul(sym,dense)
````
This could never happen in user code because Dense does not implement Symmetric. It's compile time impossible for those matrix types to be used. Similarly, the type/size loops above could call for a 17x6 symmetric matrix. This is impossible; the symmetric type must be square.

Comment (btracey) [inline mat64/list_test.go]: This is different from legalSize. In a matrix multiply,  {m,k} x {l,n}  k != l is not a valid size for the inputs, but k == l is not forced at compile time.

Comment (btracey) [inline mat64/list_test.go]: Probably legalSize --> validSize in addition to dimsLegal --> legalDims. Legal = would be compile time enforced, valid = would be runtime enforced.

File: mat64/list_test.go (hunk 5 above)
Comment (btracey) [inline mat64/list_test.go]: Oh, I didn't realize you could ! the result from a function like that. Changed.

File: mat64/list_test.go (hunk 6 above)
Comment (btracey) [inline mat64/list_test.go]: Let me know how to improve the comments.

File: mat64/list_test.go (hunk 12 above)
Comment (btracey) [inline mat64/list_test.go]: Done.

File: mat64/list_test.go (hunk 8 above)
Comment (btracey) [inline mat64/list_test.go]: Took your later suggestion.

File: mat64/list_test.go (hunk 9 above)
Comment (btracey) [inline mat64/list_test.go]: Much better, thanks.

File: mat64/list_test.go (hunk 6 above)
Comment (btracey) [inline mat64/list_test.go]: legalSize --> dimsOK as you suggest below.

File: mat64/list_test.go (hunk 7 above)
Comment (btracey) [inline mat64/list_test.go]: Done.

File: mat64/list_test.go (hunk 4 above)
Comment (btracey) [inline mat64/list_test.go]: done.

File: mat64/list_test.go (hunk 10 above)
Comment (btracey) [inline mat64/list_test.go]: Done.

File: mat64/list_test.go (hunk 11 above)
Comment (btracey) [inline mat64/list_test.go]: Done.

File: mat64/list_test.go (hunk 15 above)
Comment (btracey) [inline mat64/list_test.go]: Totally right. Thanks.

File: mat64/list_test.go (hunk 14 above)
Comment (btracey) [inline mat64/list_test.go]: Done.

File: mat64/list_test.go (hunk 16 above)
Comment (btracey) [inline mat64/list_test.go]: Because it has been changed to make the check in makeTransposeOf where it should be.

File: mat64/list_test.go (hunk 17 above)
Comment (btracey) [inline mat64/list_test.go]: Done.

File: mat64/list_test.go (hunk 19 above)
Comment (btracey) [inline mat64/list_test.go]: Done.

File: mat64/list_test.go (hunk 18 above)
Comment (btracey) [inline mat64/list_test.go]: Sorry! Nonzero is one word in my mind. I forget gonum treats it as hyphenated. Done.

File: mat64/list_test.go (hunk 21 above)
Comment (btracey) [inline mat64/list_test.go]: Done.

What's the convention with moving this into the if statement? 
````
if panicked, _ := panic(); panicked{
}
````
On the one hand, panicked is ephemeral, but on the other hand nonZero is being meaningfully modified in the function call.

File: mat64/list_test.go (hunk 20 above)
Comment (btracey) [inline mat64/list_test.go]: Done.

File: mat64/list_test.go (hunk 22 above)
Comment (btracey) [inline mat64/list_test.go]: Done.

File: mat64/symmetric_test.go (hunk 23 above)
Comment (btracey) [inline mat64/symmetric_test.go]: Done, but I made it addSymer.

**Comment by btracey** [comment]
PTAL

File: mat64/list_test.go (hunk 21 above)
Comment (kortschak) [inline mat64/list_test.go]: Ugliness/aesthetics.

File: mat64/list_test.go (hunk 24 above)
Comment (kortschak) [inline mat64/list_test.go]: Sorry, I mean what is the logic used within the test that makes it a valid test for potential shadowing. It seems to me that it has false positives in the general case, but we we are not doing the general case (no views here). This should be explained in the docs for the function - the text above would make a good basis for that explanation - with the domain of cases where the test will be valid.

I also think you should rename from canShadow to couldShadow to indicate the looseness of the check. To illustrate this, if we have `a := make([]int, 100); b = a[50:60]` then elements from `a` and `b` can shadow each other. However you are making a test about types `a := make([]int, 100); b := make([]int, 100)` - in principle if you knew nothing about `a` and `b` other than their type and length you could claim that they might shadow each other.

File: mat64/list_test.go (hunk 25 above)
Comment (kortschak) [inline mat64/list_test.go]: You can use them, but only as the type of a.

File: mat64/dense_test.go [hunk 27]
```diff
@@ -397,6 +405,15 @@ func (s *S) TestAdd(c *check.C) {
 		c.Check(a.Equals(r), check.Equals, true, check.Commentf("Test %d: %v Add %v expect %v got %v",
 			i, test.a, test.b, test.r, unflatten(a.mat.Rows, a.mat.Cols, a.mat.Data)))
 	}
+
+	method := func(receiver, a, b Matrix) {
+		rd := receiver.(*Dense)
```
kortschak: s/*Dense/Adder/

Sorry, I thought I had written this one before, but must have missed it.

mat64/list_test.go
```diff
@@ -0,0 +1,684 @@
...
+	"fmt"
+	"math/rand"
+	"reflect"
+
+	"github.com/gonum/blas/blas64"
+	"github.com/gonum/floats"
+	"gopkg.in/check.v1"
+)
+
+// The sizes are legal when the two matrices have the same rectangular shape.
```
kortschak: Make these func comments follow the doc comment style, "dimsOKSameRectangular returns whether ...".

mat64/list_test.go
```diff
@@ -0,0 +1,684 @@
...
+	if _, ok := a.(Symmetric); !ok {
+		return false
+	}
+	if _, ok := b.(Symmetric); !ok {
+		return false
+	}
+	return true
+}
+
+// legalDims returns true if {m,n} is a valid dimension of the given matrix type.
```
kortschak: s/true if/whether/

mat64/list_test.go
```diff
@@ -0,0 +1,684 @@
...
+			panic("bad type")
+		case *TriDense:
+			return mat
+		case *basicTriangular:
+			return asBasicTriangular(mat)
+		}
+	}
+}
+
+// makeRandOf returns a new random matrix of the underlying matrix type.
```
kortschak: s/new random matrix/new randomly filled n×m matrix/

mat64/list_test.go
```diff
@@ -0,0 +1,684 @@
...
+		}
+		matrix = returnAs(mat, t)
+	}
+	if mr, mc := matrix.Dims(); mr != m || mc != n {
+		panic(fmt.Sprintf("makeRandOf for %T returns wrong size: %d%d != %d%d", a, m, n, mr, mc))
+	}
+	return matrix
+}
+
+// makeCopyOf makes a copy of the matrix.
```
kortschak: s/makes/returns/

mat64/list_test.go
kortschak: OK. So the kind of wording that you should use would be modeled on how reflect describes this kind of situation. You are mimicking the compile-time checks at runtime. The subjunctive is perfect for this:

"Skip the test if any argument would not be assignable to the method's corresponding input parameter or it is not possible to construct an argument of the requested size."

mat64/list_test.go
```diff
@@ -0,0 +1,684 @@
...
+					continue
+				}
+				if !legalDims(bMat, test.br, test.bc) {
+					continue
+				}
+				a := makeRandOf(aMat, test.ar, test.ac)
+				b := makeRandOf(bMat, test.br, test.bc)
+
+				// Compute the true answer if the sizes are legal.
+				legal := dimsOK(test.ar, test.ac, test.br, test.bc)
```
kortschak: I think there was confusion here. The func should be legalDims and the status var should dimsOK. legal by itself doesn't have enough context to span this many lines.

mat64/list_test.go
```diff
@@ -0,0 +1,684 @@
...
+							if bTrans {
+								bCall = makeTransposeCopyOf(b).T()
+							}
+							method(receiver, aCall, bCall)
+						}
+
+						// Test the method for a zero-value of the receiver.
+						errStr := fmt.Sprintf("%T.%s(%T, %T), sizes: %#v, atrans %v, btrans %v", receiver, name, a, b, test, aTrans, bTrans)
+						zero := makeRandOf(receiver, 0, 0)
+						p, err := panics(func() { f(zero, a, b) })
```
kortschak: s/p/panicked/g

mat64/list_test.go
```diff
@@ -0,0 +1,684 @@
...
+							if !panicked {
+								c.Errorf("Did not panic with wrong number of rows: " + errStr)
+							}
+							wrongSize = makeRandOf(receiver, rr, rc+1)
+							panicked, _ = panics(func() { f(wrongSize, a, b) })
+							if !panicked {
+								c.Errorf("Did not panic with wrong number of columns: " + errStr)
+							}
+						case *TriDense, *SymDense:
+							// Only work for squares
```
kortschak: Full sentence please.

mat64/list_test.go
```diff
@@ -0,0 +1,684 @@
...
+							if !panicked {
+								c.Errorf("Did not panic with wrong number of rows: " + errStr)
+							}
+						}
+
+						// TODO(btracey): Add test for Issue 147 if it is agreed that should be fixed.
+
+						// If the type and size of an input and ouput match, check
+						// that the function still works when one of them is the
+						// receiver
```
kortschak: Period.

mat64/symmetric_test.go
```diff
@@ -76,6 +80,10 @@ func (s *S) TestSymAtSet(c *check.C) {
 	c.Check(t.At(1, 2), check.Equals, 12.0)
 }
 
+type addSymer interface {
```
kortschak: Since this is only used in the definition of method below, it can be defined there in the func.

mat64/symmetric_test.go
```diff
@@ -119,6 +127,15 @@ func (s *S) TestSymAdd(c *check.C) {
...
 	}
+
+	method := func(receiver, a, b Matrix) {
+		rd := receiver.(addSymer)
+		rd.AddSym(a.(Symmetric), b.(Symmetric))
+	}
+	denseComparison := func(receiver, a, b *Dense) {
+		receiver.Add(a, b)
+	}
+	testTwoInput(c, "Add", &SymDense{}, method, denseComparison, legalTypesSym, dimsOKSameSquare)
```
kortschak: s/Add/AddSym/

mat64/list_test.go
```diff
@@ -0,0 +1,684 @@
...
+	}
+	if mr, mc := matrix.Dims(); mr != c || mc != r {
+		panic(fmt.Sprintf("makeTransposeOf for %T returns wrong size: %d%d != %d%d", a, c, r, mr, mc))
+	}
+	return matrix
+}
+
+// sameType returns true if a and b have the same underlying type.
+func sameType(a, b Matrix) bool {
+	if reflect.ValueOf(a).Type() == reflect.ValueOf(b).Type() {
```
kortschak: `return reflect.ValueOf(a).Type() == reflect.ValueOf(b).Type()`

mat64/list_test.go
```diff
@@ -0,0 +1,684 @@
...
+	case *Vector:
+		m := makeCopyOf(a)
+		matrix = Transpose{m}
+	case Transpose:
+		v := t.Matrix.(*Vector)
+		m := makeCopyOf(v)
+		matrix = m
+	}
+	if mr, mc := matrix.Dims(); mr != c || mc != r {
+		panic(fmt.Sprintf("makeTransposeOf for %T returns wrong size: %d%d != %d%d", a, c, r, mr, mc))
```
kortschak: s/%d%d/%d×%d/g

mat64/list_test.go
```diff
@@ -0,0 +1,684 @@
...
+			for i := 0; i < m; i++ {
+				for j := 0; j <= i; j++ {
+					mat.SetTri(i, j, rand.Float64())
+				}
+			}
+		}
+		matrix = returnAs(mat, t)
+	}
+	if mr, mc := matrix.Dims(); mr != m || mc != n {
+		panic(fmt.Sprintf("makeRandOf for %T returns wrong size: %d%d != %d%d", a, m, n, mr, mc))
```
kortschak: s/%d%d/%d×%d/g

mat64/list_test.go
```diff
@@ -0,0 +1,684 @@
...
+	"math/rand"
+	"reflect"
+
+	"github.com/gonum/blas/blas64"
+	"github.com/gonum/floats"
+	"gopkg.in/check.v1"
+)
+
+// The sizes are legal when the two matrices have the same rectangular shape.
+func dimsOKSameRectangular(ar, ac, br, bc int) bool {
```
kortschak: I liked the names of these before.

mat64/list_test.go
```diff
@@ -0,0 +1,684 @@
...
+		return false
+	}
+	if ac != bc {
+		return false
+	}
+	return true
+}
+
+// The sizes are legal when the two matrices have the same square shape.
+func dimsOKSameSquare(ar, ac, br, bc int) bool {
```
kortschak: Same.

mat64/dense_test.go
btracey: Ah, I see. Yes. Sorry.

mat64/list_test.go
btracey: Done.

mat64/list_test.go
btracey: Done.

mat64/list_test.go
btracey: Done.

mat64/list_test.go
btracey: Done.

mat64/list_test.go
btracey: Yes. Thanks.

mat64/list_test.go
btracey: Done.

mat64/list_test.go
btracey: Done.

mat64/list_test.go
btracey: The logic is 
1) Are the types the same size
2) Are the dimensions the same. 
The extra layer of complexity comes from the fact it could be A.T() in one of the calls.

You're right that it doesn't test most instances of shadowing, but we don't support any complicated instances of shadowing (nor should we). The check in matrix methods is on the matrix pointer, not on the overlap between data slices. We do support A.Op(A,B) and the like, which is what is attempting to be tested. Would "could share pointer" be a better name?

btracey: * couldSharePointer

btracey: I'll also clean up that method (return instead of the ifs)

mat64/list_test.go
btracey: Got it. Reverted.

mat64/list_test.go
btracey: Thanks. Much better.

mat64/list_test.go
btracey: Sorry. Got all of them this time.

mat64/list_test.go
kortschak: maybeSame gets across what we want with the appropriate level of uncertainty.

btracey: SGTM.

mat64/list_test.go
btracey: Fixing this part of the code now. It doesn't do what it says it does for the transpose cases.

mat64/symmetric_test.go
btracey: Huh. I've somehow never seen that before. Thanks.

kortschak: Orthogonality is cool.

mat64/list_test.go
```diff
@@ -0,0 +1,675 @@
...
+							}
+							method(receiver, aCall, bCall)
+						}
+
+						// Test the method for a zero-value of the receiver.
+						errStr := fmt.Sprintf("%T.%s(%T, %T), sizes: %#v, atrans %v, btrans %v", receiver, name, a, b, test, aTrans, bTrans)
+						zero := makeRandOf(receiver, 0, 0)
+						panicked, err := panics(func() { f(zero, a, b) })
+						if !dimsOK && !panicked {
+							c.Errorf("Did not panic with illegal size: " + errStr)
```
kortschak: Throughout use a verb when you use Errorf.

mat64/symmetric_test.go
```diff
@@ -119,6 +123,19 @@ func (s *S) TestSymAdd(c *check.C) {
 			}
 		}
 	}
+
+	type addSymer interface {
```
kortschak: I was thinking in the minimal scope (in the definition of `method`), but this is fine.

btracey:
Huh. Well, what do you know, when you write correct tests, you expose issues. 

Right now, the following code doesn't work
````
a.Add(a.T(), b) 
````
Add does not recognize the fact that the receiver and the first input argument share a matrix pointer (within the transpose). Thus, as the matrices are added, the data gets corrupted. 

I'm not sure what we want to do about this. I think we either want to allow the same arguments or we don't. That is, I think if we allow a.Add(a, b), we should also allow a.Add(a.T(), b). There's a reasonable argument to be made that we shouldn't allow the same matrices. It's nominally to save memory, but in a lot of cases we allocate memory anyway. It's thus surprising that, in a loop for instance, it's better to allocate a new receiver outside the loop (as in mat.Pow()). We could enforce this by panicking when the pointers are the same. This would also resolve 147 (since sharing is prohibited). Alternately, we could allow the receiver matching, I'll fix Add, and keep fixing the maybeSame tests.

File: mat64/list_test.go (hunk 43 above)
Comment (btracey) [inline mat64/list_test.go]: Done.

File: mat64/symmetric_test.go (hunk 44 above)
Comment (btracey) [inline mat64/symmetric_test.go]: Done.

**Comment by kortschak** [comment]
That's not surprising. The current approach we use to test whether we are working in the same place is to test pointers. The pointers and types are not the same when a matrix is wrapped in a transpose, so we need to write a helper that does the proper test and replace the equality test throughout.

I don't think that we should panic for performance related errors; it cuts off development paths and will impact on uptake. We do what is asked - which in this case is to allocate a new matrix. When people see a performance hit, they should profile and they will see that they did something silly.

**Comment by btracey** [comment]
I wasn't arguing on performance, that's just as a side effect. We can either say "you can't work in the same place" or "you can work in the same place". The two have different costs. I'll fix the Add tests and keep fixing the shadowing code.

**Comment by kortschak** [comment]
What we should be saying is that the reality is that if you do unpaired transpose (and some other operations, e.g. Pow, Exp, Mul etc.) you can't work in the same place, but we will make it look like you can. This probably warrants a section in the docs.

**Comment by btracey** [comment]
I agree. Similarly, when you do certain operations (cholesky decomposition) on non-mat64 matrix types they may be marshaled into a form we can use (SymDense)

**Comment by btracey** [comment]
Okay, fixed the maybeSame tests. I tried to make minimal changes to the Add method such that it still works (there are a bunch of optimizations that can be done). With this way of dealing with Transpose we are on the road to fixing #147, so I added that as part of the tests. I do think the package is much easier conceptually if the memory you put into a Dense/Vector is modified in place rather than replaced sometimes and not other times.

**Comment by kortschak** [comment]
Just to clarify, when I use the term "same place" above I am referring to the data slice.

File: mat64/dense.go [hunk 45]
```diff
@@ -113,6 +113,31 @@ func (m *Dense) reuseAs(r, c int) {
...
+	if ut, ok := a.(Untransposer); ok {
+		return ut.Untranspose(), true
+	}
+	return a, false
+}
+
+// uniqueMemoryDense returns m with different memory from a. It also returns
+// a callback to defer for cleanup at the return of the call if necessary.
+func uniqueMemoryDense(m *Dense, a Matrix) (*Dense, func()) {
+	if m != a {
```
kortschak: I think this should not be here. Explanation below.

mat64/dense_arithmetic.go
```diff
@@ -124,12 +124,21 @@ func (m *Dense) Norm(ord float64) float64 {
...
 	if ar != br || ac != bc {
 		panic(ErrShape)
 	}
 
+	aMat, _ := untranspose(a)
+	bMat, _ := untranspose(b)
 	m.reuseAs(ar, ac)
+	var f func()
+	m, f = uniqueMemoryDense(m, aMat)
+	if f == nil {
```
kortschak: You need to do a pointer equality test either here only (above the call to uniqueMemoryDense test m == aMat) or both in uniqueMemoryDense and here). I think the clarity is reduced by doing it this way as well.

mat64/dense_arithmetic.go
```diff
@@ -124,12 +124,21 @@ func (m *Dense) Norm(ord float64) float64 {
...
 	br, bc := b.Dims()
-
 	if ar != br || ac != bc {
 		panic(ErrShape)
 	}
 
+	aMat, _ := untranspose(a)
+	bMat, _ := untranspose(b)
 	m.reuseAs(ar, ac)
+	var f func()
```
kortschak: s/f/replace/g

mat64/dense.go
```diff
@@ -113,6 +113,31 @@ func (m *Dense) reuseAs(r, c int) {
...
+func untranspose(a Matrix) (Matrix, bool) {
+	if ut, ok := a.(Untransposer); ok {
+		return ut.Untranspose(), true
+	}
+	return a, false
+}
+
+// uniqueMemoryDense returns m with different memory from a. It also returns
+// a callback to defer for cleanup at the return of the call if necessary.
+func uniqueMemoryDense(m *Dense, a Matrix) (*Dense, func()) {
```
kortschak: I'm not happy with the name, but I can't think of something better right now.

mat64/dense_arithmetic.go
```diff
@@ -124,12 +124,21 @@ func (m *Dense) Norm(ord float64) float64 {
 func (m *Dense) Add(a, b Matrix) {
 	ar, ac := a.Dims()
 	br, bc := b.Dims()
-
 	if ar != br || ac != bc {
 		panic(ErrShape)
 	}
 
+	aMat, _ := untranspose(a)
```
kortschak: I don't understand how this can be correct.

mat64/dense.go
btracey: I'll remove the Dense at least. Throughout mat64 if unspecified it implies Dense.

mat64/dense_arithmetic.go
btracey: Okay. I was trying to make a minimal resuable set, but I see your point. I think what you're suggesting is:
````
m.reusaAs(ar,as)
var f func()
if m == a {
    m, f = uniqueMemory(m, aMat)
    defer f()
} else if m == b {
    m, f = uniqueMemory(m, bMat)
    defer f()
}
````
Correct?

mat64/dense_arithmetic.go
btracey: I don't like replace. It talks of switching instead of re-put-in-place. How about "restore"?

mat64/dense_arithmetic.go
btracey: I don't know what you mean. I'm pretty sure the untranspose code is correct. The rest of the Add function uses the original matrices a and b, rather than aMat and bMat. So, the actual code path followed is the final one as a and b are neither RawMatrixers or Vectorers. A better implementation of Add would be to use aMat so that we can get the RawMatrixer fast path, and take into account the actual transposes. I didn't do that in order to minimize the surface of this commit.

kortschak: Sorry, I missed that you are using a and b. All good then.

mat64/dense_arithmetic.go
kortschak: Much better.

mat64/dense_arithmetic.go
kortschak: Yes, that's right.

mat64/dense.go
kortschak: Still not convinced, I'll keep thinking.

btracey: independentData is another set of words. Still not great, but will maybe help us brainstorm.

mat64/dense_arithmetic.go
btracey: This change causes the tests to fail. I'm trying to figure out what the problem is.

btracey: Can you run a single test with gocheck? I want to do the equivalent of go test -run=TestAdd

btracey: Sorry, found the documentation.

btracey: Ah. Found the error. I blindly copied the code above, but the comparison should be with aMat and bMat, not a and b.

mat64/dense.go
btracey: Done.

mat64/dense_arithmetic.go
btracey: Done.

mat64/dense_arithmetic.go
btracey: Done.

mat64/dense_arithmetic.go
btracey: Yea, just trying to keep as much as possible from the original Add function so this commit is mostly about list_test.

mat64/dense.go
kortschak: Make it a method on *Dense and call it workspaceWith.

btracey: I like making it a method and using the word workspace, but I don’t like workspaceWith. The “With” doesn’t imply anything. Maybe “isolateWorkspace” or something like that?

kortschak: I agree. "With" was intended to indicate that you are working with the
second matrix, I was leaning towards "Considering" but didn't like
that. 

isolatedWorkspace and you have a deal - isolateWorkspace suggests
working in place (which we can't do).

Also add named returns, w and restore purely for documentation.

btracey:
Yes, sorry, I misread your message the first time. Fixed.

btracey:
Oops. Fixed/improved comment for isolatedWorkspace.

kortschak:
This LGTM if you are ready.

btracey:
Thanks for the review.

PR #161, opened 2015-07-29
Modify MulVec and add tests for Dense.Mul

btracey:
This begins the adaptation for the new T() method. With this change, fix testTwoInput to work properly with Vector receivers, and significantly simplify the code

### DIFF
--- a/mat64/dense_test.go
+++ b/mat64/dense_test.go
@@ -647,6 +647,20 @@ func (s *S) TestMul(c *check.C) {
 		temp.mat.Data = nil
 		c.Check(func() { temp.Mul(a, b) }, check.PanicMatches, "blas: index of c out of range", check.Commentf("Test %d", i))
 	}
+
+	method := func(receiver, a, b Matrix) {
+		rd := receiver.(Muler)
+		rd.Mul(a, b)
+	}
+	denseComparison := func(receiver, a, b *Dense) {
+		receiver.Mul(a, b)
+	}
+
+	legalSizeMul := func(ar, ac, br, bc int) bool {
+		return ac == br
+	}
+
+	testTwoInput(c, "Mul", &Dense{}, method, denseComparison, legalTypesAll, legalSizeMul)
 }
 
 func (s *S) TestMulTrans(c *check.C) {
@@ -1504,7 +1518,7 @@ func denseMulTransBench(b *testing.B, size int, rho float64) {
 	b.StartTimer()
 	for i := 0; i < b.N; i++ {
 		var n Dense
-		n.MulTrans(a, false, d, true)
+		n.Mul(a, d.T())
 		wd = &n
 	}
 }
@@ -1521,7 +1535,7 @@ func denseMulTransSymBench(b *testing.B, size int, rho float64) {
 	b.StartTimer()
 	for i := 0; i < b.N; i++ {
 		var n Dense
-		n.MulTrans(a, false, a, true)
+		n.Mul(a, a.T())
 		wd = &n
 	}
 }
--- a/mat64/list_test.go
+++ b/mat64/list_test.go
@@ -55,11 +55,20 @@ func legalTypesSym(a, b Matrix) bool {
 	return true
 }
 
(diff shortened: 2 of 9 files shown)


btracey:
PTAL @kortschak 

The main change to testTwoInput is to move the transposed types into the matrix list. The old way did not work properly with Vector receivers because of the Transpose magic. This code is much simpler. The Transpose types are added explicitly in the list rather than with a loop to make it easier to isolate single failure examples.

I can delete MulTrans and the relative test if you would like. I left them in for now until the full functionality is replaced (next PR).

In general I don't like deleting tests, but I feel that the MulVec tests were doing a less sophisticated version of testTwoInput, so I removed them.

btracey:
Oh, right. Two questions
1) I can implement Vector pools in this commit if you'd like, or wait until the next one.
2) It is commented that stride needs to be zero for the matrix to be zero. Is the same true for Vector (with inc == 0), or is testing for zero length sufficient to be isZero? The old test had been on Inc, but I don't fully remember why.

kortschak:
It would be nicer if this were in a number of smaller more directed commits (fix tests; modify mulvec; delete old tests), but that would difficult to get to from here presumably.

Add pools after this.

The must is an invariant must, not a should must. The stride is only ever zero when rows, cols (and the caps) are set to zero either by a call to Reset (note the comment there) or because the Dense is newly allocated. If you require such an invariant then you can do it, but length should be fine - I chose stride because a zero strided matrix is never valid.

Oh, right. Two questions
1) I can implement Vector pools in this commit if you'd like, or wait until the next one.
2) It is commented that stride needs to be zero for the matrix to be zero. Is the same true for Vector (with inc == 0), or is testing for zero length sufficient to be isZero? The old test had been on Inc, but I don't fully remember why.

mat64/dense_test.go
```diff
@@ -647,6 +647,20 @@ func (s *S) TestMul(c *check.C) {
...
 	}
+
+	method := func(receiver, a, b Matrix) {
+		rd := receiver.(Muler)
+		rd.Mul(a, b)
+	}
+	denseComparison := func(receiver, a, b *Dense) {
+		receiver.Mul(a, b)
+	}
+
```
kortschak: Delete these two empty lines?

mat64/list_test.go
```diff
@@ -119,12 +125,31 @@ func returnAs(a, t Matrix) Matrix {
 	}
 }
 
+// retranspose returns the matrix m inside an Untransposer of the type
+// of a.
+func retranspose(a, m Matrix) Matrix {
+	_, ok := a.(Untransposer)
```
kortschak: This can be done more elegantly.

```
switch a.(type) {
case Transpose:
    return Transpose{m}
case TransposeTri:
    return TransposeTri{m.(Triangular)}
case Untransposer:
    panic("unknown transpose type")
default:
    panic("not a transposer")
}
```

mat64/list_test.go
```diff
@@ -395,6 +374,8 @@ func testTwoInput(c *check.C,
 			Inc: 10,
 		},
 	}
+	_ = strideVec // This is here to avoid unused variable errors when debugging.
```
kortschak: Will this go away later?

mat64/vector.go
```diff
@@ -188,48 +188,30 @@ func (v *Vector) DivElemVec(a, b *Vector) {
...
 }
 
-// MulVec computes a * b if trans == false and a^T * b if trans == true. The
-// result is stored into the receiver. MulVec panics if the number of columns in
-// a does not equal the number of rows in b.
-func (v *Vector) MulVec(a Matrix, trans bool, b *Vector) {
-	ar, ac := a.Dims()
+// MulVec computes a * b. The result is stored into the receiver.
+// MulVec panics if the number of columns in a does not equal the number of rows in b.
+func (v *Vector) MulVec(aMat Matrix, b *Vector) {
```
kortschak: aMat?

btracey:
I'll make a new PR.

btracey:
Changes proposed in stages.

PR #162, opened 2015-07-29
Fix and simplify testTwoInput.

btracey:
The current testTwoInput does not work properly with *Vector methods, such as MulVec, due to the magic performed getting Transpose{*Vector}. Rather than add yet more special cases, instead move the Transposed matrices into the types list. As an added bonus, this allows great simplification to the code as the two transpose loops can be deleted.

### DIFF
--- a/mat64/list_test.go
+++ b/mat64/list_test.go
@@ -55,11 +55,20 @@ func legalTypesSym(a, b Matrix) bool {
 	return true
 }
 
+// legalTypesNotVecVec returns whether the first input is an arbitrary Matrix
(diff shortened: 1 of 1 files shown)


btracey:
The plan is to leave the "_ = strideVec" line. It's nice to be able to comment out large parts of the types list without also needing to add and remove the same line every time. If you really prefer I can remove it.

kortschak:
LGTM if you add to the comment exactly the purpose of the blank assignment - from your last comment, subselecting types (your word will undoubtedly be better).

PR #163, opened 2015-07-29
Update Dense.Mul and test.

btracey:
Update Dense.Mul to use the new pattern for isolating a workspace and test the dense.Mul code with testTwoInput


kortschak:
LGTM

PR #165, opened 2015-08-02
Clean up Vector MulVec code, add pools for Vector, and add special ca…

btracey:
…se for vector-vector multiplication


kortschak:
LGTM

PR #166, opened 2015-08-02
Add Special cases to Dense.Mul

btracey:
There are many special cases in multiplying two matrices where significant efficiencies can be gained. This PR incorporates the efficiencies that had been in MulTrans and adds the remaining cases where one of the input arguments is a Dense matrix


btracey:
listTwoTest can be annoying to debug, but it does catch a lot of bugs.

kortschak:
LGTM

PR #168, opened 2015-08-03
Delete MulTrans.

btracey:
It is redundant now that Mul correctly deals with fast paths.


kortschak:
LGTM

We need to go through gonum/... and find uses of it.

btracey:
Yep. That's next.

btracey:
Could you do graph if there are any uses?

PR #170, opened 2015-08-04
Refactor LUFactor into LU, update Solve, and add specific SolveLU rou…

btracey:
…tines

Updates Issue 169

### DIFF
--- a/mat64/lu.go
+++ b/mat64/lu.go
@@ -7,206 +7,194 @@ package mat64
 
 import (
 	"math"
+
+	"github.com/gonum/blas"
+	"github.com/gonum/blas/blas64"
+	"github.com/gonum/lapack/lapack64"
 )
 
-type LUFactors struct {
-	LU    *Dense
-	Pivot []int
-	Sign  int
+// LU is a type for creating and using the LU factorization of a matrix.
+type LU struct {
+	lu    *Dense
+	pivot []int
 }
 
-// LU performs an LU Decomposition for an m-by-n matrix a.
+// Factorize computes the LU factorization of the square matrix a and stores the
+// result. The LU decomposition will complete regardless of the singularity of a.
 //
-// If m >= n, the LU decomposition is an m-by-n unit lower triangular matrix L,
-// an n-by-n upper triangular matrix U, and a permutation vector piv of length m
-// so that A(piv,:) = L*U.
-//
-// If m < n, then L is m-by-m and U is m-by-n.
-//
-// The LU decompostion with pivoting always exists, even if the matrix is
-// singular, so LU will never fail. The primary use of the LU decomposition
-// is in the solution of square systems of simultaneous linear equations. This
-// will fail if IsSingular() returns true.
-func LU(a *Dense) LUFactors {
-	// Use a "left-looking", dot-product, Crout/Doolittle algorithm.
-	m, n := a.Dims()
-	lu := a
-
-	piv := make([]int, m)
-	for i := range piv {
-		piv[i] = i
-	}
-	sign := 1
-
-	// Outer loop.
-	luColj := make([]float64, m)
-	for j := 0; j < n; j++ {
-
-		// Make a copy of the j-th column to localize references.
-		for i := 0; i < m; i++ {
-			luColj[i] = lu.at(i, j)
-		}
-
-		// Apply previous transformations.
-		for i := 0; i < m; i++ {
-			luRowi := lu.RawRowView(i)
-
-			// Most of the time is spent in the following dot product.
-			kmax := min(i, j)
-			var s float64
-			for k, v := range luRowi[:kmax] {
-				s += v * luColj[k]
-			}
-
-			luColj[i] -= s
-			luRowi[j] = luColj[i]
-		}
-
-		// Find pivot and exchange if necessary.
-		p := j
-		for i := j + 1; i < m; i++ {
-			if math.Abs(luColj[i]) > math.Abs(luColj[p]) {
-				p = i
-			}
-		}
-		if p != j {
-			for k := 0; k < n; k++ {
-				t := lu.at(p, k)
-				lu.set(p, k, lu.at(j, k))
-				lu.set(j, k, t)
-			}
-			piv[p], piv[j] = piv[j], piv[p]
-			sign = -sign
-		}
-
-		// Compute multipliers.
-		if j < m && lu.at(j, j) != 0 {
-			for i := j + 1; i < m; i++ {
-				lu.set(i, j, lu.at(i, j)/lu.at(j, j))
-			}
-		}
-	}
-
-	return LUFactors{lu, piv, sign}
-}
-
-// IsSingular returns whether the the upper triangular factor and hence a is
-// singular.
-func (f LUFactors) IsSingular() bool {
-	lu := f.LU
-	_, n := lu.Dims()
-	for j := 0; j < n; j++ {
-		if lu.at(j, j) == 0 {
-			return true
-		}
-	}
-	return false
-}
-
-// L returns the lower triangular factor of the LU decomposition.
-func (f LUFactors) L() *Dense {
-	lu := f.LU
-	m, n := lu.Dims()
-	if m < n {
-		n = m
-	}
-	l := NewDense(m, n, nil)
-	for i := 0; i < m; i++ {
-		for j := 0; j < n; j++ {
-			if i > j {
-				l.set(i, j, lu.at(i, j))
-			} else if i == j {
-				l.set(i, j, 1)
-			}
-		}
-	}
-	return l
-}
-
-// U returns the upper triangular factor of the LU decomposition.
-func (f LUFactors) U() *Dense {
-	lu := f.LU
-	m, n := lu.Dims()
-	u := NewDense(m, n, nil)
-	for i := 0; i < m; i++ {
-		for j := 0; j < n; j++ {
-			if i <= j {
-				u.set(i, j, lu.at(i, j))
-			}
-		}
-	}
-	return u
-}
-
-// Det returns the determinant of matrix a decomposed into lu. The matrix
-// a must have been square.
-func (f LUFactors) Det() float64 {
-	lu, sign := f.LU, f.Sign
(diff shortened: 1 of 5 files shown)


btracey:
Do you think LU.Factorize should return a boolean? Seems the proper time to handle the error is at the Solve function. One can use the factorization in the other functions successfully (det, etc.) even if the matrix is singular.

btracey:
Thinking about it, we need a better solution to Permutation. It's confusing that it works backward on the indices. I'm not sure what the best way to resolve this issue is. Invert the pivots as returned by LU? Have Permutation and PermutationInv?

kortschak:
>Do you think LU.Factorize should return a boolean? Seems the proper time to handle the error is at the Solve function. One can use the factorization in the other functions successfully (det, etc.) even if the matrix is singular.

If we save effort by knowing early that the factorisation was of a singular matrix then yes, otherwise no.

>Thinking about it, we need a better solution to Permutation. It's confusing that it works backward on the indices.

Is this just a consequence of the way LAPACK represents permutation matrices?

>I'm not sure what the best way to resolve this issue is. Invert the pivots as returned by LU? Have Permutation and PermutationInv?

What are you suggesting is the difference in behaviour between these two?

mat64/lu.go
```diff
@@ -6,207 +6,189 @@
...
-	l := NewDense(m, n, nil)
-	for i := 0; i < m; i++ {
-		for j := 0; j < n; j++ {
-			if i > j {
-				l.set(i, j, lu.at(i, j))
-			} else if i == j {
-				l.set(i, j, 1)
-			}
+// ExtractL extracts the lower triangular matrix from an LU factorization.
+func (t *TriDense) ExtractL(lu *LU) {
```
kortschak: Bikeshedding: `LFrom`.

mat64/lu.go
```diff
@@ -6,207 +6,189 @@
...
-	lu := f.LU
-	m, n := lu.Dims()
-	u := NewDense(m, n, nil)
-	for i := 0; i < m; i++ {
-		for j := 0; j < n; j++ {
-			if i <= j {
-				u.set(i, j, lu.at(i, j))
-			}
+// ExtractU extracts the upper triangular matrix from an LU factorization.
+func (t *TriDense) ExtractU(lu *LU) {
```
kortschak: Bikeshedding: `UFrom`.

mat64/lu_test.go
```diff
@@ -5,88 +5,108 @@
...
+		b := NewDense(n, bc, nil)
+		for i := 0; i < n; i++ {
+			for j := 0; j < bc; j++ {
+				b.Set(i, j, rand.NormFloat64())
+			}
 		}
-		u := lf.U()
-		if t.u != nil {
-			c.Check(u.Equals(t.u), check.Equals, true)
+		lu := &LU{}
```
kortschak: var lu LU

mat64/matrix.go
```diff
@@ -359,7 +361,9 @@ func Det(a Matrix) float64 {
 	if a, ok := a.(Deter); ok {
 		return a.Det()
 	}
-	return LU(DenseCopyOf(a)).Det()
+	lu := &LU{}
```
kortschak: var lu LU (previous comment was supposed to be here - lu is not used as a non-receiver parameter).

mat64/matrix.go
```diff
@@ -379,11 +383,14 @@ func Inverse(a Matrix) (*Dense, error) {
 func Solve(a, b Matrix) (x *Dense, err error) {
 	switch m, n := a.Dims(); {
 	case m == n:
-		lu := LU(DenseCopyOf(a))
-		if lu.IsSingular() {
+		lu := &LU{}
```
kortschak: var lu LU - not used as non-receiver parameter.

btracey:
>>Do you think LU.Factorize should return a boolean? Seems the proper time to handle the error is at the Solve function. One can use the factorization in the other functions successfully (det, etc.) even if the matrix is singular.

>If we save effort by knowing early that the factorisation was of a singular matrix then yes, otherwise no.

I don't think we save any work. As the PR stands right now, we can save the work of computing the determinant. It's generally trivial, as it is O(n) relative to the O(n^3) of the full operation. If we allow the user to modify LU in any way (such as rank-one updates), we will have to add the determinant computation back in anyway. Furthermore, if we adopt the lapack/matlab/scipy suggestion of checking for the condition number, then we'll have to do that in the solve anyway.

This also argues that solves should return an error. We can have ErrSingular which is when the matrix is exactly singular (U(i,i) == 0), and ErrNearSingular which provide the condition number warning. 

For reference
Scipy discussion: https://github.com/numpy/numpy/issues/3755
Julia Discussion: https://github.com/JuliaLang/julia/issues/4286
Lapack function for condition number given LU factorization: http://www.netlib.org/lapack/explore-html/db/de4/dgecon_8f.html#a188b8d30443d14b1a3f7f8331d87ae60

I don't see an equivalent for QR; it looks like lapack assumes least-squares solutions are always okay.

>>Thinking about it, we need a better solution to Permutation. It's confusing that it works backward on the indices.

>Is this just a consequence of the way LAPACK represents permutation matrices?

No, it's that when you form the factorization rows can swapped in a certain order. In order to form the correct P you need to do the "unswapping" step, which is running the indexes backward.

>>I'm not sure what the best way to resolve this issue is. Invert the pivots as returned by LU? Have Permutation and PermutationInv?

>What are you suggesting is the difference in behaviour between these two?

Basically, one way is the "forward algorithm" where you loop over the indices and swap the row. There is then the inverse of that, where you swap looping over the indices backward. The documentation seems clearest in the forward case "P is a matrix that will swap row 0 with pivot[0], then row 1 with pivot[1], etc.". If that is indeed the behavior, then lu.Pivot() should return lu.pivot reversed.

mat64/lu.go
btracey: Done.

mat64/lu.go
btracey: Done.

mat64/matrix.go

mat64/matrix.go

kortschak:
>I don't think we save any work. As the PR stands right now, we can save the work of computing the determinant. It's generally trivial, as it is O(n) relative to the O(n^3) of the full operation. If we allow the user to modify LU in any way (such as rank-one updates), we will have to add the determinant computation back in anyway. Furthermore, if we adopt the lapack/matlab/scipy suggestion of checking for the condition number, then we'll have to do that in the solve anyway.

>This also argues that solves should return an error. We can have ErrSingular which is when the matrix is exactly singular (U(i,i) == 0), and ErrNearSingular which provide the condition number warning. 

Agreed. Make a type Condition float64 that satisfies error and return that. This allows both cases to be encapsulated. Methods can be added for convenience: IsSingular, e.g.

>>Is this just a consequence of the way LAPACK represents permutation matrices?

>No, it's that when you form the factorization rows can swapped in a certain order. In order to form the correct P you need to do the "unswapping" step, which is running the indexes backward.

This is essentially what I mean.

>Basically, one way is the "forward algorithm" where you loop over the indices and swap the row. There is then the inverse of that, where you swap looping over the indices backward. The documentation seems clearest in the forward case "P is a matrix that will swap row 0 with pivot[0], then row 1 with pivot[1], etc.". If that is indeed the behavior, then lu.Pivot() should return lu.pivot reversed.

Making the pivot the right order for iterating forward makes sense and providing good documentation. Adding API should be a last resort here.

mat64/lu.go
```diff
@@ -6,207 +6,189 @@
...
-	}
-	l := NewDense(m, n, nil)
-	for i := 0; i < m; i++ {
-		for j := 0; j < n; j++ {
-			if i > j {
-				l.set(i, j, lu.at(i, j))
-			} else if i == j {
-				l.set(i, j, 1)
-			}
+// ExtractL extracts the lower triangular matrix from an LU factorization.
```
kortschak: Update doc, and below.

btracey:
The change proposed for Permutation is wrong. Just inverting the errors doesn't fix it. I instead changed the API to be more general and as a bonus, also correct.

I didn't fix the Solve errors. The tests need big changes. Rather than change everything twice I'll just clean up Solve when it becomes a method on Dense.

I put the Condition documentation with the type so that the documentation could be shared across all of the functions. If QR finds the matrix is exactly singular, the decomposition must stop, so we cannot guarantee that a solve is always finished. We can possibly guarantee that the solve will always stop if Condition == inf, but the current documentation is a hedge until we're sure. I don't think it's good policy to stop if the condition number is large and finite, as in many cases things turn out okay anyway. Proper error handling would check that things are okay based on whatever measure the user thinks is important.

btracey:
Sorry, Permute has some issues, fixing now.

btracey:
Done.

btracey:
Fixes some documentation problems. Also allows us to see if the build is actually fixed.

btracey:
Oh darn. Still bad stack pointer.

btracey:
Oh, no, sorry, that's with beta3. My bad.

kortschak:
You need to rebase to see the effect of master on the push.

btracey:
Yea.

Woo! Checks pass!

kortschak:
The seem more complicated a type than necessary. A single slice with swap positions should be enough - following the cycle is not difficult. I think we had code in at some stage that did this and if not, there certainly is some in biogo.matrix that you can use.

btracey:
The problem with Permute is that you can't just invert the order of the slice, that's not correct. You have to swap row n-1 first, then row n-2 (not 0 first then 1). We could have a "backward" boolean to run in the reverse direction. This allows for an arbitrary number of swaps (which is both good and bad).

btracey:
One could also just run the effective swap algorithm in lu and see where the 1s end up, then reconstruct the necessary pivots.

kortschak:
This last feels better.

btracey:
Algorithm updated. PTAL.

btracey:
@kortschak is this better?

kortschak:
Absolutely. I sent an LGTM by mail this morning, but it obviously didn't get through.

PR #172, opened 2015-08-06
mat64: check for zero sized parameter in Copy and CopySym

### LINKED ISSUE #171: SymDense copy Panicks on zero valued matrix
Right now, the following code panics with "blas64.Symmetric not upper".
````
s := &mat64.SymDense{}
var s2 mat64.SymDense
s2.Copy(s)
````
The panic is that s has 0 for the blas.Uplo field, and not blas.Upper (or blas.Lower for that matter). 

I assume this should not panic and we should only check for invariants when the size is not zero. Assuming that is the correct behavior, we should probably update testTwoInput to give the real zero value for zero size. Wouldn't catch this bug, but the future testOneInput would catch this.

kortschak:
Fixes #171.

@btracey Please take a look.

### DIFF
--- a/mat64/dense.go
+++ b/mat64/dense.go
@@ -417,6 +417,9 @@ func (m *Dense) Copy(a Matrix) (r, c int) {
 	r, c = a.Dims()
 	r = min(r, m.mat.Rows)
 	c = min(c, m.mat.Cols)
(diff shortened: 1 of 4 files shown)


btracey:
LGTM

PR #173, opened 2015-08-06
mat64: use guaranteed zero values in test matrix

kortschak:
New functions may at some stage perform operations that make a returned Matrix not actually a zero value. We say we allow zero values, so make sure these will always be tested.

Mentioned in CopySym panic issue comment.

@btracey Please take a look.

### DIFF
--- a/mat64/list_test.go
+++ b/mat64/list_test.go
@@ -376,8 +376,11 @@ func testTwoInput(c *check.C,
 	types := []Matrix{
 		&Dense{},
 		&SymDense{},
+		NewSymDense(0, nil),
+		&TriDense{},
 		NewTriDense(0, true, nil),
 		NewTriDense(0, false, nil),
+		&Vector{},
 		NewVector(0, nil),
 		strideVec,
 		&basicMatrix{},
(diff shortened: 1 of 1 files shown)


btracey:
Not LGTM

This change doesn't have the intended effect. The types list is used to generate values of the given size. In particular, makeRandOf reflects on the type and then makes a value of the size however it wants. You could give any matrix of the particular type and the behavior wouldn't change. The exception is that the Vector cases were specialized to construct strided vectors. Similarly, we should probably do some specialization to check for matrix views  (so, allocate a 10x 12 matrix that is a view of a 20x 20 or something). The different zero values don't have an effect, nor do I think they should.

Instead, I think it's better to have an explicit zero loop

````
for _, aMat := range types {
   for _, bMat := range types {
         for _, test := range []struct {
               ar, ac, br, bc
          }{
              {0, 0, 0,0},
              {5,5, 0,0},
          // other zero sized tests
        }{
           // zero test code
        }
      // The other range loop that is here now
````
The "zero test code" would use the functions zeroOf(a Matrix) (which would return &Dense{}, etc.) and makeNewOf (which returns NewDense(0,0, nil))

If the goal is also to make sure the receiver (not an argument) is okay having been allocated both ways, we could check that using a test similar to how zero and nonZero are done now.

kortschak:
Abandoned. This was secondary to the other PR I sent today. Good to get your explanation.

PR #175, opened 2015-08-07
mat64: fix matrix formatting support

kortschak:
After discussion at https://groups.google.com/d/topic/gonum-dev/0J9-itOjz08/discussion

This is still fiddling around the edges, but I think the loss of overloading the '#' modifier is a positive step.

### DIFF
--- a/mat64/format.go
+++ b/mat64/format.go
@@ -9,15 +9,66 @@ import (
 	"strconv"
 )
 
-// Format prints a pretty representation of m to the fs io.Writer. The format character c
+// Formatted returns a fmt.Formatter for the matrix m using the given options.
+func Formatted(m Matrix, options ...FormatOption) fmt.Formatter {
+	f := formatter{
+		matrix: m,
+		dot:    '.',
+	}
+	for _, o := range options {
+		o(&f)
+	}
+	return f
+}
+
+type formatter struct {
+	matrix Matrix
+	prefix string
+	margin int
+	dot    byte
+}
+
+// FormatOption is a functional option for matrix formatting.
+type FormatOption func(*formatter)
+
+// Prefix sets the formatted prefix to the string p. Prefix is a string that is prepended to
+// each line of output.
+func Prefix(p string) FormatOption {
+	return func(f *formatter) { f.prefix = p }
+}
+
+// Excerpt sets the maximum number of rows and columns to print at the margins of the matrix
+// to m. If m is zero or less all elements are printed.
+func Excerpt(m int) FormatOption {
+	return func(f *formatter) { f.margin = m }
+}
+
+// DotByte sets the dot character to b. The dot character is used to replace zero elements
+// if the result is printed with the fmt ' ' verb flag. Without a DotByte option, the default
+// dot character is '.'.
+func DotByte(b byte) FormatOption {
+	return func(f *formatter) { f.dot = b }
+}
+
+// Format satisfies the fmt.Formatter interface.
+func (f formatter) Format(fs fmt.State, c rune) {
+	if c == 'v' && fs.Flag('#') {
+		fmt.Fprintf(fs, "%#v", f.matrix)
+		return
+	}
+	format(f.matrix, f.prefix, f.margin, f.dot, fs, c)
+}
+
+// format prints a pretty representation of m to the fs io.Writer. The format character c
 // specifies the numerical representation of of elements; valid values are those for float64
-// specified in the fmt package, with their associated flags. In addition to this, a '#' for
-// all valid verbs except 'v' indicates that zero values be represented by the dot character.
-// The '#' associated with the 'v' verb formats the matrix with Go syntax representation.
+// specified in the fmt package, with their associated flags. In addition to this, a space
+// preceding a verb indicates that zero values should be represented by the dot character.
 // The printed range of the matrix can be limited by specifying a positive value for margin;
 // If margin is greater than zero, only the first and last margin rows/columns of the matrix
 // are output.
-func Format(m Matrix, margin int, dot byte, fs fmt.State, c rune) {
+//
+// format will not provide Go syntax output.
+func format(m Matrix, prefix string, margin int, dot byte, fs fmt.State, c rune) {
 	rows, cols := m.Dims()
 
 	var printed int
@@ -41,8 +92,6 @@ func Format(m Matrix, margin int, dot byte, fs fmt.State, c rune) {
 	)
 	switch c {
 	case 'v', 'e', 'E', 'f', 'F', 'g', 'G':
-		// Note that the '#' flag should have been dealt with by the type.
-		// So %v is treated exactly as %g here.
 		if c == 'v' {
 			buf, maxWidth = maxCellWidth(m, 'g', printed, prec)
 		} else {
@@ -59,12 +108,18 @@ func Format(m Matrix, margin int, dot byte, fs fmt.State, c rune) {
 		pad[i] = ' '
 	}
 
+	first := true
 	if rows > 2*printed || cols > 2*printed {
+		first = false
 		fmt.Fprintf(fs, "Dims(%d, %d)\n", rows, cols)
 	}
 
-	skipZero := fs.Flag('#')
+	skipZero := fs.Flag(' ')
 	for i := 0; i < rows; i++ {
+		if !first {
+			fmt.Fprint(fs, prefix)
+		}
+		first = false
 		var el string
 		switch {
 		case rows == 1:
--- a/mat64/format_example_test.go
+++ b/mat64/format_example_test.go
@@ -1,0 +1,99 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package mat64_test
+
+import (
+	"fmt"
+
+	"github.com/gonum/matrix/mat64"
+)
+
+func Example_Formatted() {
+	a := mat64.NewDense(3, 3, []float64{1, 2, 3, 0, 4, 5, 0, 0, 6})
+
+	// Create a matrix formatting value with a prefix ...
+	fa := mat64.Formatted(a, mat64.Prefix("    "))
+
+	// and then print with and without zero value elements.
+	fmt.Printf("with all values:\na = %v\n\n", fa)
+	fmt.Printf("with only non-zero values:\na = % v\n\n", fa)
+
+	// Modify the matrix...
+	a.Set(0, 2, 0)
+
+	// and print it without zero value elements.
+	fmt.Printf("after modification with only non-zero values:\na = % v\n\n", fa)
+
+	// Output:
+	// with all values:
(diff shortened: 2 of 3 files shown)


kortschak:
@btracey @soniakeys Please take a look.

mat64/format.go
```diff
@@ -11,13 +11,14 @@ import (
```
btracey: I haven't interacted with this function before. At first glance, the signature is pretty daunting. My comments are made with that in mind.

mat64/format.go
```diff
@@ -11,13 +11,14 @@ import (
 
 // Format prints a pretty representation of m to the fs io.Writer. The format character c
 // specifies the numerical representation of of elements; valid values are those for float64
-// specified in the fmt package, with their associated flags. In addition to this, a '#' for
-// all valid verbs except 'v' indicates that zero values be represented by the dot character.
-// The '#' associated with the 'v' verb formats the matrix with Go syntax representation.
+// specified in the fmt package, with their associated flags. In addition to this, a space
```
btracey: I would say "specified in the fmt package (g, g etc.)"

mat64/format.go
```diff
@@ -11,13 +11,14 @@ import (
...
+// preceding a verb indicates that zero values should be represented by the dot character.
```
btracey: I don't understand "space preceeding a verb". It seems from the signature dot is its own thing and the verb is its own thing.

mat64/format.go
```diff
@@ -11,13 +11,14 @@ import (
...
 // The printed range of the matrix can be limited by specifying a positive value for margin;
```
btracey: The margin comment is confusing. Is it trying to say that it just prints the boarder of the matrix?

mat64/format.go
```diff
@@ -11,13 +11,14 @@ import (
...
 // If margin is greater than zero, only the first and last margin rows/columns of the matrix
 // are output.
-func Format(m Matrix, margin int, dot byte, fs fmt.State, c rune) {
+//
+// Format will not provide Go syntax output.
```
btracey: After this comment, I still don't know what prefix is for, and I don't know how I get a fmt.State. I think the resolution to prefix has  something to do with my "preceeding a verb" question. I don't expect an explanation of fmt.State, but a pointer would be nice.

mat64/format.go
kortschak: dot is the character to use to fill cells with when the space modifier is applied to a verb and the cell is 0.

mat64/format.go
kortschak: Currently, the behaviour is e.g. for margin=3:

```
Dims(1, 10)
[ 1   2   3  ...  ...   8   9  10]
```

mat64/format.go
kortschak: The central issue here is that Format is really a helper for people who want to provide fmt.Formatter support for their matrix type (nominally us when we sort out the wrinkles, but also others). The problem is that fmt.Formatter is very poorly documented (for the full extent of the docs see [here](http://golang.org/pkg/fmt/#Formatter) and [here](http://golang.org/pkg/fmt/#State)). As I wrote in the golang-dev thread that sparked this, Format is really a stub that people can use to do all the work of formatting a matrix with minimal work on their part - and uniformly across all types. In short, people should actually not care at all what fs or c are.

btracey: I see the issue with the documentation. 

My gut reaction is that this function is complicated. When I want ot print a matrix, I'm usually looking for a quick way to see the contents of a matrix. With this signature I'd probably just say "meh", call fmt.Println("mat = ", a), and deal with it.

That doesn't mean this function shouldn't exist. I see the merits of the arguments here. I'm thinking.

kortschak: I have an idea that I should have thought of before - what I use in my ct package. I'll push it in a few more minutes.

kortschak:
PTAL

This could make use of examples to help out.

mat64/format.go
```diff
@@ -9,15 +9,46 @@ import (
 	"strconv"
 )
 
-// Format prints a pretty representation of m to the fs io.Writer. The format character c
+// Formatted returns a fmt.Formatter for the matrix m based on the provided prefix, margin and
```
btracey: This is a lot nicer. Is the prepend for '\t' or something? In the json case often json  has no newlines or tabs. For matrix, I assume we're adding braces/ | around the matrix?

In my thinking, "dot" does not replace zero elements, but unused elements. For example, symmetric may have no elements that are zero but has half of the elements unused.

btracey: Oh, I see, dot was already there. It's unrelated to my suggestion. Nevermind.

kortschak:
>In my thinking, "dot" does not replace zero elements, but unused elements. For example, symmetric may have no elements that are zero but has half of the elements unused.

Originally I added dot purely for zero elements because it makes it easier to see what's going on in a sparse matrix. We can add the unused, but that can be done later. I'm also thinking that perhaps functional options might be the best thing here. The API looks OK to me now, but that's probably the icing.

kortschak:
PTAL

mat64/format.go
btracey: We should see what @soniakeys  thinks.

mat64/format.go
```diff
@@ -9,15 +9,66 @@ import (
...
+		o(&f)
+	}
+	return f
+}
+
+type formatter struct {
+	matrix Matrix
+	prefix string
+	margin int
+	dot    byte
```
soniakeys: Could dot be a rune?

mat64/format.go
```diff
@@ -9,15 +9,66 @@ import (
...
+// FormatOption is a functional option for matrix formatting.
+type FormatOption func(*formatter)
+
+// Prefix sets the formatted prefix to the string p. Prefix is a string that is prepended to
+// each line of output.
+func Prefix(p string) FormatOption {
+	return func(f *formatter) { f.prefix = p }
+}
+
+// Margin sets the format margin to m. If m is greater than zero, the margin option indicates
```
soniakeys: "Margin" doesn't have quite the right meaning for me.  Practically it's more like "corners" than margins.  Conceptually, it's more of an "exerpt."  "Take" would have meaning for programmers.  (TakeCorners?)  But it's a small thing.  Especially if the word margin is used like this in other software, then margin is fine.

soniakeys:
This is getting much more usable.  One possibility would be to export `formatter`.  This would increase the API surface but would have the advantage that the struct could be reused, modified and reused, or modified after construction.  `Formatted` could just return the struct rather than the interface.  It might not be worth the API clutter though.  The typical use case is certainly to use the formatter once and drop it.  Supporting reuse might be supporting optimization that is never needed.

mat64/format.go
kortschak: If we want to do all the things necessary to handle variable width dots then yes. I'd rather not do that.

mat64/format.go
kortschak: Excerpt works for me. No other software I've seen does this.

btracey: Why not just "edge". It's what it does -- print the edge of the matrix.

kortschak:
Although the examples don't show it, the formatter can be used as an output handle because the matrix being held is a reference. This is a mechanism for reuse (though not with a different matrix or a different set of options. I don't really like the idea of exporting formatter or making Formatted return the struct because it weakens the documentation effect that we have now; it is very clear what is being returned, a value that formats the given matrix via the fmt.Formatter mechanism. Returning a mat64.Formatter mean that we would now need to add documentation explaining the formatting mechanism.

If it becomes clear that more flexibility is required, we can export or otherwise alter this.

Last night I played around with making the brackets configurable (so we can output text that is directly comparable with matlab or numpy etc, or can be used as input to those. It sort of worked, but the corner cases multiplied and so I've dropped it until it's actually asked for.

mat64/format.go
kortschak: Edge is nice, but unfortunately overloaded and doesn't cover all the behaviours (damn English); it may print the ends rather than the edges, not all the edge is printed, and if the matrix is small enough all is printed.

soniakeys:
LGTM then, that's all I had.  I tried some simple cases with it and it was pretty easy to use.

kortschak:
I'll change the Margin FormatOption to Excerpt.

btracey:
Why are format options better than just having a settings struct? Options make it much harder to see the range of things one can do, they partially obscure the behavior (passing through stateful functions), and increase the API surface.

soniakeys:
> Why are format options better than just having a settings struct?

There are some blog posts out there that give the pitch, but I'm not completely comfortable with them myself.  One advantage of a options functions is that they are--optional.  With a settings struct you still have to pass something, nil typically, even  when you are happy taking all the defaults.

    fmt.Println(mat64.Formatted(m))

becomes

    fmt.Println(mat64.Formatted(m, nil))

kortschak:
The API doc surface get arguably bigger (they really are the same size just more diffuse, but godoc does a good job at keeping the option functions together so the effect is pretty close to zero). However, at the call site, as @soniakeys mentions, the impact is lower with the functional option approach. You either need to pass a nil or zero options struct, or expose the type and then call with `mat64.Formatter{Matrix: m}` or near equivalent.

I don't think it's reasonable to say that it obscures the behaviour. The options in a struct are just as opaque with respect to the behaviour which is entirely defined by the code in format.

mat64/format.go
kortschak: >In particular, some people might like to substitute ASCII alternatives.

Yes, as I mentioned I had a play with this. I'm happy to work on that some more, but we can do that later. I'll file an issue for that if this PR is merged.

kortschak:
PTAL for Excerpt docs.

soniakeys:
Doc is okay.  I like it.

kortschak:
@btracey are you OK with this?

mat64/format.go
```diff
@@ -9,15 +9,66 @@ import (
...
+// FormatOption is a functional option for matrix formatting.
+type FormatOption func(*formatter)
+
+// Prefix sets the formatted prefix to the string p. Prefix is a string that is prepended to
+// each line of output.
+func Prefix(p string) FormatOption {
+	return func(f *formatter) { f.prefix = p }
+}
+
+// Excerpt sets the maximum number of rows and columns to print at the margins of the matrix
```
btracey: How about something like "Excerpt prints a subset of the matrix elements. It prints the mxm block at each corner of the matrix, which can be useful for examining large matrices. If m is zero or less all elements are printed."

btracey:
Yes. Thanks for the example. LGTM.

mat64/format.go
kortschak: The use of the word corners is incorrect, but I will put some literature to in the example - I'll split the big example out and put it as an Excerpt example with the motivation text.

kortschak:
I've split the example into two and shown the behaviour for vector as well.

PTAL

kortschak:
@btracey *ping*

btracey:
LGTM

PR #176, opened 2015-08-08
mat64: reduce Maybe API surface area

kortschak:
I have removed the named type since it adds nothing, but have kept the maybe functions themselves since they provide the functionality that would be written and (this is the important part) do checks on the validity of the mat64.Error that would likely not be done by user code - it can be done, but probably wouldn't be.

Also fix Condition Error's fmt call.

@btracey Please take a look.


btracey:
LGTM.

PR #179, opened 2015-08-09
mat64: note that the view on the row is transposed

kortschak:
The alternative is to return a Transpose{v}, which hides the *Vector.

@btracey Please take a look.

### DIFF
--- a/mat64/dense.go
+++ b/mat64/dense.go
@@ -257,7 +257,8 @@ func (m *Dense) SetRow(i int, src []float64) int {
 	return min(len(src), m.mat.Cols)
 }
 
-// RowView returns a Vector reflecting row i, backed by the matrix data.
+// RowView returns row i of the matrix data represented as a column vector,
+// backed by the matrix data.
 //
 // See RowViewer for more information.
 func (m *Dense) RowView(i int) *Vector {


mat64/dense.go
```diff
@@ -257,7 +257,7 @@ func (m *Dense) SetRow(i int, src []float64) int {
 	return min(len(src), m.mat.Cols)
 }
 
-// RowView returns a Vector reflecting row i, backed by the matrix data.
+// RowView returns a Vector reflecting row i transposed, backed by the matrix data.
```
btracey: How about "RowView returns a (column) Vector reflecting row i, backed by the matrix data"

kortschak: While I'm a great fan of parenthetic writing, they don't read well in documentation.

btracey: What I don't like about your current wording is that it implies something special is happening, which it isn't. The method is doing the logical thing. The documentation just needs a nudge to remind the user that the Vector=column invariant isn't being broken.

Alternate idea:
"RowView returns row i of the matrix data represented as a column vector"

kortschak: That looks fine. I'll take it.

btracey: LGTM

PR #187, opened 2015-08-13
mat64: don't allow invalid uplo values

kortschak:
@btracey Please take a look.

### DIFF
--- a/mat64/triangular.go
+++ b/mat64/triangular.go
@@ -114,8 +114,21 @@ func (t *TriDense) Dims() (r, c int) {
 	return t.mat.N, t.mat.N
 }
 
+// Triangle returns the dimension of t and whether t is an upper triangular
+// matrix. The returned boolean upper is only valid when n is not zero.
 func (t *TriDense) Triangle() (n int, upper bool) {
-	return t.mat.N, t.mat.Uplo == blas.Upper
+	return t.mat.N, !t.isZero() && t.isUpper()
+}
+
+func (t *TriDense) isUpper() bool {
+	switch t.mat.Uplo {
+	case blas.Upper:
+		return true
+	case blas.Lower:
+		return false
+	default:
+		panic(badTriangle)
+	}
 }
 
 // T performs an implicit transpose by returning the receiver inside a Transpose.
--- a/mat64/index_bound_checks.go
+++ b/mat64/index_bound_checks.go
@@ -8,8 +8,6 @@
 
 package mat64
 
-import "github.com/gonum/blas"
-
 // At returns the element at row r, column c.
 func (m *Dense) At(r, c int) float64 {
 	return m.at(r, c)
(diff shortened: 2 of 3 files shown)


mat64/triangular.go
```diff
@@ -115,7 +115,18 @@ func (t *TriDense) Dims() (r, c int) {
 }
 
 func (t *TriDense) Triangle() (n int, upper bool) {
-	return t.mat.N, t.mat.Uplo == blas.Upper
+	return t.mat.N, t.isUpper()
```
btracey: I assume we want the following to continue working:
````
var t TriDense
t.Cholesky(a, true)
````
In this case, there will be cases where t has an invalid uplo. I think we should say something like "For a zero-sized matrix Triangle returns false by convention, though the orientation is flexible"

File: mat64/list_test.go [hunk 2]
```diff
@@ -383,7 +384,8 @@ func testTwoInput(c *check.C,
 		&basicMatrix{},
 		&basicVectorer{},
 		&basicSymmetric{},
-		&basicTriangular{},
```
btracey: For the reason I say below, it can't be that we always have a valid uplo. I think we should leave in the explicit zero value, and then add to the "make rand of" triangular block. Here, if the requested size == 0, the returned triangular has the uplo of the input to makeRandOf. This way we can ensure code works when the receiver has an invalid uplo.

mat64/triangular.go
kortschak: In Cholesky the TriDense's uplo is not queried (I should check for other cases wher the value uplo == Upper is used). The only thing that doesn't work after this is calling Triangle on the zero value AFAICS.

mat64/list_test.go
kortschak: I don't understand the rationale for this; at 380 and 381 we are already creating both triangles with valid uplo states.

btracey: Because we can accidentally write functions that assume Uplo is set properly, even though it is not always. See for example https://github.com/gonum/matrix/issues/171 .

mat64/triangular.go
btracey: Agreed. I just don't think it should panic in that case.

btracey: "That case" meaning calling Triangle on a zero-sized matrix. I'd rather it return false.

mat64/list_test.go
kortschak: This is the stronest argument that this should be a panic.

mat64/triangular.go
kortschak: Still thinking, but I could add ��!t.isZero() &&��. It feels unsatisfying, and I'd prefer to just document that Triangle may not be called on a zero value, but I'll bend.

kortschak:
PTAL

I've weakened the assertion and documented that upper is only valid when n != 0.

Also, the `panic("mat64: triangular set out of bounds")` is SetTri is user-facing, so it should be a mat64.Error, but that ca be fixed later.

btracey:
LGTM.

Issue #138, opened 2015-05-27
mat64: consider retiring less used interface definitions


kortschak:
The interface definitions in matrix.go served a useful purpose in defining the direction of development for mat64. We now have a fairly well defined set of interfaces based on the concrete types, so with the exception of interfaces that are used within the package or are likely to be useful to clients we should start considering retirement for these.

Please add you thoughts on which should be retired and which should stay.

kortschak:
My proposed list of interfaces to delete based on a quick glance:

 - [ ] Matrix
 - [ ] Mutable
 - [ ] Vectorer
 - [ ] VectorSetter
 - [ ] RowViewer
 - [ ] RawRowViewer
 - [ ] ColViewer
 - [ ] RawColViewer
 - [ ] Cloner
 - [ ] Reseter
 - [ ] Copier
 - [ ] Viewer
 - [ ] Grower
 - [x] Normer
 - [x] TransposeCopier
 - [x] Transposer
 - [x] Deter
 - [x] Inver
 - [ ] Adder
 - [ ] Suber
 - [ ] ElemMuler
 - [ ] ElemDiver
 - [ ] Equaler
 - [ ] ApproxEqualer
 - [ ] Scaler
 - [ ] Sumer
 - [ ] Muler
 - [ ] MulTranser
 - [x] Exper
 - [x] Power
 - [ ] Dotter
 - [x] Stacker
 - [x] Augmenter
 - [x] Applyer
 - [x] Tracer
 - [x] Uer
 - [x] Ler
 - [ ] BandWidther
 - [ ] RawMatrixSetter
 - [ ] RawMatrixer
 - [ ] RawVectorer
 - [ ] Symmetric
 - [ ] RawSymmetricer
 - [ ] Triangular
 - [ ] RawTriangular

Collaborators feel free to edit this post to add or delete check marks (please note if you make a change what the change was).

btracey:
Equaler and Deter are gone.

PR #188, opened 2015-08-13
mat64: add Copy to TriDense and drop Ler/Uer interfaces and methods

kortschak:
Depends #187.
Updates #138.

@btracey Please take a look.

### DIFF
--- a/mat64/triangular.go
+++ b/mat64/triangular.go
@@ -184,6 +184,63 @@ func (t *TriDense) Reset() {
 	t.mat.Data = t.mat.Data[:0]
 }
 
+// Copy makes a copy of elements of a into the receiver. It is similar to the
+// built-in copy; it copies as much as the overlap between the two matrices and
+// returns the number of rows and columns it copied. Only elements within the
+// receiver's non-zero triangle are set.
+//
+// See the Copier interface for more information.
+func (t *TriDense) Copy(a Matrix) (r, c int) {
+	r, c = a.Dims()
+	r = min(r, t.mat.N)
+	c = min(c, t.mat.N)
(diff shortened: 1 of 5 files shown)


mat64/triangular.go
```diff
@@ -171,6 +171,63 @@ func (t *TriDense) Reset() {
...
+		row := make([]float64, c)
+		if t.isUpper() {
+			for i := 0; i < r; i++ {
+				a.Row(row, i)
+				copy(t.mat.Data[i*t.mat.Stride+i:i*t.mat.Stride+c], row[i:])
+			}
+		} else {
+			for i := 0; i < r; i++ {
+				a.Row(row, i)
+				copy(t.mat.Data[i*t.mat.Stride:i*t.mat.Stride+i+1], row[:c-i])
```
btracey: Shouldn't this be row[:i+1] ?

kortschak: Yes, it should.

btracey:
LGTM

We should definitely add a fast past for TriDense - TriDense copying.

kortschak:
What is the expected behavior u.Copy(l) in that case? I would expect just the diagonal.

btracey:
Agreed.

kortschak:
Fixed slow path - upper/lower logic was inverted. Nice to see our testing is tight.

btracey:
Shoot! I thought I checked that, sorry. Yes, it is good.

btracey:
It caught a bunch of bugs for me in the case of Dense multiplication where the underlying inputs are both *Dense. It's more complicated than one thinks at first.

Issue #169, opened 2015-08-03
Redesign Solve


btracey:
(Adapted from an off-line discussion)

Lapack is almost at a state where it can implement the full functionality of the current Solve.

I don’t think the current Solve function is idiomatic within mat64. The result of the operation is a matrix, so it should be stored in-place into a type like any other operation. I have found that linear solving for a vector is very common, so I would suggest
````
func (m *Dense) Solve(a, b Matrix) error // possibly bool
func (v *Vector) SolveVec(a Matrix, b *Vector) error // possibly bool
````
I have learned in dealing with Lapack that the factorizations have special forms for storing data. The PLU decomposition stores L and U in a single matrix, as does the QR factorization. I believe this merits the continued existence of the LUFactors and QRFactor types. I would guess we can find better names (at the very least we should pick one of Factor and Factors), but the current ones aren’t terrible. Possibly we could have the name just be “LU”, and then we could have 

````
// Factorize finds the LU factorization of the matrix
func (lu *LU) Factorize(a Matrix) 
````

As now, the user can extract the types. It possibly should be something like
````
// GetL extracts the L of an LU decomposition
func (t *Triangular) GetL(lu *LU) 
````

It is common to want to perform Solve many times using the same matrix. It is much more efficient to perform the factorization once and then use the factorization many times to perform different solves. Additionally, the various decompositions have differing tradeoffs when it comes to linear solving. For example, LU has a lower constant, but is less robust. It seems nice to enable solving using specific methods for users with special needs (or, if Solve fails, they can error check and try QR or whatever). Both of these imply having a special solve like we now have with Cholesky
````
// SolveLU computes the solution to a set of equations using a pre-computed LU factorization. 
func (m *Dense) SolveLU(lu *LU, trans bool, b Matrix)
````
The trans field is necessary because LU doesn’t actually implement Matrix. We would similarly provide a method for QR and the corresponding *Vector.

[...]

PR #195, opened 2015-08-14
Fix LQ implementation to use Lapack, add Solve routines and tests

btracey:


btracey:
Alright! After this one more PR and #169  is done!

btracey:
PTAL @kortschak

mat64/lq.go
```diff
@@ -9,185 +9,204 @@ import (
...
+// TODO(btracey): Add in the "Reduced" forms for extracting the m×m orthogonal
+// and upper triangular matrices.
+
+// LFromLQ extracts the m×n lower trapezoidal matrix from a LQ decomposition.
+func (m *Dense) LFromLQ(lq *LQ) {
+	r, c := lq.lq.Dims()
+	m.reuseAs(r, c)
+
+	// Disguise the LQ as a lower triangular
+	t := &TriDense{
```
kortschak: Can we have this using a helper `asTriDense(*Dense) *TriDense`. (Did that end up getting done in the previous PR?)

mat64/lq.go
```diff
@@ -9,185 +9,204 @@ import (
...
-				l.set(j, i, lq.at(j, i))
-			} else if i == j {
-				l.set(j, i, v)
-			}
-		}
+	if r == c {
+		return
+	}
+	// Zero right of the triangular.
+	for i := 0; i < r; i++ {
```
kortschak: I just deleted a zeroLower and zeroUpper. Shoudl we bring those back and use them here?

mat64/lq.go
```diff
@@ -9,185 +9,204 @@ import (
...
+//  1. If m >= n and trans == false, find X such that || A*X - B||_2 is minimized.
+//  2. If m < n and trans == false, find the minimum norm solution of A * X = B.
+//  3. If m >= n and trans == true, find the minimum norm solution of A^T * X = B.
+//  4. If m < n and trans == true, find X such that || A*X - B||_2 is minimized.
+// The solution matrix, X, is stored in place into the receiver.
+func (m *Dense) SolveLQ(lq *LQ, trans bool, b Matrix) error {
+	r, c := lq.lq.Dims()
+	br, bc := b.Dims()
+
+	// The LQ solve algorithm stores the result in-place into B. The storage
```
kortschak: This comment is confusing because it is overloading the LAPACK b and B and the parameter b.

mat64/lq.go
btracey: These are more specialized than those functions. They zero outside the leading triangle.

mat64/lq.go
btracey: Sure.

mat64/lq.go
btracey: Done.

btracey:
Updated the comments to remove the impossible cases (for lq, m <= n, so no point in talking about the other cases).

PTAL

mat64/lq.go
kortschak: I meant the command at line 143. The comment is confusing because there is no such thing for in place into B when B is a matrix (lower-case 'm'). If the capital Bs are lower-cased and the capital X is lower-cased I think that fixes it.

btracey:
Okay, tried another stab at the internal comment.

kortschak:
Thank you.

LGTM

PR #196, opened 2015-08-15
mat64: drop TCopy method and associated interface

kortschak:
Depending on the desired behaviour, m.TCopy(a) can be replaced by m.Copy(a.T()) or m.Clone(a.T()). This significantly clarifies the intention over the previous situation; TCopy's semantics were weird - it did not obey the normal copy semantics (it allocated when the receiver was zero) and did not obey normal clone semantics (it did not destroy the surrounding matrix).

Updates #138.

@btracey  Please take a look.

TODO(kortschak) add "Fixes # 194." before merge.

### DIFF
--- a/mat64/dense.go
+++ b/mat64/dense.go
@@ -37,9 +37,6 @@ var (
 	_ Scaler  = matrix
 	_ Applyer = matrix
 
-	_ TransposeCopier = matrix
-	// _ TransposeViewer = matrix
-
 	_ Tracer = matrix
 	_ Normer = matrix
 	_ Sumer  = matrix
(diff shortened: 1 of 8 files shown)


kortschak:
The tests are not good enough - Vectorer and basic Matrix are not tested, but they pass with test alteration. When testOneInput is done things will be OK.

mat64/dense.go
```diff
@@ -413,21 +431,39 @@ func (m *Dense) Clone(a Matrix) {
...
+		return r, c
+	}
 	r = min(r, m.mat.Rows)
 	c = min(c, m.mat.Cols)
 	if r == 0 || c == 0 {
 		return 0, 0
 	}
 
-	switch a := a.(type) {
+	aU, trans := untranspose(a)
```
btracey: This code looks wrong if m == aU, as in a.Copy(a.T())?

btracey: Oh, I see your note.

kortschak: Note that I've changed docs on Copier to say that aliased data means the behaviour from a call to Copy is undefined.

btracey:
LGTM, though when we add testOneInput we'll need to special case the a.Copy(a.T())

kortschak:
Not just that. I just made a testOneInput derived from your testTwoInput and it doesn't handle some of the cases for Clone either - there needs to be a isShapeRestricted option. There are only a small number of single matrix parameter methods, more have a single matrix and other types.

btracey:
I wonder if it's really worth it. Clone and Copy behave differently than all of the other functions. Maybe better to just have "testClone" that works on Clone/Copy for the matrix types.

kortschak:
Even those are different - have a look at the test-one branch.

btracey:
Yea. The copy functions may just need to be hand tested.

PR #199, opened 2015-08-15
Fixsolve

btracey:


btracey:
PTAL @kortschak 

I removed the test when a is all zeros for skinny and tall. It turns out Lapack doesn't think this condition is an error. I'm not sure why since A is singular. Technically Lapack does return a true solution (any x is a minimizer of the equation), but still. It seems better to me to leave the lapack behavior as is (for consistency with other implementations) and fix the issue in the future with a condition number check.

mat64/list_test.go
```diff
@@ -39,6 +39,14 @@ func legalSizeSameSquare(ar, ac, br, bc int) bool {
 	return true
 }
 
+// legalSizeSolve returns whether the two matrices can be used in a linear solve.
+func legalSizeSolve(ar, ac, br, bc int) bool {
+	if ar != br {
```
kortschak: `return ar == br`

mat64/matrix_test.go
```diff
@@ -90,209 +90,3 @@ func (s *S) TestMaybe(c *check.C) {
...
-				{0, 0},
-				{0, 0},
-			},
-			b: [][]float64{
-				{3},
-				{2},
-			},
-			x: nil,
-		},
-		{
```
kortschak: Please put these back, commented out with a note.

mat64/solve.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/gonum/blas"
+	"github.com/gonum/lapack/lapack64"
+)
+
+// Solve solves a minimum-norm solution to a system of linear equations defined
+// by the matrices a and b. If a is singular or near-singular a Condition error
+// is returned. Please see the documentation for Condition for more information.
+//
+// The minimization problem solved depends on the input parameters.
+//  1. If m >= n and trans == false, find x such that || a*x - b||_2 is minimized.
```
kortschak: s/|| a/||a/g. I think I saw this elsewhere too, but forgot - those fixed in another PR.

mat64/solve.go
```diff
@@ -0,0 +1,87 @@
...
+	m.reuseAs(ac, bc)
+	var restore func()
+	if m == aMat {
+		m, restore = m.isolatedWorkspace(aMat)
+		defer restore()
+	} else if m == bMat {
+		m, restore = m.isolatedWorkspace(bMat)
+		defer restore()
+	}
+	// TODO(btracey): Add a test for the condition number of A.
```
kortschak: Could this go into this PR?

mat64/solve.go
```diff
@@ -0,0 +1,87 @@
...
+		return nil
+	default:
+		// Solve using QR/LQ.
+
+		// Copy a since the corresponding LAPACK argument is modified during
+		// the call.
+		var aCopy Dense
+		aCopy.Clone(a)
+
+		x := getWorkspace(max(ar, ac), bc, false)
```
kortschak: You leak a workspace, `defer putWorkspace(x)`.

mat64/solve_test.go
```diff
@@ -0,0 +1,289 @@
...
+			},
+			b: [][]float64{
+				{3},
+				{2},
+			},
+			ans:       nil,
+			shouldErr: true,
+		},
+	} {
+		a := NewDense(len(test.a), len(test.a[0]), nil)
```
kortschak: Use `flatten` here and below.

mat64/solve_test.go
```diff
@@ -0,0 +1,289 @@
...
+			for i := 0; i < len(test.ans); i++ {
+				for j := 0; j < len(test.ans[0]); j++ {
+					ans.Set(i, j, test.ans[i][j])
+				}
+			}
+		}
+
+		var x Dense
+		err := x.Solve(a, b)
+		if err != nil {
```
kortschak: `if err != nil && !test.shouldErr {` ...

mat64/solve.go
```diff
@@ -0,0 +1,87 @@
...
+	br, bc := b.Dims()
+	if ar != br {
+		panic(ErrShape)
+	}
+
+	aMat, _ := untranspose(a)
+	bMat, _ := untranspose(b)
+	m.reuseAs(ac, bc)
+	var restore func()
+	if m == aMat {
```
kortschak: What happens when m == aMat == bMat?

mat64/solve_test.go
```diff
@@ -0,0 +1,289 @@
...
+			if !test.shouldErr {
+				c.Errorf("Unexpected solve error: %s", err)
+			}
+			continue
+		}
+		if err == nil && test.shouldErr {
+			fmt.Println("a =", a)
```
kortschak: These probably don't want to be here.

mat64/solve.go
```diff
@@ -0,0 +1,87 @@
...
+
+package mat64
+
+import (
+	"github.com/gonum/blas"
+	"github.com/gonum/lapack/lapack64"
+)
+
+// Solve solves a minimum-norm solution to a system of linear equations defined
+// by the matrices a and b. If a is singular or near-singular a Condition error
```
kortschak: This is not true, only ErrSingular is returned.

kortschak:
>It seems better to me to leave the lapack behavior as is (for consistency with other implementations) and fix the issue in the future with a condition number check.

At the moment the behaviour does not match the docs - it says a Condition error is returned. Either change the docs to say that singular matrices result in ErrSingular and add a BUG comment saying that skinny and wide zero matrices are currently not detected due to LAPACK behaviour or add the check to this PR.

mat64/list_test.go
btracey: Done.

mat64/matrix_test.go
btracey: Done.

mat64/solve.go
btracey: Done.

mat64/solve.go
btracey: Oops. Fixed to return Condition. I'll remove ErrSingular in the next PR, as it's still referenced in documentation.

mat64/solve.go
btracey: Thanks.

mat64/solve_test.go
btracey: Done.

mat64/solve_test.go
btracey: No, it's not the same. If err != nil, then we don't want to compare x with the answer. We want to continue if err != nil regardless of test.shouldErr

mat64/solve_test.go
btracey: Shoot. Sorry. Missed that in my check.

mat64/solve.go
btracey: I think it's too much work for this PR. This is done in lapack with Dgecon. I believe Dgecon uses the LU decomposition, which is useful for m == n. I believe it is also possible to compute with the QR/LQ decomposition. If that is true, then I'd rather use that algorithm and then use qr.Solve, that way we don't have to compute both the LU and the QR for solve.

mat64/solve.go
btracey: What do you think would happen? Because of the elseif the temporary memory is only allocated once. This case is tested in testTwoInput  (aMaybeSame and bMaybeSame)

btracey:
PTAL

mat64/solve_test.go
kortschak: Yeah, OK.

mat64/solve.go
kortschak: I think this is case-by-case. Here, we can get a fast path and return I.

btracey: Actually, we have to copy anyway. We shouldn't need to check this, right?

kortschak: That's true. And even without that, it's impossible for a or b to be m unless they are all square, so the check would only be in the LU case. We could add a fast path for a == b though.

btracey: What would be fast pathed? In LU b still has to be copied into m. In QR it still has to be copied twice because A and B are modified independently in Gels.

kortschak: In the LU case if a == b then we don't need to factorise and can just fill m with I. No?

btracey: Of course. I don't know when that would happen, but there's not a lot of cost either.

mat64/solve.go
kortschak: I'm confused about the scheme that's being used to decide whether an entity is labelled with a capital or not. It seems to me that we should be using maths notation if we are describing the mathematical entity and the parameter literally if we are describing a paramter."x" is not a parameter, so it must be a mathematical matrix. At the very least the capitalisation should be consistently applied to each individual entity.

mat64/solve.go
btracey: Oh, you know what, we made it that copy doesn't work if m and a are the same. So, m.Copy(b) is wrong if b == m, thus we need that test for LU.

kortschak: The Copy behaviour for m == a is an unspecified no-op (probably we could specify it). If it's a guaranteed no-op, the current behaviour in the LU case is correct, otherwise we should put in a defensive check and not copy when m == a.

mat64/solve.go
btracey: I agree I am also confused.

In solve, it should be a and b because they are actual inputs. In LU and QR it should be “A” since the matrix is implicitly represented. In all cases X should be capitalized since it is always implicit?

mat64/solve.go
btracey: Not if  a = Transpose{m}

btracey: In particular, a.Copy(a.T()), does not work at the moment. Maybe it should.

mat64/solve.go
kortschak: Yes, that seems the way it should work.

btracey:
Updated the Solve case to only worry about shadowing when m == b and btranspose.

mat64/solve.go
kortschak: a.Copy(a.T()) was why I made it undefined. We have to allocate if we cover that. I comfortable with that decision.

There needs to be a broader issue of shadowing. We can't detect all cases, but we can go a long way towards covering types that have a Raw* method.

kortschak:
Not worth changing (maybe later), but splitting the range for setting the eye doesn't seem worthwhile - the cost of the aditional bounds checks and the change in branch behaviour probably offsets the benefit of only writing to a location once. I'd just zero the row and the set the diagonal (I mention this because I saw you doing this elsewhere before).

LGTM

mat64/solve.go
btracey: > On Aug 15, 2015, at 10:58 PM, Dan Kortschak <notifications@github.com> wrote:
> 
> In mat64/solve.go <https://github.com/gonum/matrix/pull/199#discussion_r37142021>:
> 
> > +//  2. If m < n and trans == false, find the minimum norm solution of a * x = b.
> > +//  3. If m >= n and trans == true, find the minimum norm solution of a^T * x = b.
> > +//  4. If m < n and trans == true, find X such that || a*x - b||_2 is minimized.
> > +// The solution matrix, x, is stored in place into the receiver.
> > +func (m *Dense) Solve(a, b Matrix) error {
> > +	ar, ac := a.Dims()
> > +	br, bc := b.Dims()
> > +	if ar != br {
> > +		panic(ErrShape)
> > +	}
> > +
> > +	aMat, _ := untranspose(a)
> > +	bMat, _ := untranspose(b)
> > +	m.reuseAs(ac, bc)
> > +	var restore func()
> > +	if m == aMat {
> a.Copy(a.T()) was why I made it undefined. We have to allocate if we cover that. I comfortable with that decision. There needs to be a broader issue of shadowing. We can't detect all cases, but we can go a long way towards covering types that have a Raw* method.

It’s a reasonable decision for copy. I don’t know about the normal case. I’m not sure I want to have to reason about code like
c := a.View(1,2, 10, 9)
c.Solve(a,b)
d.Mul(a,b)

btracey:
Yea, you're probably right. The cost is at worst a cache miss. In addition to what you say, the vectorization is harder/less effective.

PR #201, opened 2015-08-16
mat64: simplify and clarify I loops

kortschak:
Following from previous discussion. Also changed `h := I` loops because of (legal but surprising) label reuse.

@btracey Please take a look.


btracey:
LGTM

PR #204, opened 2015-09-11
Add condition number check to solving with LQ and QR factorization

btracey:


btracey:
The next PR will be fixing mat64.Solve.

kortschak:
LGTM

PR #206, opened 2015-09-11
Change Cholesky to be a struct like QR, LU, and LQ. Add condition num…

btracey:
…ber check to Solve routines

### DIFF
--- a/mat64/cholesky.go
+++ b/mat64/cholesky.go
@@ -6,6 +6,8 @@
 package mat64
 
 import (
+	"math"
+
 	"github.com/gonum/blas"
 	"github.com/gonum/blas/blas64"
 	"github.com/gonum/lapack/lapack64"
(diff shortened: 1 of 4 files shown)


btracey:
It's so nice to see greens the first time.

mat64/cholesky.go
```diff
@@ -13,50 +15,65 @@ import (
...
+}
+
+// updateCond updates the condition number of the Cholesky decomposition. If
+// norm > 0, then that norm is used as the norm of the original matrix A, otherwise
+// the norm is estimated from the decompositon.
+func (c *Cholesky) updateCond(norm float64) {
+	n := c.chol.mat.N
+	work := make([]float64, 3*n)
+	if norm < 0 {
+		// This is an approximation. By the defintion of a norm, ||AB|| <= ||A|| ||B||.
```
kortschak: definition

mat64/cholesky.go
```diff
@@ -13,50 +15,65 @@ import (
...
-			Stride: n,
-			Diag:   blas.NonUnit,
-			Data:   use(t.mat.Data, n*n),
-		}
-		if upper {
-			t.mat.Uplo = blas.Upper
-		} else {
-			t.mat.Uplo = blas.Lower
-		}
+	if c.chol == nil {
```
kortschak: Can't this just use the else case operation and lose the condition test?

kortschak:
LGTM

PR #207, opened 2015-09-12
Add determinant routine to Cholesky

btracey:


mat64/cholesky.go
```diff
@@ -68,7 +68,19 @@ func (c *Cholesky) Factorize(a Symmetric) (ok bool) {
 	return ok
 }
 
-// TODO(btracey): Add in UFromChol and LFromChol
+// Det returns the determinant of the matrix that has been factorized.
+func (c *Cholesky) Det() float64 {
+	return math.Exp(c.Det())
```
kortschak: This is not good: math.Exp(c.LogDet()).

kortschak:
LGTM after non-terminating recursion removed.

PR #209, opened 2015-09-13
Increase the stability of the computation of Det, and add LogDet func…

btracey:
…tion to enable enhanced numeric stability

### DIFF
--- a/mat64/lu.go
+++ b/mat64/lu.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/gonum/blas"
 	"github.com/gonum/blas/blas64"
+	"github.com/gonum/floats"
 	"github.com/gonum/lapack/lapack64"
 )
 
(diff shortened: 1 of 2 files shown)


btracey:
Using LogSumExp is much better than multiplying the elements directly for computing the determinant (for example if there is one very large diagonal element and another very small element).

 I could use some help wordsmithing the explanation for why LogDet. Many matrix expressions have something like const / abs(det(A)). It's much better to do exp( log(const) - abs(logdet)) than it is to try and compute the determinant explicitly. This is the sort of case I am trying to describe.

mat64/lu.go
```diff
@@ -68,14 +69,29 @@ func (lu *LU) Factorize(a Matrix) {
...
-// Det returns the determinant of the matrix that has been factorized.
+// Det returns the determinant of the matrix that has been factorized. In many
+// expressions, using LogDet will be more numerically stable.
 func (lu *LU) Det() float64 {
+	det, sign := lu.LogDet()
+	return math.Exp(det) * sign
+}
+
+// LogDet returns the log of the determinant and the sign of the determinant
+// for the matrix that has been factorized. Many expressions multiply or divide
```
kortschak: I'm not sure that this needs to be spelled out, but:

"Numerical stability in product and division expressions is generally improved by working in log space."

kortschak:
LGTM

kortschak:
Updates #138.

Issue #215, opened 2015-09-18
Create example for Cholesky factorization


ericdfournier:
The updated routines for handling Cholesky decomposition -- which are now based upon the passing of a *Cholesky struct -- do not seem to contain a function for initializing those structures.

Perhaps a 'NewCholesky()' function, or something of the like, which takes as input arguments 'DenseTri' inputs could be added to facilitate this process.

Thanks to all of the contributors for your hard work in making this project great. 

- Eric Fournier

btracey:
The short answer us you can say

var chol Cholesky
chol.Factorize(A)

ericdfournier:
Thanks!

btracey:
Reopening this to remind us to create an Example.

kortschak:
Closed by #216.

PR #216, opened 2015-09-19
Add example code for computing and using Choleksy decomposition.

btracey:
Fixes issue #215 .

### DIFF
--- a/mat64/cholesky_example_test.go
+++ b/mat64/cholesky_example_test.go
@@ -1,0 +1,68 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package mat64_test
+
+import (
+	"fmt"
+
+	"github.com/gonum/matrix/mat64"
+)
+
+func ExampleCholesky() {
+	// Construct a new SymDense. Only the upper triangular data elements are used
+	// so the lower triangular elements could alternately be zero.
+	a := mat64.NewSymDense(4, []float64{
+		108, -21, -9, -96,
+		-21, 7, 5, 15,
+		-9, 5, 61, 25,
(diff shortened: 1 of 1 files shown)


mat64/cholesky_example_test.go
```diff
@@ -0,0 +1,58 @@
+package mat64_test
```
kortschak: License header.

mat64/cholesky_example_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"fmt"
+
+	"github.com/gonum/matrix/mat64"
+)
+
+func ExampleCholesky() {
+	a := mat64.NewSymDense(4, []float64{108, -21, -9, -96, -21, 7, 5, 15, -9, 5, 61, 25, -96, 15, 25, 142})
```
kortschak: Format the data over four lines and it would be nice to have a comment explaining that the lower triangle could be zeroed.

mat64/cholesky_example_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+	fmt.Printf("a = %0.4v\n", mat64.Formatted(a, mat64.Prefix("    ")))
+
+	// Compute the cholesky factorization.
+	chol := &mat64.Cholesky{}
```
kortschak: You are using concrete values below (t and test), so do the same here or change those to this form (personally I prefer the concrete form, but in examples I'm not as fussed).

kortschak:

mat64/cholesky_example_test.go
btracey: Done.

mat64/cholesky_example_test.go
btracey: Good idea. Done.

PR #217, opened 2015-09-19
Change Equal to be a function

btracey:


mat64/matrix.go
```diff
@@ -328,6 +322,77 @@ func Det(a Matrix) float64 {
 	return math.Exp(det) * sign
 }
 
+// Equal returns whether element a and b have the same size and contain all equal
+// elements.
+func Equal(a, b Matrix) bool {
+	ar, ac := a.Dims()
```
kortschak: Add `if a == b { return true }` for a faster path.

kortschak:
LGTM

mat64/matrix.go
btracey: It's not true with NaN (stupid NaN).

kortschak: It's actially worse that that. If a user defined Matrix has a concrete type base and the type is of contains a sloce, map, chan or func, the test would panic since that are not comparable. Those are unlikely, but not impossible.

PR #218, opened 2015-09-20
mat64: change wording of comment to avoid overloaded English

kortschak:
The word "alternate" in non-North American English means to change from one to a another (*v*.) successively or the other that is alternated to, i.e. every second one (*v*.). In North American English it is increasingly being used to signify "alternative", however it retains the original meaning. So avoid the term.

In the Cholesky comment it could be interpreted to mean that every second element is zero. Foolishly, true, but let's not allow that.

@btracey Please take a look.

### DIFF
--- a/mat64/cholesky_example_test.go
+++ b/mat64/cholesky_example_test.go
@@ -11,13 +11,15 @@ import (
 )
 
 func ExampleCholesky() {
-	// Construct a new SymDense. Only the upper triangular data elements are used
-	// so the lower triangular elements could alternately be zero.
+	// Construct a new SymDense. Only the upper triangular data
+	// elements are used, so the lower triangular elements may
+	// be set zero without altering the semantics.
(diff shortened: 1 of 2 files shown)


btracey:
Lgtm

To be fair, the elements aren't used as all, so alternatively they could alternately be set to zero as well

kortschak:
Like so?

```
a := mat64.NewSymDense(4, []float64{
	108, 0, -9, 0,
	-21, 0, 5, 0,
	-9, 0, 61, 0,
	-96, 0, 25, 0,
})
```

I think we might be inventing a new optimisation.

btracey:
No, just the elements in the lower triangular portion

kortschak:
Ahh, but they are not alternates unless you have an interesting measure for the unit.

kortschak:
Oh, sorry, yes I see what you mean.

PR #219, opened 2015-09-21
Add Symmetric Rank K function

btracey:


btracey:
I added the scaffolding around reuseAs before realizing it wasn't needed. I can remove from this PR if you'd like, but it seemed silly to delete the effort.

We should have an equivalent to Outer for RankK. Name idea? OuterK?

kortschak:
>I added the scaffolding around reuseAs before realizing it wasn't needed.

Do you mean isolatedWorkSpace and the pools? Where is it likely to be used?

>We should have an equivalent to Outer for RankK. Name idea? OuterK? 

SGTM

btracey:
I assume it would be used if we added more operations to Symmetric (multiply symmetric matrices for example).

btracey:
Would you like me to put OuterK into this PR or the next one?

kortschak:
>I assume it would be used if we added more operations to Symmetric (multiply symmetric matrices for example).

I'm torn. I don't want to make extra work for you, but I don't like leaving dead code lying around. I'll leave it up to you, but in case you decide the remove those lines, the commit holding them will be recoverable here 1e99c958acea6b7de4879a8c6f4650df69ec51b7.

>Would you like me to put OuterK into this PR or the next one?

Add here I reckon.

btracey:
Pool deleted. OuterK added.

kortschak:
LGTM

PR #222, opened 2015-09-21
Update Cholesky example to construct a symmetric positive definite ma…

btracey:
…trix


btracey:
I think this is more informative. Shows an easy way a symmetric positive definite matrix can occur and avoids the SymDense literal issues.

kortschak:
Nice. LGTM

Issue #38, opened 2014-11-18
The number of interfaces cause many edge cases


jonlawlor:
This is following up on discussion in https://github.com/gonum/stat/pull/19

As currently defined, the interfaces in mat64/ make it so that you might get a Matrix which doesn't implement various matrix operations (such as Col, Row, Add, Mul, ...) which makes implementing functions which accept a Matrix full of edge cases.  A caller has to see if the input Matrix implements arithmetic (for example) and then perhaps have a fallback, and another fallback, and so on.  It makes it difficult to implement and test.

In addition, many of the methods panic when there is no blas engine.  We should probably register goblas by default when it is ready, which will reduce the dimensions of the possible interfaces.

One way to do this is to make methods like Row, Col, Add, Mul, etc. have sigs like
```go
Col(x Matrix, i int) z []float64 { ... }
```
Which would then chose from the available interfaces of x and y to produce a result.  For example, if x implements Vectorer, then it could just call x.Col(i), otherwise, if it implements RawMatrixer, then it could just call RawMatrix() and pull the values out at the appropriate indexes.  If that is not available, then it could call At.

Alternatively, all of the interfaces could be folded into Matrix, so that *all* Matrixes would have to implement Col, and Row, and etc.  That would be easier if goblas was always on.

btracey:
I wonder if the package is too object-oriented at the moment, for reasons that @jonlawlor mentions. It can be hard to perform operations on abstract matrices. In some cases, this is a good idea (the signatures of Add and AddSym are different for a reason), but in other cases, like Col, it doesn't seem to be. 

An alternate proposal would be the following: A specific functionality is a method if, and only if, it does one of the following:
1) Implements a package-level interface (i.e. the methods of Matrix)
2) Modifies the values in the receiver (many of the arithmetic operations)
3) Exposes data that could modify the value of the receiver (RawRowView)
4) Involves data specific to the matrix class.

If the functionality does not fit into one of those categories, it is a package-level function, taking in a Matrix (or a specific matrix class). 

Here's what I see based on the current methods of dense. Number is the category under which it falls, F means will become a function. Specific other cases.
```
Add - 2
Apply - 2
At - 1
Augment - 2
Caps - 4
Clone - 2
Col - F
ColView - 3
Copy - 2
Dims - 1
DivElem - 2
Dot - F
Equals - F
EqualsApprox - F
Exp - 2
Grow - 2
L - Should be a method on Triangular
[...]

PR #225, opened 2015-09-22
Make Row and Col functions and add tests

btracey:


btracey:
Should I delete the Vectorer interfaces as well? They specify the previous copy/like behavior which doesn't match the current behavior. It's also not clear to me what their utility is now that rows are easily extractable. 

This just has the single fast path for Dense, which matches the previous fast path. There should be a fast path for row/column viewer, as well as the other basic types. I can either add them to this PR, or I can add a TODO.

mat64/dense.go
```diff
@@ -231,22 +209,7 @@ func (m *Dense) SetCol(j int, src []float64) int {
...
-	}
-
-	if dst == nil {
-		dst = make([]float64, m.mat.Cols)
-	}
-	copy(dst, m.rowView(i))
-
-	return dst
-}
+// TODO(btracey): Change the comments and behavior for SetRow.
```
btracey: Oops. I'll have to deal with this too.

btracey:
Updates #38

kortschak:
>Should I delete the Vectorer interfaces as well? They specify the previous copy/like behavior which doesn't match the current behavior. It's also not clear to me what their utility is now that rows are easily extractable. 

Not sure. Maybe the methods should be renamed to CopyCol, CopyRow - though I don't like these. The issue of their utility is not settled though, a matrix that is not built on a blas64 type gets a slow path, but might have very fast access to rows and columns internally. That was the intention of this type.

kortschak:
What does this do to benchmarks?

btracey:
We can also just change the defined behavior of Vectorer so that it returns the whole row. I agree with the plausible utility of the general type

btracey:
(and then add a Vectorer fast path)

kortschak:
>(and then add a Vectorer fast path)

To Row and Col? This seems like wrapping a fast path in a slow path to me.

btracey:
The utility is that my external can take as input a Matrix, and I can extract a row using Row. If the matrix has a fast path (it is a *Dense or a Vectorer), this is automatically called during mat64.Row without the type-switch logic happening in my package code. This is the motivation for #38 in the first place.

btracey:
Huh, running benchmarks with several -cpu values causes the tests to fail. Didn't realize the benchmark cpu flags affected the tests in any way.

btracey:
(Looking into the cause as well, obviously)

kortschak:
> this is automatically called during mat64.Row without the type-switch logic happening in my package code

OK, so long as it is not replacing Vectorer switching in the arithmetic code in mat64.

btracey:
Agreed.

kortschak:
cpu flags are test-wide, not just for benchmarks. They are important for concurrency testing.

btracey:
Updated documentation for Vectorer and VectorSetter. This keeps the types in as they are likely useful, but documents that they should copy the full row.

kortschak:
I'd like to know the benchmark impact before I answer this last commit. It adds a bounds check each row or column in a path that, internally, already knows the bounds are satisfied. Externally, I can understand the reasoning, but it is unfortunate for internal calls.

mat64/matrix.go
```diff
@@ -88,32 +88,26 @@ type Mutable interface {
...
 
 // A VectorSetter can set rows and columns in the represented matrix.
 type VectorSetter interface {
-	// SetRow sets the values of the specified row to the values held in a slice of float64.
-	// It will panic if the index is out of bounds. The number of elements copied is
-	// returned and will be the minimum of the length of the slice and the number of columns
-	// in the matrix.
-	SetRow(i int, src []float64) int
+	// SetRow sets the values in the specified rows of the matrix to the values
+	// in src. len(src) must equal the numer of columns in the receiver.
```
kortschak: "number" (and below).

btracey:
What do you mean by "adds a bounds check"? The Vectorer interface previously specified that the routine would panic if the index was out of bounds, so the check was there already.

kortschak:
Ahh, sorry, I say the extra bounds check here https://github.com/gonum/matrix/pull/225/files#diff-bce3caa5fd269b104e5b15efb844e493R202 and extrapolated incorrectly.

btracey:
BenchmarkRow using Dense of various sizes would be enough to measure? It's probably both the most common case and will have the least overhead (so that the length check plausibly matters).

kortschak:
SGTM

btracey:
Benchmark pushed. bound.txt is with the bounds check uncommented in the "else" branch.
````
brendan:~/Documents/mygo/src/github.com/gonum/matrix/mat64$ benchstat bound.txt nobound.txt 
name     old time/op  new time/op  delta
Row10    54.0ns ± 1%  54.7ns ± 1%  +1.23%   (p=0.000 n=9+9)
Row100   65.6ns ± 1%  66.5ns ± 2%  +1.35%   (p=0.017 n=9+9)
Row1000   210ns ± 0%   210ns ± 2%    ~     (p=0.659 n=9+10)
````
Since the benchmarks seem to suggest having the bounds check is actually faster, seems like it has negligible impact.

kortschak:
LGTM after spelling fixed. Thanks.

PR #226, opened 2015-09-23
Fix putWorkspaceVec to put into Vector pool and not Dense pool

btracey:


btracey:
PTAL @kortschak

kortschak:
Yes, just saw that a second ago... whoops.

LGTM

PR #227, opened 2015-09-23
Make Dot a function

btracey:


btracey:
Updates #38

kortschak:
Updates #138.

LGTM

mat64/list_test.go
```diff
@@ -68,6 +68,15 @@ func sameAnswerFloat(a, b interface{}) bool {
 	return a.(float64) == b.(float64)
 }
 
+// sameAnswerFloat returns whether the two inputs are both NaN or within tol
```
vladimir-ch: sameAnswerFloatApprox

mat64/list_test.go
```diff
@@ -68,6 +68,15 @@ func sameAnswerFloat(a, b interface{}) bool {
...
+// of each other.
+func sameAnswerFloatApprox(a, b interface{}) bool {
+	if math.IsNaN(a.(float64)) {
+		return math.IsNaN(b.(float64))
```
vladimir-ch: I'm not sure how important it is but in case of NaN this approximate equality is not commutative.

mat64/matrix.go
```diff
@@ -383,6 +377,47 @@ func Det(a Matrix) float64 {
 	return math.Exp(det) * sign
 }
 
+// Dot returns the sum of the element-wise products of the elements of the
+// receiver and b. Dot panics if the matrix sizes are unequal.
```
vladimir-ch: There is no receiver anymore.

mat64/list_test.go
btracey: Why? It returns true if both are NaN, and returns false if either are NaN (unless there's a bug in the floats call)

vladimir-ch: Ah, I see, sorry for the noise. It's late here.

btracey: No problem. It's early here :)

PR #228, opened 2015-09-23
Make Sum a function and remove Sumer interface.

btracey:


btracey:
Updates #38  #138

kortschak:
Add TODO for adding cases for other blas64 type returning types.

LGTM

PR #230, opened 2015-09-24
mat64: remove interfaces

### LINKED ISSUE #138: mat64: consider retiring less used interface definitions

kortschak:
Fixes #138.

@btracey Please take a look.


mat64/dense.go
```diff
@@ -462,9 +449,9 @@ func (m *Dense) Stack(a, b Matrix) {
 }
 
 // Augment creates the augmented matrix of a and b, where b is placed in the
-// greater indexed columns.
-//
-// See the Augmenter interface for more information.
+// greater indexed columns. Augment will panic if the two input matrices do
+// not have the same number of rows or the augmented matrix is not the same
```
btracey: "or the constructed augmented matrix", this way there is no confusion with a being the augmented matrix.

btracey:
LGTM minus minor comment.

Another big issue down!

PR #233, opened 2015-09-25
mat64: add variadic product method

### LINKED ISSUE #121: mat64: Feature request for variadic matrix multiplication
This is a feature request for 
````
func (m *Dense) MulBatch(a ...Matrix)
````
The computation time for matrix multiplication can be significantly faster depending on the ordering. For example, multiplying
````
(n, 1) x (1, n) x (n, n)
````
It's much faster to do the (1,n) x (n,n) first, and then multiply the final two vectors, then to do the vector multiplication and then multiply the two (n,n) matrices. This problem can be solved through dynamic programming. In gonum, we can do even better than this because of the Matrix interface. We can use reflection to see the underlying Matrix type (SymDense/BandPacked/etc.) and use the order of operations counts from the multiplication. This allows automatic efficient multiplication order of arbitrary types (pretty neat)

We probably want the function signature to pair a matrix type with a "transpose", as complicated matrix multiply expressions almost always have them.

### PR DESCRIPTION
**Pull request by kortschak**
Fixes #121.

@btracey Please take a look.

### DIFF
--- a/mat64/product.go
+++ b/mat64/product.go
@@ -46,6 +46,7 @@ func (m *Dense) Product(factors ...Matrix) {
 
 	m.reuseAs(p.stack[0].Dims())
 	m.Copy(p.stack[0])
+	putWorkspace(p.stack[0])
 }
 
 // debugProductWalk enables debugging output for Product.
@@ -149,8 +150,8 @@ func (p *multiplier) multiplySubchain(i, j int) {
 
 	b := p.factor(j)
 	a := p.factor(i)
-	_, ac := a.Dims()
-	br, _ := b.Dims()
+	ar, ac := a.Dims()
+	br, bc := b.Dims()
 	if ac != br {
 		// Panic with a string since this
 		// is not a user-facing panic.
@@ -164,11 +165,15 @@ func (p *multiplier) multiplySubchain(i, j int) {
 			i, ar, ac, result(p.onStack[i]), j, br, bc, result(p.onStack[j]))
 	}
 
-	// TODO(kortschak) Use pool for r. This requires that
-	// we know whether a/b came from factors or the stack.
-	var r Dense
+	r := getWorkspace(ar, bc, false)
 	r.Mul(a, b)
-	p.push(&r, i, j)
+	if p.onStack[i] {
+		putWorkspace(a.(*Dense))
+	}
+	if p.onStack[j] {
+		putWorkspace(b.(*Dense))
+	}
+	p.push(r, i, j)
 }
 
 func (p *multiplier) push(m *Dense, i, j int) {
--- a/mat64/product.go
+++ b/mat64/product.go
@@ -42,11 +42,11 @@ func (m *Dense) Product(factors ...Matrix) {
 
 	p := newMultiplier(m, factors)
 	p.optimize()
-	p.multiply()
+	result := p.multiply()
 
-	m.reuseAs(p.stack[0].Dims())
-	m.Copy(p.stack[0])
-	putWorkspace(p.stack[0])
+	m.reuseAs(result.Dims())
+	m.Copy(result)
+	putWorkspace(result)
 }
 
 // debugProductWalk enables debugging output for Product.
@@ -131,13 +131,16 @@ func (p *multiplier) optimize() {
 }
 
 // multiply walks the optimal operation tree found by optimize,
-// leaving the final result in the stack.
-func (p *multiplier) multiply() {
+// leaving the final result in the stack. It returns the
+// product, which may be copied but should be returned to
+// the workspace pool.
+func (p *multiplier) multiply() *Dense {
 	p.multiplySubchain(0, len(p.factors)-1)
 	if debugProductWalk {
 		r, c := p.stack[0].Dims()
 		fmt.Printf("\tpop result (%d×%d) cost=%d\n", r, c, p.table.at(0, len(p.factors)-1).cost)
 	}
+	return p.stack[0]
 }
 
 func (p *multiplier) multiplySubchain(i, j int) {
--- a/mat64/product.go
+++ b/mat64/product.go
@@ -40,34 +40,7 @@ func (m *Dense) Product(factors ...Matrix) {
 		return
 	}
 
-	// Check size early, but don't yet allocate data.
-	if !m.isZero() {
-		if fr, _ := factors[0].Dims(); fr != r {
-			panic(ErrShape)
-		}
-		if _, lc := factors[len(factors)-1].Dims(); lc != c {
-			panic(ErrShape)
-		}
-	}
-
-	dims := make([]int, len(factors)+1)
-	dims[0] = r
-	dims[len(dims)-1] = c
-	for i, f := range factors[1:] {
-		_, pc := factors[i].Dims()
-		cr, _ := f.Dims()
-		dims[i+1] = cr
-		if pc != cr {
-			panic(ErrShape)
-		}
-	}
-
-	p := multiplier{
-		factors: factors,
-		dims:    dims,
-		table:   newTable(len(factors)),
-		onStack: make([]bool, len(factors)),
-	}
+	p := newMultiplier(m, factors)
 	p.optimize()
 	p.multiply()
 
@@ -80,13 +53,60 @@ const debugProductWalk = false
 
 // multiplier performs operation order optimisation and tree traversal.
 type multiplier struct {
+	// factors is the ordered set of
+	// factors to multiply.
 	factors []Matrix
-	dims    []int
+	// dims is the chain of factor
+	// dimensions.
+	dims []int
 
+	// table contains the dynamic
+	// programming costs and subchain
+	// division indices.
 	table table
 
+	// stack holds intermediate results
+	// in the product tree. onStack
+	// indicates whether a matrix to
+	// be multiplied is on the stack or
+	// in the input factors.
 	stack   []*Dense
 	onStack []bool
+}
+
+func newMultiplier(m *Dense, factors []Matrix) *multiplier {
+	// Check size early, but don't yet
+	// allocate data for m.
+	r, c := m.Dims()
+	fr, fc := factors[0].Dims() // newMultiplier is only called with len(factors) > 2.
+	if !m.isZero() {
+		if fr != r {
+			panic(ErrShape)
+		}
+		if _, lc := factors[len(factors)-1].Dims(); lc != c {
+			panic(ErrShape)
+		}
+	}
+
+	dims := make([]int, len(factors)+1)
+	dims[0] = r
+	dims[len(dims)-1] = c
+	pc := fc
+	for i, f := range factors[1:] {
+		cr, cc := f.Dims()
+		dims[i+1] = cr
+		if pc != cr {
+			panic(ErrShape)
+		}
+		pc = cc
+	}
+
+	return &multiplier{
+		factors: factors,
+		dims:    dims,
+		table:   newTable(len(factors)),
+		onStack: make([]bool, len(factors)),
+	}
 }
 
 // optimize determines an optimal matrix multiply operation order.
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: mat64/product.go [hunk 1]
```diff
@@ -0,0 +1,190 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package mat64
+
+import "fmt"
+
+// Product calculates the product of the given factors and places the result in
+// the receiver. The order of multiplication operations is optimized to minimize
```
btracey: Can we say something more vague in the comment? For example, if we have a smart algorithm that notes the difference in FLOPS (or clock time) between Gemm and Trmm we shouldn't be constrained not to use it.

mat64/product.go
```diff
@@ -0,0 +1,190 @@
...
+	//
+	// e.g. http://www.jofcis.com/publishedpapers/2014_10_10_4299_4306.pdf
+	//
+	// In the case that this is replaced, retain this code in
+	// tests to compare against.
+
+	r, c := m.Dims()
+	switch len(factors) {
+	case 0:
+		if r != 0 || c != 0 {
```
btracey: Is this the right behavior? Alternately it could just panic if the list has zero length. 

Stupid zero arguments.

mat64/product.go
```diff
@@ -0,0 +1,190 @@
...
+		if _, lc := factors[len(factors)-1].Dims(); lc != c {
+			panic(ErrShape)
+		}
+	}
+
+	dims := make([]int, len(factors)+1)
+	dims[0] = r
+	dims[len(dims)-1] = c
+	for i, f := range factors[1:] {
+		_, pc := factors[i].Dims()
```
btracey: We can cut the number of calls to Dims() in half by caching the column and comparing to the next row.

mat64/product.go
kortschak: I think it is the right behaviour. The product of 0 matrices is an absent matrix and out isZero state is that.

mat64/product.go
kortschak: I don't really want this to become that complicated. I think there is room elsewhere in gonum to do that kind of thing (for example the thing @jonlawlor is working on).

mat64/product.go
```diff
@@ -0,0 +1,190 @@
...
+	p.multiply()
+
+	m.reuseAs(p.stack[0].Dims())
+	m.Copy(p.stack[0])
+}
+
+// debugProductWalk enables debugging output for Product.
+const debugProductWalk = false
+
+// multiplier performs operation order optimisation and tree traversal.
```
btracey: American optimize?

btracey:
Were you working off of a written algorithm when you coded this? If so it would be easier for me to verify.

mat64/product.go
```diff
@@ -0,0 +1,190 @@
...
+
+// optimize determines an optimal matrix multiply operation order.
+func (p *multiplier) optimize() {
+	if debugProductWalk {
+		fmt.Printf("chain dims: %v\n", p.dims)
+	}
+	for f := 1; f < len(p.factors); f++ {
+		for i := 0; i < len(p.factors)-f; i++ {
+			j := i + f
+			p.table.set(i, j, entry{cost: int(^uint(0) >> 1)})
```
btracey: math.MaxInt32

mat64/product.go
btracey: I don't see the harm in being more vague.

kortschak:
There is a similar implementation in the WP article.

mat64/product.go
kortschak: When I'm in internal comments, I'll have my little counter cultural resistance.

mat64/product.go
kortschak: Why?

mat64/product.go
btracey: SGTM

mat64/product.go
btracey: Why not? It's much less opaque using the constant in Math.

mat64/product.go
kortschak: If I delete the basis clause, it's a lie, but I will do that.

btracey: We don't have to lie, I was thinking of saying something like "Product selects the order of matrix multiplications and so in many cases will be faster than writing individual multiplies by hand". We can leave as is also, I don't feel that strongly.

mat64/product.go
kortschak: It's not an int32. I'll add a constant.

btracey: It's not an int32, but neither is math.MaxInt32

kortschak: True, but math.MaxInt32 is not the largest value that can be held by int. Now if the client is striking that, they have problems, but the question will arise later about why math.MaxInt32 was used when the values could be higher.

btracey: Then why is math.MaxInt32 being used? http://play.golang.org/p/4vLKeBrbms

btracey: Oh, I understand now. Sorry.

btracey: It's fine to leave as is.

kortschak: To answer that question http://play.golang.org/p/XpkwnFWV_H amd64p32 is used on NaCl and there int is the same size as int32.

btracey: Yep. Exactly. I wasn't thinking until it hit me.

mat64/product.go
```diff
@@ -0,0 +1,190 @@
...
+	factors []Matrix
+	dims    []int
+
+	table table
+
+	stack   []*Dense
+	onStack []bool
+}
+
+// optimize determines an optimal matrix multiply operation order.
```
btracey: Would you mind adding some quick comments?
1) table.at(i,j).cost contains the cost of multiplying the matrices from i to j
2) table.at(i,j).k is which index to subdivide (I think, still parsing.)

mat64/product.go
```diff
@@ -0,0 +1,190 @@
...
+// leaving the final result in the stack.
+func (p *multiplier) multiply() {
+	p.walk(0, len(p.factors)-1)
+	if debugProductWalk {
+		r, c := p.stack[0].Dims()
+		fmt.Printf("\tpop result (%d×%d) cost=%d\n", r, c, p.table.at(0, len(p.factors)-1).cost)
+	}
+}
+
+func (p *multiplier) walk(i, j int) {
```
btracey: Could we call this something like "multiplySubset"?

mat64/product.go
```diff
@@ -0,0 +1,190 @@
...
+	if i == j {
+		return
+	}
+
+	p.walk(i, p.table.at(i, j).k)
+	p.walk(p.table.at(i, j).k+1, j)
+
+	var r Dense
```
btracey: I see that p.factor is some sort of optimisation, but I don't quite follow. My natural inclination would be something like
````
func (p *multiplier) multiplySubsequence(i,j int) Matrix {
     a := multiplySubsequence(i, p.table.at(i,j).k)
     b := mulitplySubsequence(i, p.table.at(i,j).k+1,j)
    var r Dense
    r.Mul(a,b)
    return r
}
````
Why is this algorithm better than mine above?

File: mat64/product.go (hunk 7 above)
Comment (kortschak) [inline mat64/product.go]: multiplySubchain

Comment (btracey) [inline mat64/product.go]: great.

File: mat64/product.go (hunk 8 above)
Comment (kortschak) [inline mat64/product.go]: Where are the factors being input in that recurrence?

File: mat64/product_test.go [hunk 9]
```diff
@@ -0,0 +1,142 @@
...
+
+type dims struct{ r, c int }
+
+var productTests = []struct {
+	n       int
+	factors []dims
+	product dims
+	panics  bool
+}{
+	{
```
btracey: Could you add some crazy chain like
````
factors: []dims{{1, 1000}{1000, 2}{2,2}}
````
That way if we change the algorithm, we have an obvious sign of something going wrong.

File: mat64/product.go (hunk 8 above)
Comment (btracey) [inline mat64/product.go]: I think 
````
if i == j{
     return factors[i]
}
````

Comment (kortschak) [inline mat64/product.go]: Yes, that will do it. I want to add pool handling. That makes this stack a little easier to use. I'll address the comments you have here, then make that change, then see about using the call stack instead of an explicit stack in multiplier.

File: mat64/product.go [hunk 10]
```diff
@@ -0,0 +1,190 @@
...
+
+package mat64
+
+import "fmt"
+
+// Product calculates the product of the given factors and places the result in
+// the receiver. The order of multiplication operations is optimized to minimize
+// the number of floating point operations on the basis that all matrix
+// multiplications are general.
+func (m *Dense) Product(factors ...Matrix) {
```
kortschak: Would you like me to return `(cost int)`?

mat64/product_test.go
btracey: Sorry, that doesn't work with naively testing the result. Nevermind.

mat64/product.go
btracey: I don't think so.

mat64/product.go
```diff
@@ -0,0 +1,190 @@
...
+	// The operation order optimisation is the naive O(n^3) dynamic
+	// programming approach and does not take into consideration
+	// finer-grained optimisations that might be available.
+	//
+	// TODO(kortschak) Consider using the O(nlogn) or O(mlogn)
```
btracey: I wonder what the practical limit of n is before underflow/overflow/numerical noise kills you. Given that these tests pass with 1e-14 tolerance, I suppose a lot.

btracey:
That's the end of my first round of comments. Tests look good. Thanks for putting this together!

kortschak:
PTAL

kortschak:
There was a failure in Solve. Maybe another travis weirdness.

btracey:
I think it's a random number / tolerance issue. I'm working on that fix right now.

mat64/product.go
```diff
@@ -0,0 +1,219 @@
...
+		r, c := p.stack[0].Dims()
+		fmt.Printf("\tpop result (%d×%d) cost=%d\n", r, c, p.table.at(0, len(p.factors)-1).cost)
+	}
+}
+
+func (p *multiplier) multiplySubchain(i, j int) {
+	if i == j {
+		return
+	}
+
```
btracey: Sorry if I'm bike shedding, but what about
````
func multiplySubchain(i,j int) (m Matrix, tmp bool) {
     if i == j {
         return p.factors[i], false
     }
     a, aTmp := p.multiplySubchain(i, p.table.at(i, j).k)
     b, bTmp := p.mulitplySubchain(p.table.at(i,j).k +1, j)
     r := getWorkspace(ar, bc, false)
     r.Mul(a,b)
     if aTmp{
          putWorkspace(a.(*Dense))   
     }
    if bTmp {
         putWorkspace(b.(*Dense))
    }
   return r, true
     
}

btracey:
LGTM except for the alternate recursion algorithm suggestion.

btracey:
Is it worth it to have "par bool" and compute the subset products concurrently?

kortschak:
Solve is still failing for me.

No, I don't think it's worthwhile - Mul already call parallelised code via your Dgemm.

btracey:
Huh. Do you see the failure locally? I don't. Looking at the traces, it seems to only be failing with a zero receiver. I don't see any reason why that should be true.

kortschak:
Yes. I see it locally. To make it pass with many iterations, I need to increase the tolerance to 1e-7, but only when many other tests are being run - only once in a long series at 1e-8.

kortschak:
PTAL I've made the stack change.

btracey:
Nice. Thanks. LGTM.

I guess we'll just have to loosen the Solve tolerance.

kortschak:
I think that will hide something. I suspect that something odd is going on. It's a rare failure and it only happens with other tests running (this could just be the PRNG state that that evolves though). Also, while I was looking, there is a commented out block that looks like it can be reactivated based on your recent work.

kortschak:
Thanks for the input.

btracey:
I ran into this problem full force when looking at Dgetri and friends (matrix inverse). Generating random matrices as we are now on average gives pretty ill conditioned matrices. Not necessarily enough to trigger the warning, but enough to change by more than a couple of digits depending on the solution procedure.

btracey:
Heh. https://twitter.com/brendantracey/status/647270044372631552

Issue #232, opened 2015-09-24
list two test should have comparison tolerance


btracey:
Different operations have different numerical stability. In Copy, the resulting matrices should be exactly equal, while Inverse is numerically unstable and needs a much higher tolerance. As one size fits all is not a good approach, each test should specify the desired tolerance.

PR #234, opened 2015-09-25
Add tolerance for testtwoinput equality

btracey:


btracey:
Fixes #234

btracey:
If you'd like I can delete equalApprox and use EqualApprox in this PR.

kortschak:
I was thinking of doing that previously, but we don't test EqualApprox AFAIR.

LGTM

kortschak:
This is #234. I don't understand the fixes comment.

btracey:
Sorry, I meant #232 

http://godoc.org/github.com/gonum/matrix/mat64#EqualApprox ?

kortschak:
No, I mean that when I was looking to delete equalApprox in the testing code it looked to me that it is a simpler algorithm (it does no type-dependent tricks), so it should be left.

btracey:
Ah, I see. Great.

Issue #178, opened 2015-08-09
mat64: Inverse should be a method, not a function


kortschak:
From https://groups.google.com/d/msg/gonum-dev/ytUeXNzSMgs/uaIPhicQFQAJ

PR #235, opened 2015-09-25
Delete inverse function and make a method on Dense

btracey:


btracey:
Fixes #178

mat64/dense_test.go
```diff
@@ -1243,6 +1243,52 @@ func (s *S) TestOuter(c *check.C) {
...
+			}),
+			tol: 1e-14,
+		},
+	} {
+		var got Dense
+		err := got.Inverse(test.a)
+		if err != nil {
+			c.Errorf("Error computing inverse: %v", err)
+		}
+		if !EqualApprox(&got, test.want, test.tol) {
```
kortschak: Can we have a wantEye.Mul(got, test.a) test? Also, I think equalApprox should be used for reasons I gave previously.

kortschak:
LGTM with minor comment.

btracey:
Failure is the same problem with Solve.

PR #236, opened 2015-09-25
Loosen solve tolerance.

btracey:
The problem is that randomly generated matrices can be very ill-conditioned. Loosen the solve tolerance until we have a better way of generating random matrices. An example of generating well-conditioned matrices is in the tests for Dgetri


kortschak:
LGTM with a sigh.

btracey:
Sounds about right.

PR #237, opened 2015-09-25
mat64: renable cond-dependent tests

kortschak:
@btracey Please take a look.

It's not entirely clear how these got into the state they were.


btracey:
LGTM

They got to there because for some reason Gels does not consider an all-zeros matrix to be singular (??). When the error checking was only on if it's exactly singular, Gels gave the wrong result, thus no error was returned and the tests failed. Now that we check the condition number, we do correctly return an error.

kortschak:
I was wondering why the matrices were transposed.

PR #248, opened 2015-09-28
mat64: remove gocheck dependency

kortschak:
The text of many of the test errors should be improved, but that can be a subsequent PR which should go over the tests and clean up some of the older style that persisted from biogo.matrix.

@btracey Please take a look.

### DIFF
--- a/mat64/dense_test.go
+++ b/mat64/dense_test.go
@@ -12,7 +12,6 @@ import (
 
 	"github.com/gonum/blas/blas64"
 	"github.com/gonum/floats"
-	"gopkg.in/check.v1"
 )
 
 func asDense(d *Dense) Matrix {
@@ -33,7 +32,7 @@ func asBasicTriangular(t *TriDense) Triangular {
 	return (*basicTriangular)(t)
 }
 
-func (s *S) TestNewDense(c *check.C) {
+func TestNewDense(t *testing.T) {
 	for i, test := range []struct {
 		a          []float64
 		rows, cols int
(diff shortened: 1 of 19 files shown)


jonlawlor:
Nice.

mat64/dense_test.go
```diff
@@ -881,29 +1017,30 @@ func (s *S) TestCloneT(c *check.C) {
...
-		r.Clone(a.T())
-		c.Check(Equal(&r, t), check.Equals, true, check.Commentf("Test %d: %v transpose = %v", i, test.a, test.t))
-
-		rr.Clone(r.T())
-		c.Check(Equal(&rr, a), check.Equals, true, check.Commentf("Test %d: %v transpose = %v", i, test.a, test.t))
-
-		zero(r.mat.Data)
-		r.Clone(a.T())
-		c.Check(Equal(&r, t), check.Equals, true, check.Commentf("Test %d: %v transpose = %v", i, test.a, test.t))
+		for j := 0; j < 2; j++ {
```
btracey: This is nice.

mat64/dense_test.go
```diff
@@ -1234,16 +1390,17 @@ func (s *S) TestOuter(c *check.C) {
...
-		c.Check(Equal(&m, want), check.Equals, true, check.Commentf("Test %v. Want %v, Got %v", i, want, m))
-		// Check with the same matrix
-		m.Outer(NewVector(len(test.x), test.x), NewVector(len(test.y), test.y))
-		c.Check(Equal(&m, want), check.Equals, true, check.Commentf("Test %v. Want %v, Got %v", i, want, m))
+		for j := 0; j < 2; j++ {
+			// Check with a new matrix - and then again.
+			m.Outer(NewVector(len(test.x), test.x), NewVector(len(test.y), test.y))
+			if !Equal(&m, want) {
+				t.Errorf("unexpected result for Outer test %d iteration %d: got: %+v want: %+v", i, j, m, want)
+			}
```
btracey: zero here?

btracey:
LGTM with minor comment. Thank you.

We should consider using floats.Same / floats.Equal for []float64 instead of reflect where appropriate.

mat64/dense_test.go
kortschak: That would be a logic change, and I didn't write the original test, so I don't want to do that. If you think there should have been a zero between the two calls (at L1240 in the original) then you should send a PR.

kortschak:
>We should consider using floats.Same / floats.Equal for []float64 instead of reflect where appropriate.

Agreed. There are many things to clean up, but now it can happen at a more reasonable pace.

btracey:
Yep, totally agreed. Just wanted to mention.

PR #252, opened 2015-09-29
mat64: fix SymOuterK and SymRankOne

### LINKED ISSUE #250: mat64: SymDense.SymOuterK should zero data slice
This fails:

```
func TestSymOuter(t *testing.T) {
	x := NewVector(5, []float64{1, 2, 3, 4, 5})
	var s1, s2 SymDense
	s1.SymOuterK(x)
	s2.SymOuterK(x)
	s2.SymOuterK(x)
	if !Equal(&s1, &s2) {
		t.Error("unexpected result from repeat")
	}
}
```

~~I suspect SymRankOne is also incorrect when using a zero matrix that has been used previously, since the data slice is not zeroed out - useZeroed should be used in the w.isZero() case.~~ There are other issues with SymRankOne though.

/cc @btracey

kortschak:
Fixes #250.

@btracey Please take a look.

### DIFF
--- a/mat64/symmetric.go
+++ b/mat64/symmetric.go
@@ -177,30 +177,24 @@ func (s *SymDense) CopySym(a Symmetric) int {
 // the result in the receiver
 //  s = a + alpha * x * x'
 func (s *SymDense) SymRankOne(a Symmetric, alpha float64, x *Vector) {
-	n := s.mat.N
-	if x.Len() != n {
+	n := x.Len()
+	if a.Symmetric() != n {
 		panic(ErrShape)
 	}
-	var w SymDense
-	if s == a {
-		w = *s
-	}
-	if w.isZero() {
-		w.mat = blas64.Symmetric{
+	if s.isZero() {
+		s.mat = blas64.Symmetric{
 			N:      n,
 			Stride: n,
 			Uplo:   blas.Upper,
-			Data:   use(w.mat.Data, n*n),
+			Data:   use(s.mat.Data, n*n),
 		}
-	} else if n != w.mat.N {
+	} else if n != s.mat.N {
 		panic(ErrShape)
 	}
 	if s != a {
-		w.CopySym(a)
+		s.CopySym(a)
 	}
-	blas64.Syr(alpha, x.mat, w.mat)
-	*s = w
-	return
+	blas64.Syr(alpha, x.mat, s.mat)
 }
 
 // SymRankK performs a symmetric rank-k update to the matrix a and stores the
@@ -236,9 +230,35 @@ func (s *SymDense) SymRankK(a Symmetric, alpha float64, x Matrix) {
 // SymRankOne
 //  s = x * x'
 func (s *SymDense) SymOuterK(x Matrix) {
-	r, _ := x.Dims()
-	s.reuseAs(r)
-	s.SymRankK(s, 1, x)
+	n, _ := x.Dims()
+	switch {
+	case s.isZero():
+		s.mat = blas64.Symmetric{
+			N:      n,
+			Stride: n,
+			Data:   useZeroed(s.mat.Data, n*n),
+			Uplo:   blas.Upper,
+		}
+		s.SymRankK(s, 1, x)
+	case s.mat.Uplo != blas.Upper:
+		panic(badSymTriangle)
+	case s.mat.N == n:
+		if s == x {
+			w := getWorkspaceSym(n, true)
+			w.SymRankK(w, 1, x)
+			s.CopySym(w)
+			putWorkspaceSym(w)
+		} else {
+			// Only zero the upper triangle.
+			for i := 0; i < n; i++ {
+				ri := i * s.mat.Stride
+				zero(s.mat.Data[ri+i : ri+n])
+			}
+			s.SymRankK(s, 1, x)
+		}
+	default:
+		panic(ErrShape)
+	}
 }
 
 // RankTwo performs a symmmetric rank-two update to the matrix a and stores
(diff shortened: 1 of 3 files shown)


kortschak:
This is more difficult than I thought it was going to be. The error in TestSymOuterK is one I'm having difficulties resolving - the answer should be easy, but it's evading me. The error in TestSymRankOne I think it an error in testTwoInput. Would you take a look, @btracey?

mat64/symmetric.go
```diff
@@ -177,8 +177,8 @@ func (s *SymDense) CopySym(a Symmetric) int {
 // the result in the receiver
 //  s = a + alpha * x * x'
 func (s *SymDense) SymRankOne(a Symmetric, alpha float64, x *Vector) {
-	n := s.mat.N
-	if x.Len() != n {
+	n := x.Len()
+	if a.Symmetric() != n {
 		panic(ErrShape)
 	}
 	var w SymDense
```
btracey: The code below is wrong. We allocate a new w matrix and then test if it's zero.
````
func (s *SymDense) SymRankOne(a Symmetric, alpha float64, x *Vector) {
	n := x.Len()
	if a.Symmetric() != n {
		panic(ErrShape)
	}
	if s.isZero() {
		s.mat = blas64.Symmetric{
			N:      n,
			Stride: n,
			Uplo:   blas.Upper,
			Data:   use(s.mat.Data, n*n),
		}
	} else if n != s.mat.N {
		panic(ErrShape)
	}
	if s != a {
		s.CopySym(a)
	}
	blas64.Syr(alpha, x.mat, s.mat)
}
````

File: mat64/symmetric.go [hunk 2]
```diff
@@ -237,7 +236,30 @@ func (s *SymDense) SymRankK(a Symmetric, alpha float64, x Matrix) {
 //  s = x * x'
 func (s *SymDense) SymOuterK(x Matrix) {
 	r, _ := x.Dims()
-	s.reuseAs(r)
+
```
btracey: The problem here is that there is no test for s == x. If this is the case, then zeroing s also zeros x

btracey: Basically, the equivalent of isolatedWorkspace is needed.

btracey:
I think list_test is correct, but somewhat misleading. Filed #253

mat64/symmetric.go
kortschak: Yeah, I tried that and list test said that s was unmodified in some cases (it must have been too late).

btracey: Interesting. I didn't actually implement it to verify. Would you like me to take a stab?

kortschak:
The issue with SymOuterK was tiredness, the issue with SymRankOne still exists - it is a consequence of assuming that an increment of size from a zero matrix will result in an invalid size. This is not always true.

kortschak:
There is some weird naming in list_test.go that is making things difficult, e.g. [zero](https://github.com/gonum/matrix/blob/master/mat64/list_test.go#L926) by name and provenance looks like it should have zero dims, but doesn't.

btracey:
I don't understand. We never assume that an increment from a zero-size matrix is an invalid size. testTwoInput assumes that an increment from a non-zero size is an incorrect size.  The line is
````
wrongSize := makeRandOf(receiver, rr+1, rc+1)
````
Where rr and rc are the resulting size of the operation. Since the input vector always has non-zero length, (the sizePairs are always positive), the resulting matrix from SymRankOne is always a non-zero matrix. 

In fact, this test should panic. If the input vector is size n, the resulting matrix is size nxn, and so an (n+1)x(n+1) matrix should panic.

The problem with SymRankOne is that we don't actually test it. The code now is

````
var w SymDense
if s == a {
   w = *s
}
if w.isZero() { 
   // etc
}
````
This only tests the size of the receiver if s == a. Otherwise it tests the size of w. w was just created as a zero value, so that test always passes. The code segment  https://github.com/gonum/matrix/pull/252/files#diff-a387f8fb664139ef10e122c64aa31b1fL184 changes to actually test the size of s. On my machine, if I download this PR as is and replace the SymRankOne code with my proposed code all of the tests pass.

**Comment by btracey** [comment]
zero is the receiver that starts with zero size and ends up having nonzero size because we adjust receiver sizes. I'd be happy with a different name, but it's in contrast to nonZero which starts with a non-zero size.

**Comment by kortschak** [comment]
Ahh, sorry, I must have been far too tired last night. When I read "The code below is wrong" I thought you meant the snippet you posted. Yes, fixed. Sometimes it's amazing how a thought train can be so canalised that the obvious answer (remove something) can be blocked by what is in front of you.

**Comment by kortschak** [comment]
> I'd be happy with a different name, but it's in contrast to nonZero which starts with a non-zero size.

"wasZero"

File: mat64/symmetric.go [hunk 3]
```diff
@@ -236,9 +230,34 @@ func (s *SymDense) SymRankK(a Symmetric, alpha float64, x Matrix) {
...
+	case s.mat.N == n:
+		if s == x {
+			w := getWorkspaceSym(n, true)
+			w.SymRankK(w, 1, x)
+			s.CopySym(w)
+			putWorkspaceSym(w)
+		} else {
+			// Only zero the upper triangle.
+			for i := 0; i < n; i++ {
+				zero(s.mat.Data[i*n+i : (i+1)*n])
```
btracey: This should use s.mat.Stride

btracey:
LGTM with the one comment.

mat64/symmetric.go
kortschak: Done.

kortschak:
Thanks for the help on this.

btracey:
No problem. Thanks for catching the bugs.

PR #254, opened 2015-09-29
mat64: add vector sizes and refactor size lists

### LINKED ISSUE #253: Missing sizes in list_test
List test is missing a relevent set of size tests. It needs
````
{5, 5, 5, 1},
{5, 5, 1, 5},
{5, 1, 5, 5},
{1, 5, 5, 5},
````
so that we can have a Symmetric and a Vector that are greater than size 1.

kortschak:
This does not fix the problem in #252 that prompted the issue.

Fixes #253.

@btracey Please take a look.


btracey:
LGTM

PR #260, opened 2015-10-01
mat64: add list_test coverage and make methods pass

kortschak:
Updates #240.

Modification for arithmetic methods was copy/paste from Add with operator change.

@btracey Please take a look.


btracey:
LGTM

PR #262, opened 2015-10-01
mat64: be strict with triangle for SymDense

### LINKED ISSUE #261: mat64: SymDense has paths coded that are not reachable
We don't support lower symmetric matrices via mat64, though we do via blas. However there are paths in Max, Min that cannot be reached because they examing the lower triangle of the data slice.

In itself this is not a problem, but that code cannot easily be tested (list_test tests won't allow it because they depend on CopySym which will panic when handed a lower-encoded symmetric matrix) and I don't see any real reason to keep it if there is no easy way to create a SymDense in this state.

When SetRawSymmetric is added it should panic if the blas64.Symmetric is lower (a similar check for blas.Unit should be done for SetRawTriangular when it appears).

kortschak:
Fixes #261.

@btracey Please take a look.

### DIFF
--- a/mat64/symmetric.go
+++ b/mat64/symmetric.go
@@ -20,7 +20,8 @@ var (
 
 const badSymTriangle = "mat64: blas64.Symmetric not upper"
 
-// SymDense is a symmetric matrix that uses Dense storage.
+// SymDense is a symmetric matrix that uses dense storage. SymDense
+// matrices are stored in the upper triangle.
 type SymDense struct {
 	mat blas64.Symmetric
 }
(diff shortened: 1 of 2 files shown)


mat64/symmetric.go
```diff
@@ -20,7 +20,8 @@ var (
 
 const badSymTriangle = "mat64: blas64.Symmetric not upper"
 
-// SymDense is a symmetric matrix that uses Dense storage.
+// SymDense is a symmetric matrix that uses dense storage. SymDense
+// matrices are encoded in upper the upper triangle.
```
btracey: in upper triangular format?

btracey: or just 'in the upper triangle'. The current sentence isn't english.

btracey:
LGTM with comment fixed.

kortschak:
Multitasking is a bad thing.

mat64/symmetric.go
kortschak: Done.

Issue #200, opened 2015-08-16
mat64: improve aliasing handling and documentation


kortschak:
Currently, we do almost no analysis of backing data aliasing. We can't cover all cases, but the cases we can handle (BLAS-backed types) are relatively cheap so we should do them. To cover the rest there should be a section in the package docs explaining the issue and our guarantees.

kortschak:
I have done some work to look at what we can do here and with a way of getting the addresses of data slices (though see discussion [here](https://groups.google.com/d/topic/golang-nuts/Uet5p_3JhZs/discussion)) we can determine whether blas64-backed matrices alias elements.

If we *can* safely get then the statement I propose we make is that operations that are potentially impacted by element aliasing will result in a temporary workspace being used if:

* the src and dst types implement one of the Raw methods, and
* the two types match in their blas64 backing type (i.e. their Raw methods match), and
* the address ranges of the data slice overlap, and
* there is an overlap in the elements of the two values or the strides of the two matrices differ. 

Otherwise the destination will be directly filled with the result of the operation on the source data.

This means that user-defined types that are backed by blas64 types but do not implement a Raw method are possibly subject to corruption - this should be made clear so that users provide the relevant Raw method in those cases. The rationale for this limit is that blas64 matrix types are the only real way we can get backing data addresses and are the direct link to the BLAS.

Here is a reasonable sketch for the code for the Dense type:

```
func (m *Dense) overlaps(a Matrix) bool {
	var aMat blas64.General
	if a, ok := a.(RawMatrixer); !ok {
		return false
	} else {
		if m == a {
			return true
		}
		aMat = a.RawMatrix()
	}
	mat := m.RawMatrix()

	// This block must be atomic with respect to GC moves.
	// At this stage this is true, because the GC does not
	// move.
	m0 := uintptr(unsafe.Pointer(&mat.Data[0]))
	mn := uintptr(unsafe.Pointer(&mat.Data[len(mat.Data)-1]))
	a0 := uintptr(unsafe.Pointer(&aMat.Data[0]))
	an := uintptr(unsafe.Pointer(&aMat.Data[len(aMat.Data)-1]))

	if m0 == a0 {
		// At least one element overlaps.
		return true
	}

	if mn < a0 {
		// We know m is completely before a.
		return false
	}
	if an < m0 {
		// We know m is completely after a.
		return false
	}

	if mat.Stride != aMat.Stride {
		// Too hard, so assume the worst.
		return true
	}

	if m0 < a0 {
[...]

PR #263, opened 2015-10-03
mat64: add overlap detection for blas64 returning types

kortschak:
This provides the necessary tools for adding shadowing detection. There is currently no testing for the symmetric or triangular matrix overlaps methods, but they are substantively the same as the general matrix case and it is not yet clear how views of those types will arise - clients can do it and I guess we might at some stage add ViewSym and ViewTri or some near equivalent.

The test is conservative in general (mismatched stride makes a yes if the matrix data slices overlap at all - this should never happen), but more so with the symmetric and triangular cases because of ease of coding.

Updates #200.

@btracey Please take a look.

### DIFF
--- a/mat64/dense_arithmetic.go
+++ b/mat64/dense_arithmetic.go
@@ -350,10 +350,16 @@ func (m *Dense) Mul(a, b Matrix) {
 	// temporary memory.
 	// C = A^T * B = (B^T * A)^T
 	// C^T = B^T * A.
-	if aU, ok := aU.(RawMatrixer); ok {
-		amat := aU.RawMatrix()
-		if bU, ok := bU.(RawMatrixer); ok {
-			bmat := bU.RawMatrix()
+	if aUrm, ok := aU.(RawMatrixer); ok {
+		amat := aUrm.RawMatrix()
+		if restore == nil {
+			m.checkOverlap(amat)
+		}
+		if bUrm, ok := bU.(RawMatrixer); ok {
+			bmat := bUrm.RawMatrix()
+			if restore == nil {
+				m.checkOverlap(bmat)
+			}
 			blas64.Gemm(aT, bT, 1, amat, bmat, 0, m.mat)
 			return
 		}
@@ -375,7 +381,7 @@ func (m *Dense) Mul(a, b Matrix) {
 			if aTrans {
 				c := getWorkspace(ac, ar, false)
 				var tmp Dense
-				tmp.SetRawMatrix(aU.RawMatrix())
+				tmp.SetRawMatrix(amat)
 				c.Copy(&tmp)
 				bT := blas.Trans
 				if bTrans {
@@ -412,8 +418,11 @@ func (m *Dense) Mul(a, b Matrix) {
 			return
 		}
 	}
-	if bU, ok := bU.(RawMatrixer); ok {
-		bmat := bU.RawMatrix()
+	if bUrm, ok := bU.(RawMatrixer); ok {
+		bmat := bUrm.RawMatrix()
+		if restore == nil {
+			m.checkOverlap(bmat)
+		}
 		if aU, ok := aU.(RawSymmetricer); ok {
 			amat := aU.RawSymmetric()
 			if bTrans {
@@ -432,7 +441,7 @@ func (m *Dense) Mul(a, b Matrix) {
 			if bTrans {
 				c := getWorkspace(bc, br, false)
 				var tmp Dense
-				tmp.SetRawMatrix(bU.RawMatrix())
+				tmp.SetRawMatrix(bmat)
 				c.Copy(&tmp)
 				aT := blas.Trans
 				if aTrans {
--- a/mat64/dense_arithmetic.go
+++ b/mat64/dense_arithmetic.go
@@ -29,18 +29,15 @@ func (m *Dense) Add(a, b Matrix) {
 	aMat, _ := untranspose(a)
 	bMat, _ := untranspose(b)
 	m.reuseAs(ar, ac)
-	var restore func()
-	if m == aMat {
-		m, restore = m.isolatedWorkspace(aMat)
-		defer restore()
-	} else if m == bMat {
-		m, restore = m.isolatedWorkspace(bMat)
-		defer restore()
-	}
 
-	if a, ok := a.(RawMatrixer); ok {
-		if b, ok := b.(RawMatrixer); ok {
-			amat, bmat := a.RawMatrix(), b.RawMatrix()
+	if arm, ok := a.(RawMatrixer); ok {
+		if brm, ok := b.(RawMatrixer); ok {
+			amat, bmat := arm.RawMatrix(), brm.RawMatrix()
+			if m != aMat {
+				m.checkOverlap(amat)
+			} else if m != bMat {
+				m.checkOverlap(bmat)
+			}
 			for ja, jb, jm := 0, 0, 0; ja < ar*amat.Stride; ja, jb, jm = ja+amat.Stride, jb+bmat.Stride, jm+m.mat.Stride {
 				for i, v := range amat.Data[ja : ja+ac] {
 					m.mat.Data[i+jm] = v + bmat.Data[i+jb]
@@ -48,6 +45,15 @@ func (m *Dense) Add(a, b Matrix) {
 			}
 			return
 		}
+	}
+
+	var restore func()
+	if m == aMat {
+		m, restore = m.isolatedWorkspace(aMat)
+		defer restore()
+	} else if m == bMat {
+		m, restore = m.isolatedWorkspace(bMat)
+		defer restore()
 	}
 
 	if a, ok := a.(Vectorer); ok {
@@ -84,18 +90,15 @@ func (m *Dense) Sub(a, b Matrix) {
 	aMat, _ := untranspose(a)
 	bMat, _ := untranspose(b)
 	m.reuseAs(ar, ac)
-	var restore func()
-	if m == aMat {
-		m, restore = m.isolatedWorkspace(aMat)
-		defer restore()
-	} else if m == bMat {
-		m, restore = m.isolatedWorkspace(bMat)
-		defer restore()
-	}
 
-	if a, ok := a.(RawMatrixer); ok {
-		if b, ok := b.(RawMatrixer); ok {
-			amat, bmat := a.RawMatrix(), b.RawMatrix()
+	if arm, ok := a.(RawMatrixer); ok {
+		if brm, ok := b.(RawMatrixer); ok {
+			amat, bmat := arm.RawMatrix(), brm.RawMatrix()
+			if m != aMat {
+				m.checkOverlap(amat)
+			} else if m != bMat {
+				m.checkOverlap(bmat)
+			}
 			for ja, jb, jm := 0, 0, 0; ja < ar*amat.Stride; ja, jb, jm = ja+amat.Stride, jb+bmat.Stride, jm+m.mat.Stride {
 				for i, v := range amat.Data[ja : ja+ac] {
 					m.mat.Data[i+jm] = v - bmat.Data[i+jb]
@@ -103,6 +106,15 @@ func (m *Dense) Sub(a, b Matrix) {
 			}
 			return
 		}
+	}
+
+	var restore func()
+	if m == aMat {
+		m, restore = m.isolatedWorkspace(aMat)
+		defer restore()
+	} else if m == bMat {
+		m, restore = m.isolatedWorkspace(bMat)
+		defer restore()
 	}
 
 	if a, ok := a.(Vectorer); ok {
@@ -140,18 +152,15 @@ func (m *Dense) MulElem(a, b Matrix) {
 	aMat, _ := untranspose(a)
 	bMat, _ := untranspose(b)
 	m.reuseAs(ar, ac)
-	var restore func()
-	if m == aMat {
-		m, restore = m.isolatedWorkspace(aMat)
-		defer restore()
-	} else if m == bMat {
-		m, restore = m.isolatedWorkspace(bMat)
-		defer restore()
-	}
 
-	if a, ok := a.(RawMatrixer); ok {
-		if b, ok := b.(RawMatrixer); ok {
-			amat, bmat := a.RawMatrix(), b.RawMatrix()
+	if arm, ok := a.(RawMatrixer); ok {
+		if brm, ok := b.(RawMatrixer); ok {
+			amat, bmat := arm.RawMatrix(), brm.RawMatrix()
+			if m != aMat {
+				m.checkOverlap(amat)
+			} else if m != bMat {
+				m.checkOverlap(bmat)
+			}
 			for ja, jb, jm := 0, 0, 0; ja < ar*amat.Stride; ja, jb, jm = ja+amat.Stride, jb+bmat.Stride, jm+m.mat.Stride {
 				for i, v := range amat.Data[ja : ja+ac] {
 					m.mat.Data[i+jm] = v * bmat.Data[i+jb]
@@ -159,6 +168,15 @@ func (m *Dense) MulElem(a, b Matrix) {
 			}
 			return
 		}
+	}
+
+	var restore func()
+	if m == aMat {
+		m, restore = m.isolatedWorkspace(aMat)
+		defer restore()
+	} else if m == bMat {
+		m, restore = m.isolatedWorkspace(bMat)
+		defer restore()
 	}
 
 	if a, ok := a.(Vectorer); ok {
@@ -196,18 +214,15 @@ func (m *Dense) DivElem(a, b Matrix) {
 	aMat, _ := untranspose(a)
 	bMat, _ := untranspose(b)
 	m.reuseAs(ar, ac)
-	var restore func()
-	if m == aMat {
-		m, restore = m.isolatedWorkspace(aMat)
-		defer restore()
-	} else if m == bMat {
-		m, restore = m.isolatedWorkspace(bMat)
-		defer restore()
-	}
 
-	if a, ok := a.(RawMatrixer); ok {
-		if b, ok := b.(RawMatrixer); ok {
-			amat, bmat := a.RawMatrix(), b.RawMatrix()
+	if arm, ok := a.(RawMatrixer); ok {
+		if brm, ok := b.(RawMatrixer); ok {
+			amat, bmat := arm.RawMatrix(), brm.RawMatrix()
+			if m != aMat {
+				m.checkOverlap(amat)
+			} else if m != bMat {
+				m.checkOverlap(bmat)
+			}
 			for ja, jb, jm := 0, 0, 0; ja < ar*amat.Stride; ja, jb, jm = ja+amat.Stride, jb+bmat.Stride, jm+m.mat.Stride {
 				for i, v := range amat.Data[ja : ja+ac] {
 					m.mat.Data[i+jm] = v / bmat.Data[i+jb]
@@ -215,6 +230,15 @@ func (m *Dense) DivElem(a, b Matrix) {
 			}
 			return
 		}
+	}
+
+	var restore func()
+	if m == aMat {
+		m, restore = m.isolatedWorkspace(aMat)
+		defer restore()
+	} else if m == bMat {
+		m, restore = m.isolatedWorkspace(bMat)
+		defer restore()
 	}
 
 	if a, ok := a.(Vectorer); ok {
--- a/mat64/dense_arithmetic.go
+++ b/mat64/dense_arithmetic.go
@@ -655,8 +655,13 @@ func (m *Dense) Scale(f float64, a Matrix) {
 	m.reuseAs(ar, ac)
 
 	aMat, aTrans := untranspose(a)
-	if a, ok := aMat.(RawMatrixer); ok {
-		amat := a.RawMatrix()
+	if rm, ok := aMat.(RawMatrixer); ok {
+		amat := rm.RawMatrix()
+		if m == aMat || m.checkOverlap(amat) {
+			var restore func()
+			m, restore = m.isolatedWorkspace(a)
+			defer restore()
+		}
 		if !aTrans {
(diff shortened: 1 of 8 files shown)


kortschak:
I've written in the specification for the aliasing behaviour, which is slightly more safe than the version written in the original issue - the docs include pointer equality (which was not originally included), mainly because it's easy. Please ensure I've been honest here - I may have missed some cases of that test since I only decided to formalise it half way through the series (I went back and amended some commits - hopefully I got all of them).

btracey:
High level discussion before any detail -- what behavior do we want? It seems to me like we want 1) To enable efficient computation, both in memory and clock cycles 2) To support and enforce correct code 3) To assist ease and readability

This PR has the mentality of "allocate defensively where possible, and permissive otherwise". Instead, we could have the mentality of "Don't allow overlap, unless we can prove that it's okay". We could say something along the lines of "Sharing data between src and dst is not allowed, unless it can be proven to be okay. As an exception to this rule, we allow sharing in the case of pointer equality". We can then comment along the lines of where you have now, that we can prove it is okay when the types are the same, the strides are the same, and there is no overlap.

I feel like "defensive panicking" may be better than "defensive allocating". You say that the strides should never be unequal, but it's really easy to construct
````
data := make([]float64, 64)
a := mat64.NewDense(8, 8, data)
b := mat64.NewDense(4, 16, data)
````
Or worse
````
c := mat64.NewDense(8, 4, data[:32])
````
If the user writes
````
c.Mul(a, d)
````
Do we really want to allocate defensively, or would we rather instead panic because of disallowed aliasing? It seems to me like allowing aliasing leads to code that can be very hard to reason about. Instead, we can just disallow it, provide the special case of pointer equality (because it's really useful), and have some decent aliasing detection (as in the case of equal stride but no overlap). If a user really wants to do crazy stuff with aliased src and dst matrices, we have the blas64 and lapack64 libraries for them to call directly.

**Comment by btracey** [comment]
To rephrase my comment above, we could have something like

````
Element Aliasing

Most methods in mat64 modify receiver data. It is forbidden for the modified data region of the receiver to overlap the used data area of the input arguments. The exception to this rule is if the method receiver is equal to one of the input arguments, as in the a.Pow(a, 6) call above. When data is shared between the receiver and input, mat64 will attempt to detect if the read and write regions overlap, but if it cannot prove no overlap occurs, the method will panic. At present, the following cases will not panic 
- Data doesn't overlap
- Receiver pointer equal to input pointer
- Other stuff you have (types match, strides equal, etc.)
````

We can even throw in the warning from the memory model: "If you need to read this section to understand the behavior of your program, you are being clever. Don't be clever. If you must be clever, blas64 and lapack64 may be used to call the behavior directly."

**Comment by kortschak** [comment]
The philosophical implications are pretty deep and we will be stuck with it when it's in and we freeze the API, so I want to be sure that we get this right. I think the decision about this should happen on gonum-dev. Two opinions is not enough - I'm not wedded to the PR as is, but one of the two approaches (or something better) does need to happen.

The changes necessary other than doc should be simple (your words are pretty close to what it should be if that approach is taken - I like the suggestion about cribbing from MM), though I need to check all the calls - it really should just be a replacement in the overlaps methods of "return true" with "panic(badRegion)" and a conditional panic at the tail, where badRegion has a meaningful text.

**Comment by btracey** [comment]
Agreed on all fronts.

**Comment by kortschak** [comment]
I've made the minimal change as an experiment and all tests pass except for TestSolveVec which I am trying to figure out (there are some additions to list_test that should come from this).

It seems that someone ;) is doing something clever by coercing a *Vector into a *Dense. This results in an exact match between the receiver and the parameter, but pointer difference.

**Comment by btracey** [comment]
Ohhhh, that is tricky. There are two solutions. 1) Duplicate the code 2) Test for pointer equality. If pointer equality holds, then call Dense.Solve with the same pointer, if it doesn't hold, then call Dense.Solve is it is.

**Comment by kortschak** [comment]
If I add the following code for the `off == 0` case, all tests pass, but that breaks the invariant that we don't allow operations to modify the parameter.

```
		if m.mat.Rows == a.Rows && m.mat.Cols == a.Cols && m.mat.Stride == a.Stride {
			return false
		}
```

**Comment by btracey** [comment]
You mean it breaks the invariant because it changes the data even if the pointers are unequal? We should probably be discouraging that kind of cleverness, especially since there's an easy workaround that makes the behavior explicit (what I mentioned with testing for equality).

**Comment by kortschak** [comment]
There is not opportunity to test for pointer equality by the time overlaps has the data, the blas.General (or other) type has already been extracted from the RawMatrixer.

We can fix the problem earlier though. This fixes it:

```
func (v *Vector) SolveVec(a Matrix, b *Vector) error {
	_, c := a.Dims()
	// The Solve implementation is non-trivial, so rather than duplicate the code,
	// instead recast the Vectors as Dense and call the matrix code.
	v.reuseAs(c)
	m := vecAsDense(v)
	bm := m
	if v != b {
		bm = vecAsDense(b)
	}
	return m.Solve(a, bm)
}
```

**Comment by btracey** [comment]
Yes, sorry, I should have been more clear. That is what I was thinking. I was actually thinking 
````
v.reuseAs(c)
m := vecAsDense(v)
if v == b {
   return m.Solve(a, m)
}
bm := vecAsDense(b)
return m.Solve(a, bm)
````
Which is what I meant by "if pointer equality holds, call Solve with the same pointer". Your code is better anyway.

**Comment by kortschak** [comment]
Also, I need to add check code for *Vector in the first commit here.

**Comment by kortschak** [comment]
I'll add that to the solve routines commit either way. If we go with permissive, it save an allocation.

**Comment by kortschak** [comment]
Approach discussion at the gonum-dev thread, please.

https://groups.google.com/d/topic/gonum-dev/hQiEE0K8Kac/discussion

**Comment by kortschak** [comment]
Counter proposal here: https://github.com/gonum/matrix/tree/shadow-strict

**Comment by kortschak** [comment]
A week has gone past with no input in either direction, so with off-line discussions, I'm changing the proposal to be strict rather than permissive.

@btracey PTAL

**Comment by kortschak** [comment]
@btracey Gentle ping. I'd like to get this out of the way since it touches many scattered lines.

File: mat64/dense_arithmetic.go [hunk 1]
```diff
@@ -30,17 +30,17 @@ func (m *Dense) Add(a, b Matrix) {
...
-		m, restore = m.isolatedWorkspace(bMat)
-		defer restore()
-	}
 
-	if a, ok := a.(RawMatrixer); ok {
-		if b, ok := b.(RawMatrixer); ok {
-			amat, bmat := a.RawMatrix(), b.RawMatrix()
+	if arm, ok := a.(RawMatrixer); ok {
+		if brm, ok := b.(RawMatrixer); ok {
+			amat, bmat := arm.RawMatrix(), brm.RawMatrix()
```
btracey: Given the algorithm, we don't need to isolate the workspace, just check overlap, correct?

mat64/dense_arithmetic.go
```diff
@@ -85,17 +93,17 @@ func (m *Dense) Sub(a, b Matrix) {
...
```
btracey: Same as add.

mat64/dense_arithmetic.go
```diff
@@ -141,17 +157,17 @@ func (m *Dense) MulElem(a, b Matrix) {
...
```
btracey: and here as well.

mat64/dense_arithmetic.go
```diff
@@ -251,13 +283,28 @@ func (m *Dense) Inverse(a Matrix) error {
...
+				tmp := getWorkspace(r, c, false)
+				tmp.Copy(a)
+				m.Copy(tmp)
+				putWorkspace(tmp)
+				break
+			}
+			m.Copy(a)
+		}
+	default:
+		if m != aMat {
```
btracey: If aMat is not a RawMatrixer, then it is impossible for m to equal aMat, isn't it? Thus it should just be m.Copy(a)?

mat64/dense_arithmetic.go
```diff
@@ -655,8 +711,13 @@ func (m *Dense) Scale(f float64, a Matrix) {
 	m.reuseAs(ar, ac)
 
 	aMat, aTrans := untranspose(a)
-	if a, ok := aMat.(RawMatrixer); ok {
-		amat := a.RawMatrix()
+	if rm, ok := aMat.(RawMatrixer); ok {
```
btracey: Here is the same as the add case.

mat64/dense_arithmetic.go
```diff
@@ -703,8 +761,13 @@ func (m *Dense) Apply(fn func(r, c int, v float64) float64, a Matrix) {
...
+		amat := rm.RawMatrix()
```
btracey: And here for the Add case as well I think.

mat64/doc.go
```diff
@@ -112,7 +112,33 @@
...
-// detect arbitrary overlap, and so care should be taken when using matrix views.
+// though in many cases this will cause an allocation (see Element Aliasing).
 // An exception to this rule is Copy, which does not allow a.Copy(a.T()).
+//
+// Element Aliasing
+//
+// Most methods in mat64 modify receiver data. It is forbidden for the modified
+// data region of the receiver to overlap the used data area of the input
+// arguments. The exception to this rule is when the method receiver is equal to one
+// of the input arguments, as in the a.Pow(a, 6) call above.
```
btracey: "Equal to one of the input arguments or its transpose"? Are we allowing a.Mul(a.T(), b)?

mat64/dense_arithmetic.go
kortschak: The idiom I've used here is explained in https://github.com/gonum/matrix/pull/263/files#diff-928774397024853172d191bc227c4939R23. It is somewhat misleading and I apologise for that; if given enough of an argument I would change it, but I would rather not.

In short, if checkOverlap does not return false, it panics, so the check is not one that will result in an isolation. The reason for doing it this way is that it keeps the check and the pointer comparison next to each other, which I think balances out the C-ism.

mat64/offset_appengine.go
```diff
@@ -0,0 +1,28 @@
...
+	a0 := reflect.ValueOf(a).Index(0).UnsafeAddr()
+	b0 := reflect.ValueOf(b).Index(0).UnsafeAddr()
+
+	if a0 == b0 {
+		return 0
+	}
+	if a0 < b0 {
+		return int((b0 - a0) / sizeOfFloat64)
+	}
+	return -int((a0 - b0) / sizeOfFloat64)
```
btracey: This doesn't look right. Hasn't the negative sign been added twice?

mat64/dense_arithmetic.go
kortschak: No, it can be a type conversion.

mat64/doc.go
kortschak: I'll add that, though that wording is not precise enough. I have to think about that.

mat64/offset_appengine.go
kortschak: a0 and b0 are uintptr, so I have to do arithmetic in natural space and then convert the sign.

mat64/shadow.go
```diff
@@ -0,0 +1,201 @@
...
+	mismatchedStrides = "mat64: bad region: different strides"
+)
+
+// checkOverlap returns false if the receiver does not overlap data elements
+// referenced by the parameter and panics otherwise.
+//
+// checkOverlap methods return a boolean to allow the check call to be added to a
+// boolean expression, making use of short-circuit operators.
+
+func (m *Dense) checkOverlap(a blas64.General) bool {
```
btracey: It's easy to test for all Raw types, isn't it? If this takes in Matrix, then a type switch can be used to extract the data slice from any blas type. If there is any overlap and the types are different, we panic. If both are RawMatrixers, then we can do the extended testing.

mat64/offset_appengine.go
btracey: Oh, sorry, I was thinking this returned the absolute value, but it does not.

mat64/dense_arithmetic.go
btracey: The idiom is nice, but my point is that if m == a, the algorithm without allocating an isolated workspace.

mat64/dense_arithmetic.go
btracey: Ah, tricky.

mat64/shadow.go
btracey: It seems like we could make it simple with something like

````
func (m *Dense) checkOverlap(a Matrix) bool {
    if m == a {
         return false
    }
     // Type switch to extract the data. If a is none of the Raw types, return false since we can't check.
    // If offset is in the right range, and the types are different, panic
    // If a is a RawMatrixer, do the fancier checks
}
````
This way, we can do something like
````
func (m *Dense) Method(a Matrix) {
    // Check size
    aMat, _ := untranspose(a)
    m.checkOverlap(aMat)
}
````
Thus we can check all of the necessary overlap at the top of the function once. We could even remove the boolean.

Comment (kortschak) [inline mat64/shadow.go]: This adds an extra assertion in many places. I considered it, but I don't like it.

File: mat64/dense_arithmetic.go (hunk 1 above)
Comment (kortschak) [inline mat64/dense_arithmetic.go]: If that's the case it was incorrect before this.

File: mat64/dense_arithmetic.go (hunk 5 above)
Comment (kortschak) [inline mat64/dense_arithmetic.go]: Tested. This one is required.

File: mat64/shadow.go (hunk 9 above)
Comment (btracey) [inline mat64/shadow.go]: I see. So the idea is that it's the users problem if the types are different and there is incorrect overlap. That's probably reasonable since you have to explicitly construct two different matrices with the same underlying data. The one exception is RowView and ColView of Dense. 
````
v := m.ColView(1)
v.SolveVec(m, v)
````
This is an easy case to construct, but will be incorrect. I'm not exactly sure what the correct conclusion is.

File: mat64/dense_arithmetic.go (hunk 1 above)
Comment (kortschak) [inline mat64/dense_arithmetic.go]: Done.

File: mat64/dense_arithmetic.go (hunk 2 above)
Comment (kortschak) [inline mat64/dense_arithmetic.go]: Done.

File: mat64/dense_arithmetic.go (hunk 3 above)
Comment (kortschak) [inline mat64/dense_arithmetic.go]: Done.

File: mat64/dense_arithmetic.go (hunk 6 above)
Comment (kortschak) [inline mat64/dense_arithmetic.go]: Also required.

File: mat64/shadow.go (hunk 9 above)
Comment (kortschak) [inline mat64/shadow.go]: This is why the documentation says what it says. Maybe we should special-case vectors?

**Comment by kortschak** [comment]
PTAL

File: mat64/shadow.go (hunk 9 above)
Comment (btracey) [inline mat64/shadow.go]: I'm just making sure I understand the intended behavior and its reasons (SGTM, just internalizing). But yea, possibly vectors should be special cased. Having ColView seems desirable.

Comment (kortschak) [inline mat64/shadow.go]: Thinking about this more. We must special-case vectors. The spirit of the documentation is that if you do things within the mat64 API we will protect you from yourself but if you do the equivalent of unsafe (which is why I originally called "RawMatrix" "UnsafeMatrix") then you are on your own. This is clearly within the API, and I would think possibly even something we should be permissive with, though again I'm not strongly bound to that position.

File: mat64/shadow.go [hunk 10]
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	if mat.Inc == 1 || off&mat.Inc == 0 {
+		panic(regionOverlap)
+	}
+	return false
+}
+
+// rectanglesOverlap returns whether the strided rectangles a and b overlap
+// when b is offset off elements after a but has at least one element before
```
btracey: Could you put "off" in quotes or rephrase? The documentation is really confusing when one tries to read "off" as a word instead of a variable name.

mat64/shadow.go
btracey: I don't think we should be permissive with vectors in the same way we are permissive in pointer equality. At best, it's confusing code. I do agree with your interpretation of the spirit of the rules (protect within the API), and thus the need to add a check for overlap with vectors.

mat64/shadow.go
```diff
@@ -0,0 +1,201 @@
...
+// checkOverlap methods return a boolean to allow the check call to be added to a
+// boolean expression, making use of short-circuit operators.
+
+func (m *Dense) checkOverlap(a blas64.General) bool {
+	mat := m.RawMatrix()
+	if cap(mat.Data) == 0 || cap(a.Data) == 0 {
+		return false
+	}
+
+	off := offset(mat.Data[:1], a.Data[:1])
```
btracey: Why the subslicing?

mat64/shadow.go
kortschak: No worries. When I go through and add vector shadowing in earnest I'll add checks for RawMatrixer overlap.

mat64/shadow.go
kortschak: We may have been handed something that is zero lengthed.

mat64/shadow.go
kortschak: Sure.

kortschak: Is "...offset by off elements after..." OK?

btracey: yes.

mat64/shadow.go
btracey: sgtm

kortschak: I've added a TODO to the docs to indicate this intention.

kortschak:
PTAL

mat64/doc.go
```diff
@@ -112,7 +112,34 @@
...
-// detect arbitrary overlap, and so care should be taken when using matrix views.
+// though in many cases this will cause an allocation (see Element Aliasing).
 // An exception to this rule is Copy, which does not allow a.Copy(a.T()).
+//
+// Element Aliasing
+//
+// Most methods in mat64 modify receiver data. It is forbidden for the modified
+// data region of the receiver to overlap the used data area of the input
+// arguments. The exception to this rule is when the method receiver is equal to one
+// of the input arguments, as in the a.Pow(a, 6) call above, or its implicit transpose.
```
btracey: I would add something like "mat64 attempts to check for data overlap, for example as constructed with View", possibly in the paragraph saying "don't be clever". Basically, we want to say "mat64 does checking to help you not make mistakes. There are ways to make mistakes using our API, and we try and detect and complain about those. There are lots of ways to create data overlap by circumventing the API. If you need to read this section to see if your overlap is okay, you are being clever. Don't be clever. If you need ... lapack64 ... directly".

In short, I think the current documentation doesn't do enough to explain the goal before telling people not to be clever to get around that goal.

mat64/doc.go
```diff
@@ -112,7 +112,34 @@
...
+//
+// If you need to read this section to understand the behavior of your program,
+// you are being clever. Don't be clever. If you must be clever, blas64 and
+// lapack64 may be used to call the behavior directly.
+//
+// When data is shared between the receiver and input, mat64 will attempt to detect
```
btracey: My recommendation for this documentation is:

"mat64 will use the following rules to detect overlap between the receiver and one of the inputs:
 * the input implements one of the Raw methods and
 * the Raw type matches that of the receiver and
 * the address ranges of the backing data slices overlap and
 * either the strides differ or there is an overlap in the used data elements 
If such  an overlap is detected, the method will panic.

The following cases will not panic:
 * the data slices do not overlap 
 * there is pointer identity between the receiver and input values after
    the value has been untransposed if necessary

mat64 does not detect element overlap if the input does not implement a Raw method or if the Raw method differs from that of the receiver. Correct method behavior cannot be guaranteed if there is overlap.

Currently only RawMatrixer aliasing detection is supported."

kortschak: >mat64 does not detect element overlap if the input does not implement a Raw method or if the Raw method differs from that of the receiver.

This seems redundant. Is it just there to provide emphasis for the "Correct method behavior cannot be guaranteed if there is overlap" part?

btracey: > On Oct 18, 2015, at 6:29 PM, Dan Kortschak <notifications@github.com> wrote:
> 
> In mat64/doc.go <https://github.com/gonum/matrix/pull/263#discussion_r42329418>:
> 
> >  // An exception to this rule is Copy, which does not allow a.Copy(a.T()).
> > +//
> > +// Element Aliasing
> > +//
> > +// Most methods in mat64 modify receiver data. It is forbidden for the modified
> > +// data region of the receiver to overlap the used data area of the input
> > +// arguments. The exception to this rule is when the method receiver is equal to one
> > +// of the input arguments, as in the a.Pow(a, 6) call above, or its implicit transpose.
> > +//
> > +// If you need to read this section to understand the behavior of your program,
> > +// you are being clever. Don't be clever. If you must be clever, blas64 and
> > +// lapack64 may be used to call the behavior directly.
> > +//
> > +// When data is shared between the receiver and input, mat64 will attempt to detect
> mat64 does not detect element overlap if the input does not implement a Raw method or if the Raw method differs from that of the receiver.
> 
> This seems redundant. Is it just there to provide emphasis for the "Correct method behavior cannot be guaranteed if there is overlap" part?
> 
> 

Yes, basically. The idea is “We know you’re wrong if X. We guarantee you’re fine if Y. You’re wrong but we don’t tell you if Z"

kortschak: I've modified that section slightly (your words were nicely concise for the rest) to make it really clear what we mean, and to leave open the addition of Vector/Dense checking.

mat64/doc.go
```diff
@@ -112,7 +112,47 @@
...
 // An exception to this rule is Copy, which does not allow a.Copy(a.T()).
+//
+// Element Aliasing
+//
+// Most methods in mat64 modify receiver data. It is forbidden for the modified
+// data region of the receiver to overlap the used data area of the input
+// arguments. The exception to this rule is when the method receiver is equal to one
+// of the input arguments, as in the a.Pow(a, 6) call above, or its implicit transpose.
+//
+// This prohibition is to help avoid subtle mistakes resulting from input corruption
```
btracey: 'subtle mistakes when the method needs to read from and write to the same data region. There are many ways ...'

btracey:
Nice documentation wording.

LGTM with the one minor documentation change.

mat64/doc.go
kortschak: Nice.

kortschak:
Thank you. Very good review.

btracey:
Thank you for putting this together!

PR #265, opened 2015-10-07
Add Vector.ScaleVec

vladimir-ch:
Is it ok like this?

### DIFF
--- a/mat64/list_test.go
+++ b/mat64/list_test.go
@@ -112,6 +112,7 @@ func legalTypesAll(a, b Matrix) bool {
 	return true
 }
 
+// legalTypeSym returns whether a is a Symmetric.
 func legalTypeSym(a Matrix) bool {
 	_, ok := a.(Symmetric)
 	return ok
@@ -126,6 +127,12 @@ func legalTypesSym(a, b Matrix) bool {
 		return false
 	}
 	return true
+}
+
+// legalTypeVec returns whether v is a *Vector.
+func legalTypeVec(v Matrix) bool {
+	_, ok := v.(*Vector)
+	return ok
 }
 
 // legalTypesNotVecVec returns whether the first input is an arbitrary Matrix
--- a/mat64/list_test.go
+++ b/mat64/list_test.go
@@ -66,6 +66,11 @@ func isAnySize2(ar, ac, br, bc int) bool {
 	return true
 }
 
+// isAnyVector returns true for any column vector sizes.
+func isAnyVector(ar, ac int) bool {
+	return ac == 1
+}
+
 // isSquare returns whether the input matrix is square.
 func isSquare(r, c int) bool {
 	return r == c
--- a/mat64/vector_test.go
+++ b/mat64/vector_test.go
@@ -192,6 +192,17 @@ func TestVectorScale(t *testing.T) {
 			t.Errorf("test %d: unexpected result for v = alpha * v: got: %v want: %v", i, v.RawVector(), test.want.RawVector())
 		}
 	}
+
+	for _, alpha := range []float64{0, 1, -1, 2.3, -2.3} {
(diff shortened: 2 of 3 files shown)


kortschak:
The signature doesn't match the convention, it should be v.Scale(alpha float64, a *Vector) where the dot is read as an assignment.

The full work would be to check a.Len() == v.Len() or v.isZero(), or a == v, then as appropriate copy a to v (not if a == v) and then call Scal on v as you have here.

vladimir-ch:
Changed as suggested. PTAL @kortschak

kortschak:
LGTM

mat64/vector_test.go
```diff
@@ -134,6 +134,66 @@ func TestVectorMul(t *testing.T) {
...
+		if !reflect.DeepEqual(v.RawVector(), test.want.RawVector()) {
+			t.Errorf("test %d: unexpected result for v = alpha * a: got: %v want: %v", i, v.RawVector(), test.want.RawVector())
+		}
+
+		v.CopyVec(test.a)
+		v.ScaleVec(test.alpha, &v)
+		if !reflect.DeepEqual(v.RawVector(), test.want.RawVector()) {
+			t.Errorf("test %d: unexpected result for v = alpha * v: got: %v want: %v", i, v.RawVector(), test.want.RawVector())
+		}
+	}
```
btracey: Could you add a call to testOneInput as well for testing please?

vladimir-ch: Using `testOneInput` for the first time, so could you please take another look, @btracey ?

mat64/vector_test.go
```diff
@@ -134,6 +134,77 @@ func TestVectorMul(t *testing.T) {
...
+
+	for _, alpha := range []float64{0, 1, -1, 2.3, -2.3} {
+		method := func(receiver, a Matrix) {
+			v := receiver.(*Vector)
+			v.ScaleVec(alpha, a.(*Vector))
```
btracey: We've been using interfaces here, see for example TestVectorMul

mat64/vector_test.go
```diff
@@ -134,6 +134,77 @@ func TestVectorMul(t *testing.T) {
...
+		}
+		denseComparison := func(receiver, a *Dense) {
+			receiver.Scale(alpha, a)
+		}
+		testOneInput(t, "ScaleVec", &Vector{}, method, denseComparison, legalTypeVec, isAnySize, 0)
```
btracey: This works, but sort of by accident. I'd rather see a "isAnyVector" than isAnySize, that way the fact that it must be a vector is explicitly declared in the size(

vladimir-ch: Ok, I'll add it. At first I didn't because `legalDims` does that and it felt like code duplication.

btracey: Yea, I don't know. On the one hand, you're definitely right. On the other hand, isAnySize is generally for matrices, so it's somewhat confusing when reading other versions of the test. Not clear which is better.

kortschak: Being overly tight I think is the way to go - see the obsessive name changes and nilling I did recently.

vladimir-ch: Done. It is probably better to be explicit. PTAL @btracey

mat64/list_test.go
```diff
@@ -66,6 +66,11 @@ func isAnySize2(ar, ac, br, bc int) bool {
 	return true
 }
 
+// isAnyVector returns true for any row or column vector sizes.
```
btracey: I think this should just check if the number of columns is 1. *Vector is always a column.

btracey:
LGTM with stricter definition of "AnyVector"

mat64/list_test.go
vladimir-ch: Done

Issue #214, opened 2015-09-15
Test behavior of zero-sized matrices


btracey:
list_test.go does not test zero sized matrices. I quickly checked adding them and the testing suite panics. We should make sure there is a fixed behavior when there is a zero sized argument (at least on a per function/method basis).

PR #266, opened 2015-10-07
Add special case for *Vector in mat64.Norm

vladimir-ch:
What about this, is it ok? That `case *Vector:` doesn't blend in with the other `Raw...er` cases but unlike the other blas64 types, Vector does not carry enough information for the blas calls. Would relying on Dims() be safe?

### DIFF
--- a/mat64/matrix_test.go
+++ b/mat64/matrix_test.go
@@ -263,10 +263,15 @@ func TestNormZero(t *testing.T) {
 		&TriDense{mat: blas64.Triangular{Uplo: blas.Upper, Diag: blas.NonUnit}},
 		&Vector{},
 	} {
+		want := 0.0
 		for _, norm := range []float64{1, 2, math.Inf(1)} {
-			panicked, message := panics(func() { Norm(a, norm) })
+			var got float64
+			panicked, message := panics(func() { got = Norm(a, norm) })
 			if panicked {
 				t.Errorf("unexpected panic for Norm(&%T{}, %v): %v", a, norm, message)
+			}
+			if got != want {
+				t.Errorf("unexpected result for Norm(&%T{}, %v). Want %v, got %v", a, norm, want, got)
 			}
 		}
 	}
--- a/mat64/matrix_test.go
+++ b/mat64/matrix_test.go
@@ -8,6 +8,9 @@ import (
 	"fmt"
 	"math"
 	"testing"
+
+	"github.com/gonum/blas"
+	"github.com/gonum/blas/blas64"
 )
 
 var inf = math.Inf(1)
@@ -251,6 +254,22 @@ func TestNorm(t *testing.T) {
 		return Norm(a, math.Inf(1))
 	}
 	testOneInputFunc(t, "Norm_inf", f, denseComparison, sameAnswerFloatApprox, isAnyType, isAnySize)
+}
+
+func TestNormZero(t *testing.T) {
+	for _, a := range []Matrix{
+		&Dense{},
+		&SymDense{mat: blas64.Symmetric{Uplo: blas.Upper}},
+		&TriDense{mat: blas64.Triangular{Uplo: blas.Upper, Diag: blas.NonUnit}},
+		&Vector{},
+	} {
+		for _, norm := range []float64{1, 2, math.Inf(1)} {
+			panicked, message := panics(func() { Norm(a, norm) })
+			if panicked {
+				t.Errorf("unexpected panic for Norm(&%T{}, %v): %v", a, norm, message)
+			}
+		}
+	}
 }
 
 func TestSum(t *testing.T) {
--- a/mat64/vector.go
+++ b/mat64/vector.go
@@ -85,7 +85,12 @@ func (v *Vector) ViewVec(i, n int) *Vector {
 	}
 }
 
-func (v *Vector) Dims() (r, c int) { return v.n, 1 }
+func (v *Vector) Dims() (r, c int) {
+	if v.isZero() {
+		return 0, 0
+	}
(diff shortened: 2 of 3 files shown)


kortschak:
It's OK to use Vector.n.

kortschak:
LGTM

We should make Norm return NaN for zero sized input, or panic more meaningfully than we do now.

kortschak:
Sorry, not so fast. I just checked the others and they don't panic. This will I think. I don't think it should.

kortschak:
Please add the following test and make it pass.

```
func TestNormZero(t *testing.T) {
	for _, a := range []Matrix{
		&Dense{},
		&SymDense{mat: blas64.Symmetric{Uplo: blas.Upper}},
		&TriDense{mat: blas64.Triangular{Uplo: blas.Upper, Diag: blas.NonUnit}},
		&Vector{},
	} {
		for _, norm := range []float64{1, 2, math.Inf(1)} {
			panicked, message := panics(func() { Norm(a, norm) })
			if panicked {
				t.Errorf("unexpected panic for Norm(&%T{}, %v): %v", a, norm, message)
			}
		}
	}
}
```

I'm not sure what the expected value that should be returned is, maybe NaN, then you can make the test more stringent and then get rid of the fields being set in SymDense and TriDense cases - (currently they panic with illegal triangle when there is no triangle set, and TriDense with illegal diagonal if the triangle is set correctly).

Alternatively, panic informatively in all cases.

@btracey?

btracey:
This relates to #214

I think we should have consistent behavior with zero sized matrices as inputs. In particular, either we should always panic or we should try to do whatever makes the most sense. An exception is Copy, but copy is special in many ways so I think that's okay.

I lean toward panic. For one, a lot of operations don't make sense with zero-sized matrices (norm being a great example), and for two we treat zero sized matrices specially when they are the receiver. For a non-zero matrix, using it as a receiver and as an input is "size preserving", while for a zero matrix it is not. This could possibly lead to silent bugs where things seem okay because the receiver has taken a certain size even though it had started out zero. I don't feel particularly strongly, especially if there are other compelling arguments, but I do think being as consistent as possible across zero size is for the best.

vladimir-ch:
Ok, I added the test and made it pass by returning (for now) 0 when `v.isZero()`. Zero is consistent with what lapack returns for zero-sized input.

I noticed a comment in `isZero()` that says that `v.Dims()` must return zeros if `isZero()` returns true but v.Dims() returns 0, 1 in such case. It should be fixed I guess.

kortschak:
That comment is an overstatement, but I think it would be nice to follow, purely for client sanity.

vladimir-ch:
PTAL @kortschak

mat64/vector.go
```diff
@@ -324,7 +329,7 @@ func (v *Vector) reuseAs(r int) {
 
 func (v *Vector) isZero() bool {
 	// It must be the case that v.Dims() returns
-	// zeros in this case. See comment in Reset().
```
kortschak: Add this comment back, but fix it so that there is a comment there. The text here presumably came from what exists in Dense: https://github.com/gonum/matrix/blob/master/mat64/dense.go#L315

mat64/matrix_test.go
```diff
@@ -253,6 +256,22 @@ func TestNorm(t *testing.T) {
...
 
+func TestNormZero(t *testing.T) {
+	for _, a := range []Matrix{
+		&Dense{},
+		&SymDense{mat: blas64.Symmetric{Uplo: blas.Upper}},
+		&TriDense{mat: blas64.Triangular{Uplo: blas.Upper, Diag: blas.NonUnit}},
+		&Vector{},
+	} {
+		for _, norm := range []float64{1, 2, math.Inf(1)} {
+			panicked, message := panics(func() { Norm(a, norm) })
```
kortschak: Add a closed-over var to capture the result and compare that against expectation.

mat64/vector.go
```diff
@@ -97,6 +102,10 @@ func (v *Vector) T() Matrix {
 	return Transpose{v}
 }
 
+// Reset zeros the length of the vector so that it can be reused as the
+// receiver of a dimensionally restricted operation.
+//
+// See the Reseter interface for more information.
 func (v *Vector) Reset() {
 	v.mat.Data = v.mat.Data[:0]
```
kortschak: Sorry, I meant the internal comment. Altered here to reflect the difference in the two types:

```
	// No change of Inc or n to 0 may be
	// made unless both are set to 0.
```

kortschak:

I'm troubled by the notion of a valid norm of a zero-sized entity, though the definitions do indicate that this is possible and that its value is 0. As a counter to the LAPACK case, numpy.linalg.norm fails for the zero-sized case with one and inf norms (though not for the Frobenius norm - let's not follow their example here in being inconsistent).

I think that we either return NaN or panic.

mat64/vector.go
vladimir-ch: Ah, ok. Added.

mat64/matrix_test.go
vladimir-ch: Done with the expectation being for the moment a zero.

vladimir-ch:
It seems to me that a zero-sized entity is more of a programming rather than mathematical concept (What linear transformation would a zero-sized matrix represent? What would be R^0?). It seems strange that a norm of such entity would be the same as a norm of a zero matrix or vector. If we return NaN and it gets used in a computation, it will infect it and render it useless with the source of NaN being harder to find. So we might just as well panic directly, especially if the only way to create a zero-sized anything is either by `Reset()` or by `var a mat64.Vector`.

PR #268, opened 2015-10-09
Add Cond for computing the condition number of a matrix

btracey:


btracey:
I'm not sure what to do about testing Condition for m != n  with the 1 and inf norms. Other packages (i.e. Matlab) only let you compute the two norm for non-square matrices. On the one hand, that could be reasonable behavior here, but we determine the linear solve condition number based on the 1 norm (because Trcon only does the 1 and inf norms), so it feels like we should allow clients to compute the same number we do. On the other hand, Trcon is also incorrect (see the comment in Dtrcon test). On the other other hand, Trcon gives roughly the right order of magnitude, which is all that really matters.

kortschak:
It seems to me that this is a perfect place for a [BUG comment](http://blog.golang.org/godoc-documenting-go-code).

btracey:
Added.

btracey:
This is a real failure that is reproduced on my machine on 1.3.3. A sufficiently big matrix is not being constructed.

btracey:
I see. At some point we're putting a matrix with cap of 3 into the pool, when it should have a cap of 4.

btracey:
Ah. SVD is changing the cap of the matrix.

btracey:
Replaced it with a call to NewDense and we can fix when SVD is updated.

mat64/matrix.go
```diff
@@ -274,6 +274,79 @@ func Row(dst []float64, i int, a Matrix) []float64 {
 	return dst
 }
 
+// Cond returns the condition number of the given square matrix under the given
+// norm. The condition number must be based on the 1-norm, 2-norm or ∞-norm.
+//
+// BUG(btracey): The computation of the 1-norm and ∞-norm for non-square matrices
+// is innacurate (though typically the right order of magnitude). See
```
kortschak: Use comma instead of parentheses; s/ (though typically the right order of magnitude)/. although is typically the right order of magnitude/.

mat64/matrix.go
```diff
@@ -274,6 +274,79 @@ func Row(dst []float64, i int, a Matrix) []float64 {
 	return dst
 }
 
+// Cond returns the condition number of the given square matrix under the given
```
kortschak: s/square//
You don't place that condition in the code.

mat64/matrix_test.go
```diff
@@ -126,6 +127,83 @@ func TestRow(t *testing.T) {
...
+		if !floats.EqualWithinAbsOrRel(test.condTwo, condTwo, 1e-13, 1e-13) {
+			t.Errorf("Case %d: inf norm mismatch. Want %v, got %v", i, test.condTwo, condTwo)
+		}
+		condInf := Cond(test.a, math.Inf(1))
+		if !floats.EqualWithinAbsOrRel(test.condInf, condInf, 1e-13, 1e-13) {
+			t.Errorf("Case %d: inf norm mismatch. Want %v, got %v", i, test.condInf, condInf)
+		}
+	}
+
+	f := func(a Matrix) interface{} {
```
kortschak:
```
for _, cond := range []struct{
    name string
    norm float64
} {
    {"CondOne", 1},
    {"CondTwo", 2},
    {"CondInf", math.Inf(1)},
} {
    ...
```

kortschak:

mat64/matrix.go

mat64/matrix.go

mat64/matrix_test.go
btracey: Done.

mat64/matrix_test.go
```diff
@@ -126,6 +127,85 @@ func TestRow(t *testing.T) {
...
+			t.Errorf("Case %d: inf norm mismatch. Want %v, got %v", i, test.condInf, condInf)
+		}
+	}
+
+	for _, test := range []struct {
+		name string
+		norm float64
+	}{
+		{
+			name: "CondOne",
```
kortschak: For future reference, it would be good to keep this on a single line, and names aren't really required.

The reason for the single line is that it is protective against a change against one field that doesn't match the other; the struct value is essentially an atomic entity here. The names aren't necessary since the []struct is is very local.

btracey: Okay, thanks.

PR #270, opened 2015-10-11
mat64: make zero sized matrices panic for Norm, Cond, Min and Max

kortschak:
See discussion at #266.


btracey:
LGTM

PR #272, opened 2015-10-13
Add Vector.AddScaledVec

vladimir-ch:
Without this addition, some operations on Vectors require a temporary with the present Vector API.

Possibly a better name for the method would be `ScaleAddVec`.

`Add` and `Sub` could be removed but probably we will keep them for convenience. If kept, they should call the new method.

@btracey @kortschak Please take a look.

### DIFF
--- a/mat64/list_test.go
+++ b/mat64/list_test.go
@@ -56,6 +56,12 @@ func legalSizeSolve(ar, ac, br, bc int) bool {
 	return ar == br
 }
 
+// legalSizeSameVec returns whether the two matrices are column vectors of the
+// same dimension.
+func legalSizeSameVec(ar, ac, br, bc int) bool {
+	return ac == 1 && bc == 1 && ar == br
+}
+
 // isAnySize returns true for all matrix sizes.
 func isAnySize(ar, ac int) bool {
 	return true
@@ -138,6 +144,17 @@ func legalTypesSym(a, b Matrix) bool {
 func legalTypeVec(v Matrix) bool {
 	_, ok := v.(*Vector)
 	return ok
+}
+
+// legalTypesVecVec returns whether both inputs are *Vector.
+func legalTypesVecVec(a, b Matrix) bool {
+	if _, ok := a.(*Vector); !ok {
+		return false
+	}
+	if _, ok := b.(*Vector); !ok {
+		return false
+	}
+	return true
 }
 
 // legalTypesNotVecVec returns whether the first input is an arbitrary Matrix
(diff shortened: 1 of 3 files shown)


mat64/list_test.go
```diff
@@ -140,6 +146,13 @@ func legalTypeVec(v Matrix) bool {
 	return ok
 }
 
+// legalTypesVecVec returns whether both inputs are *Vector.
+func legalTypesVecVec(a, b Matrix) bool {
+	_, oka := a.(*Vector)
```
kortschak: Would you use the idiom used in legalTypesSym please.

kortschak:
LGTM after comments addressed and please wait for @btracey.

mat64/list_test.go
vladimir-ch: Done

mat64/vector.go
```diff
@@ -137,6 +137,31 @@ func (v *Vector) ScaleVec(alpha float64, a *Vector) {
...
+
+	v.reuseAs(ar)
+
+	switch {
+	case v == a && v == b: // v <- v + alpha * v = (alpha + 1) * v
+		blas64.Scal(ar, alpha+1, v.mat)
+	case v == a && v != b: // v <- v + alpha * b
+		blas64.Axpy(ar, alpha, b.mat, v.mat)
+	case v != a && v == b: // v <- a + alpha * v
+		blas64.Scal(ar, alpha, v.mat)
```
btracey: I would guess based on the current coding of packages, that the two blas64 calls is notably slower than writing the one blas call. For large vectors, cache misses are everything (go look at our benchmarks for Daxpy vs OpenBLAS, even with all of the assembly tricks blas/native is the same performance). Double implementing everything seems bad as well. 

I would propose that we add the fast-path if both vectors have inc1.
````
case v != a && v == b: // v <- a + alpha * v
if v.mat.Inc == 1 && a.mat.Inc == 1{ 
       for i, f := range a {
              v[i] *= alpha
              v[i] += f
       }
     return
}
blas64.Scal(ar, alpha, v.mat)
blas64.Axpy(ar, 1, a.mat, v.mat)
````
This at least short circuits the most common case.

kortschak: Use internal/asm.Daxpy and DaxpyInc. They both take two vector inputs.

btracey: Yes.

kortschak: Sorry, I'm wrong. DaxpyUnitary takes the destination as well, but also requires that it have a unit increment, and DaxpyInc only takes the sources. So, unless we add a DaxpyDestInc to asm, we can't use them.

vladimir-ch:
Added two fast paths, one that uses Go loop and one that calls `asm.DaxpyUnitary`.

PTAL

btracey:
LGTM

PR #274, opened 2015-10-17
mat64: add stack trace recovery to Maybe

kortschak:
@btracey Please take a look.


btracey:
LGTM

Is there any worry about "1 << 20" vs other numbers?

kortschak:
1MB is pretty much guaranteed to get a full trace for a single goroutine I reckon. We could make it smaller. It's a pity there is no way to query how much the runtime needs to get a full trace.

Issue #275, opened 2015-10-17
mat64: consider moving Error into matrix


kortschak:
When a cmat128 exists, it will probably want to have pretty much all the same error handling as mat64. To avoid the duplication it's probably worth moving the Error type into matrix.

kortschak:
This would include `mat64.Condition` and the proposed `mat64.ErrorStack`.

kortschak:
Also, the conversion shims suggested in https://github.com/gonum/matrix/issues/134#issuecomment-143550106.

Issue #134, opened 2015-05-23
matrix: propose add cmat128 package


kortschak:
At this stage it would be a stripped down complex128 version of the mat64 interface definitions (probably just the basic arithmetic operations - maybe with their implementations via cgo in the first instance) and a Dense representation. With a view to add those operations in the future and significantly to allow us to have an available representation for results from some of the lapack routines that generate complex results on some inputs.

btracey:
I think we are in a stable enough state that this is a good idea. I'm pretty happy with our general decisions (the receiver is modified, if something is modified, it is a receiver, if not it's a function) and type interactions. The remaining concerns I have are with the higher-level functions (Solve and Eigen, for example), which will be aided by cmat128. 

As a side note, I spent a bit of time looking into go implementations of Eigen for symmetric matrices (Dsyevr, I believe) in an attempt to help the redesign along, but there are a lot of steps to get there.

kortschak:
That was the provocation. I may put together something that touches this over the weekend. It would be cmat128.Matrix, and the basic arithmetic operations (mainly due to lack of testing more than anything else).

In a semi related direction, I have been thinking of retiring some of our interfaces - they have served well as a design document, but we don't use many of them and they are probably no longer necessary. Though I am not firmly wedded to this line of thinking.

btracey:
Do we have any sense how much could be autogenerated? It would be great if
cmat64 could mostly appear from mat64 (with some help I'm site). My minimal
experience trying to use the ast package suggests at the very least I would
need examples to copy in order to help that effort.

I agree with retiring the interfaces. In my opinion, if there's a specific
interface that we know is useful, it's good to have so there's a
"canonical" location. Otherwise, we'll just make sure our signatures are
interfacable  (as we do now). For example,  stat/dist may add some
interfaces even though it doesn't use then internally, as they would be
used by a sampling package for instance

kortschak:
My plan was to split files in mat64 into *_common.go and *_real.go where methods and functions are not already in their own files, then generate from the *_common.go and already generic files.

I'll file an issue for an interface retirement plan - I guess it would be a 404(k) - sorry.

kortschak:
I've just had a bit of a play with this and I think the ideas @jonlawlor put forward in https://github.com/gonum/matrix/issues/79#issuecomment-69828645 make sense (I was using CmplxAt), but we run into cyclic dependency issues.

[...]

PR #276, opened 2015-10-18
matrix/{cmat128,conv}: new packages

kortschak:
@btracey Please take a look.

### DIFF
--- a/cmat128/matrix.go
+++ b/cmat128/matrix.go
@@ -1,0 +1,68 @@
+// Copyright ©2013 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package cmat128
+
+// Matrix is the basic matrix interface type.
+type Matrix interface {
+	// Dims returns the dimensions of a Matrix.
+	Dims() (r, c int)
+
+	// At returns the value of a matrix element at (r, c). It will panic if r or c are
+	// out of bounds for the matrix.
+	At(r, c int) complex128
+
+	// T returns the transpose of the Matrix. Whether T returns a copy of the
+	// underlying data is implementation dependent.
+	// This method may be implemented using the Transpose type, which
+	// provides an implicit matrix transpose.
+	T() Matrix
+}
+
+var (
+	_ Matrix       = Transpose{}
+	_ Untransposer = Transpose{}
+)
+
+// Transpose is a type for performing an implicit matrix transpose. It implements
+// the Matrix interface, returning values from the transpose of the matrix within.
+type Transpose struct {
+	Matrix Matrix
+}
+
+// At returns the value of the element at row i and column j of the transposed
+// matrix, that is, row j and column i of the Matrix field.
+func (t Transpose) At(i, j int) complex128 {
+	return t.Matrix.At(j, i)
+}
+
+// Dims returns the dimensions of the transposed matrix. The number of rows returned
+// is the number of columns in the Matrix field, and the number of columns is
+// the number of rows in the Matrix field.
+func (t Transpose) Dims() (r, c int) {
+	c, r = t.Matrix.Dims()
+	return r, c
+}
+
+// T performs an implicit transpose by returning the Matrix field.
+func (t Transpose) T() Matrix {
+	return t.Matrix
+}
+
+// Untranspose returns the Matrix field.
+func (t Transpose) Untranspose() Matrix {
+	return t.Matrix
+}
+
+// Untransposer is a type that can undo an implicit transpose.
+type Untransposer interface {
+	// Note: This interface is needed to unify all of the Transpose types. In
+	// the cmat128 methods, we need to test if the Matrix has been implicitly
+	// transposed. If this is checked by testing for the specific Transpose type
+	// then the behavior will be different if the user uses T() or TTri() for a
+	// triangular matrix.
+
+	// Untranspose returns the underlying Matrix stored for the implicit transpose.
+	Untranspose() Matrix
+}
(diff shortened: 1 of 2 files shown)


kortschak:
When it comes time to work on cmat128 in earnest, it will be generated as described in the proposal issue #134. Currently it is just here to provide the API hook.

cmat128/matrix.go
```diff
@@ -0,0 +1,68 @@
+// Copyright ©2013 The gonum Authors. All rights reserved.
```
btracey: 2015?

cmat128/matrix.go
```diff
@@ -0,0 +1,68 @@
...
+
+// Untranspose returns the Matrix field.
+func (t Transpose) Untranspose() Matrix {
+	return t.Matrix
+}
+
+// Untransposer is a type that can undo an implicit transpose.
+type Untransposer interface {
+	// Note: This interface is needed to unify all of the Transpose types. In
+	// the mat64 methods, we need to test if the Matrix has been implicitly
```
btracey: Mat64 -> mat128?

kortschak: Thanks.

cmat128/matrix.go
kortschak: Copied directly from the mat64 code, so it's 2013. This also reduces diff noise when it's autogenerated.

btracey:
I'm not convinced about the design of the types in matrix. The goal is to be able to communicate between the two packages if necessary. The goals seem to be 

- Exactly 1 copy
- Minimal number of fast paths

Am I missing an important goal?

 I would like it that the fast paths exist in some sort of copy/load function. From what I understand, the lapack functions that return complex numbers return them as 2x2 matrices, and not matrices of reals and imaginary. Even with the layout, the current design doesn't seem to encourage efficiency. It seems much simpler to me (in a global sense) to only have "translation" types. These do not implement Matrix, and so only exist to move data. This way we aren't worried about users trying to explicitly use matrix package types as input, and we can think about each package in isolation.

kortschak:
The design goal is simplicity.

With the types here I can say something like `cm.Copy(matrix.NewComplex(m, nil))` and make use of the future cmat128.Dense.Copy method. Other approaches to interchange require that we duplicate code here.

Sure, it means that people might use these types directly in mat64 and cmat128 calls, but the worst it will do is cost them performance.

kortschak:
*ping*

btracey:
Having matrix import mat64 and cmat128 like this means that errors can't be moved into matrix (circular imports). If there's going to be a package imported by 64 and 128, and a second package that imports both of them, I think "matrix" should be the fundamental one imported by mat64 and mat128

btracey:
All of the cmat128 code LGTM.

btracey:
OK. I see the point. Maybe we could have gonum/matrix/translate that has these types, and Error can live in matrix? Something like
````
m.Copy(translate.Complex{R: a}) 
````
Translate is nice as it can also incorporate the other weird translation types, i.e. when complex numbers aren't stored two matrices but as blocks. (translate.BlockedComplex for example)

I don't think the New types are needed. matrix.Complex{R: r, I: i} seems fine. The Complex struct isn't doing anything to protect the data, as the data is returned uncopied from Real and Imag. Seems simpler and makes it clear the data isn't behind a wall. 

I don't understand what the Realer and Imager do. Who would implement them besides Complex?

kortschak:
> Having matrix import mat64 and cmat128 like this means that errors can't be moved into matrix (circular imports).

The intention is to never import this from mat64 or cmat128.

kortschak:
>Translate is nice as it can also incorporate the other weird translation types, i.e. when complex numbers aren't stored two matrices but as blocks. (translate.BlockedComplex for example)

Possibly trans, but it seems to be a proliferation of packages.

>I don't think the New types are needed. matrix.Complex{R: r, I: i} seems fine. The Complex struct isn't doing anything to protect the data, as the data is returned uncopied from Real and Imag. Seems simpler and makes it clear the data isn't behind a wall.

NewComplex prevents mismatched sizes from being used early with an informative panic.

>I don't understand what the Realer and Imager do. Who would implement them besides Complex? 

This prevents an import cycle.

btracey:
Your "conv" idea is way better than translate.

How does the Error type work if it's in matrix? matrix needs to import mat64 to support Real (for example) but mat64 needs to import matrix in order to return a Condition error

kortschak:
>Your "conv" idea is way better than translate.

I'd prefer it be here, purely to avoid an extra name space.

>How does the Error type work if it's in matrix? matrix needs to import mat64 to support Real (for example) but mat64 needs to import matrix in order to return a Condition error

I don't understand why `mat64` needs to import matrix for Real. Why does `mat64` need `Real`? The conversion routines here are for clients. If we have a function that returns a complex matrix in `mat64` (it was my understanding that we were going to return a `[]complex128` in the one case where this is an issue, but...) we can define a `Complex` type that satisfies `cmat128.Matrix` and return that. Similarly if we need to return a real matrix in `cmat128` we can define a type for it (I can see no reason for this at all though).

Is there something I'm missing?

btracey:
It's not for Real. #275  suggests moving the Error type and common errors into this package. mat64 cannot use the error types once moved if matrix also needs to import mat64 to support Real (for instance).

kortschak:
Ohh, yes. Sorry. Foolish. Should I make this `trans` or `conv`?

btracey:
conv I think.

kortschak:
PTAL

btracey:
Sorry, I still don't understand the Realer and Imager interfaces. cmat128.Matrix and mat64.Matrix are pretty fundamentally different. In order to convert between them (say, cmat -> mat), one can just use the Real type. Are there any examples of a matrix type that would implement these interfaces that is not the Complex type? Seems like an unnecessary complexity.

LGTM otherwise

kortschak:
Say we have a mat64 function that returns a complex matrix - this is not inconceivable. Then we can't return a cmat128.Matrix by name, so we return a mat64 interface equivalent that satisfies those interfaces.

btracey:
Okay. LGTM.

PR #280, opened 2015-10-22
mat64: rename [ab]Mat => [ab]U where it is an untranspose

kortschak:
My editor (depending on its search mode) conflates aMat with amat, and a single character's case is not a good differentiator, so change Mat to U.

@btracey Please take a look.


btracey:
LGTM

PR #281, opened 2015-10-23
all: move errors into matrix

### LINKED ISSUE #275: mat64: consider moving Error into matrix

kortschak:
Closes #275.

@btracey Please take a look.

### DIFF
--- a/mat64/matrix.go
+++ b/mat64/matrix.go
@@ -888,15 +888,6 @@ func Trace(a Matrix) float64 {
 	}
 }
 
-const (
-	// condNorm describes the matrix norm to use for computing the condition number.
-	condNorm = lapack.MaxRowSum
-
-	// condNormTrans is the norm to compute on A^T to get the same result as computing
-	// condNorm on A.
-	condNormTrans = lapack.MaxColumnSum
-)
-
 func min(a, b int) int {
 	if a < b {
 		return a
--- a/mat64/matrix.go
+++ b/mat64/matrix.go
@@ -888,15 +888,13 @@ func Trace(a Matrix) float64 {
 	}
 }
 
-// condNorm describes the matrix norm to use for computing the condition number.
-var condNorm = lapack.MaxRowSum
+const (
+	// condNorm describes the matrix norm to use for computing the condition number.
+	condNorm = lapack.MaxRowSum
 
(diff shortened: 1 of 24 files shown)


kortschak:
matrix needs a package doc. I'm not sure what this should be. Probably a general documentation of the (growing) suite. Some of the comment of mat64 may want to be either copied or moved to matrix.

btracey:
Just a crazy idea, but one I thought worth mentioning before we fix the API. Do we want to change mat64 and cmat128 to just fmat and cmat? It's easier to read and type, and more symmetric.

mat64/matrix.go
```diff
@@ -889,115 +888,13 @@ func Trace(a Matrix) float64 {
...
-var condNorm = lapack.MaxRowSum
-
-// condNormTrans is the norm to compute on A^T to get the same result as computing
-// condNorm on A.
-var condNormTrans = lapack.MaxColumnSum
-
-const stackTraceBufferSize = 1 << 20
+const (
+	// condNorm describes the matrix norm to use for computing the condition number.
+	condNorm = lapack.MaxRowSum
```
btracey: This should be exported in Matrix as well, right? This way someone can write
````
// Will Solve Error?
cond := mat64.Cond(a, matrix.CondNorm)
if cond > matrix.ConditionTolerance {
   // preemptively fix things
}
````

btracey:
LGTM except condNorm.

I would guess the documentation should largely be copied, probably after being separated into mat64 specific and not. Mat64 will likely be the first exposure of many to gonum, so it should keep robust documentation in my opinion, even if that causes some copying.

kortschak:
That was what I was thinking for the documentation. It would be good to be able to structure the mat64 docs so that common documentation could be mechanically updated when changes are made. I'll look into that.

I don't like the idea of dropping the precision from the package names.

mat64/matrix.go
kortschak: Done.

kortschak:
Conceivably, we could make ConditionTolerance a var, but I don't think it is a good idea, too much opportunity for client cleverness.

PR #284, opened 2015-10-23
cmat128: add conjugate transpose

### LINKED ISSUE #283: cmat128: Matrix should have a conjugate transpose
`H` or `Conj`?

kortschak:
This is complicated by the combinatorics, but I don't see a reasonable way around that, nor is there a way around that that can be made use of by blas or lapack even if we can, nor does a sensible program ever say `m.T().H()` or the other way around.

I think this may need further thought, so this is really to get the discussion started.

Fixes #283.

@btracey Please take a look.

### DIFF
--- a/cmat128/matrix.go
+++ b/cmat128/matrix.go
@@ -18,11 +18,19 @@ type Matrix interface {
 	// This method may be implemented using the Transpose type, which
 	// provides an implicit matrix transpose.
 	T() Matrix
+
+	// H returns the conjugate transpose of the Matrix. Whether H returns a copy of the
+	// underlying data is implementation dependent.
+	// This method may be implemented using the Conjugate type, which
+	// provides an implicit matrix conjugate transpose.
+	H() Matrix
 }
 
 var (
 	_ Matrix       = Transpose{}
+	_ Matrix       = Conjugate{}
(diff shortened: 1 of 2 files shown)


kortschak:
Currently the conv code is incorrect; the Imag method of Complex returns the matrix with the sign reversed if the Complex is conjugated. It seems that this will require two extra types - one for negation of the mat64.Matrix returned by a negated imaginary part (and if we want to address the issue raised above one for for the conjugate untransposed result of .T().H() - I don't think we should do this). I don't like this at all.

btracey:
Are there important examples of needing the nonconjugate transpose? My thought at the moment is to only support implicit H. We can provide a CopyTranspose function for the less common case, and we can avoid the ugly combinatorics. It's worth going through Lapack to see if there is something important, but the first two pages of results on "nonconjugate transpose" are mostly matlab related (the ' operator in matlab does conjugate by default).

kortschak:
SGTM

kortschak:
Would you look into that? Note that even without a T, we still need the negation type.

The win in not supporting T is that for the element-wise operations we don't double up the options (though the doubling is just setting a sign constant for the operation).

kortschak:
This has been dealt with in another PR.

Issue #282, opened 2015-10-23
matrix/{,cmat128}: add package documentation


kortschak:
See https://github.com/gonum/matrix/pull/281#issuecomment-150608626 and https://github.com/gonum/matrix/pull/281#issuecomment-150679516.

The idea of doing text munging does not appeal (we already do too much of this elsewhere and I don't want to add to that). I propose we write a matrix/gendoc.go (with `//+build ignore`) that holds the canonical package documentation for each of matrix, mat64 and cmat128 as a text template and have it emit doc.go for each of the three packages. Along with this we should add a go generate test for travis the same is exists in gonum/blas.

To start this we need to sort out what are common sections and what are package specific.

\cc @btracey

btracey:
SGTM. I would recommend we flesh out cmat a little more before we answer this (for example, deciding on #284 )

kortschak:
This is not really massively impacted by #284.

btracey:
No, but how much is package specific is. Agreed that we should start thinking about it.

PR #285, opened 2015-10-24
all: generate docs

kortschak:
@btracey Please take a look.


kortschak:
*ping*

cmat128/doc.go
```diff
@@ -0,0 +1,98 @@
+// Generated by running
+//  go generate github.com/gonum/matrix
+// DO NOT EDIT.
+//
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+//
+// Package cmat128 provides implementations of complex128 matrix structures and
```
btracey: ...of matrices with complex128 elements and linear...

cmat128/doc.go
```diff
@@ -0,0 +1,98 @@
...
+// sections provide more in depth commentary.
+//
+// cmat128 provides:
+//  - Interfaces for a complex Matrix
+//
+// BLAS and LAPACK
+//
+// BLAS and LAPACK are the standard APIs for linear algebra routines. Many
+// operations in cmat128 are implemented using calls to the wrapper functions
+// in gonum/blas/blas64 and gonum/lapack/lapack64. By default, blas64 and
```
btracey: cblas128 and clapack128?

cmat128/doc.go
```diff
@@ -0,0 +1,98 @@
...
+//
+// There are many possible type combinations and special cases. No specific guarantees
+// are made about the performance of any method, and in particular, note that an
+// abstract matrix type may be copied into a concrete type of the corresponding
+// value. If there are specific special cases that are needed, please submit a
+// pull-request or file an issue.
+//
+// Invariants
+//
+// Matrix input arguments to functions are never modified. If an operation would
```
btracey: Do we need a "(unless they are also a receiver)"?

cmat128/doc.go
```diff
@@ -0,0 +1,98 @@
...
+// change Matrix data, that matrix will be the receiver of a function.
+//
+// For convenience, a matrix may be used as both a receiver and as an input, e.g.
+//  a.Pow(a, 6)
+//  v.SolveVec(a.T(), v)
+// though in many cases this will cause an allocation (see Element Aliasing).
+// An exception to this rule is Copy, which does not allow a.Copy(a.T()).
```
btracey: Maybe we should add in here something like:
"Note that a consequence of these invariants is that not all calls may share receiver and input, as many operations have a resulting shape that is not equal to either of the input arguments. In particular, a.Mul(a, b) will panic if b is not square (as the resulting matrix does not have the size of a), and a.Stack(a, b) will always panic (since the resulting shape is longer than a)"

cmat128/doc.go
```diff
@@ -0,0 +1,98 @@
...
+//
+// This prohibition is to help avoid subtle mistakes when the method needs to read
+// from and write to the same data region. There are ways to make mistakes using the
+// cmat128 API, and cmat128 functions will detect and complain about those.
+// There are many ways to make mistakes by excursion from the cmat128 API via
+// interaction with raw matrix values.
+//
+// If you need to read the rest of this section to understand the behavior of
+// your program, you are being clever. Don't be clever. If you must be clever,
+// blas64 and lapack64 may be used to call the behavior directly.
```
btracey: blas128 and lapack128

doc.go
```diff
@@ -0,0 +1,83 @@
...
+//
+// This section provides a quick overview of the matrix package. The following
+// sections provide more in depth commentary.
+//
+// matrix provides:
+//  - Error type definitions
+//  - Error recovery mechanisms
+//  - Common constants used by mat64 and cmat128
+//
+// BLAS and LAPACK
```
btracey: Do we really want this all here? matrix itself doesn't have these properties, though the errors relate to them. Maybe the matrix documentation should just say it provides error types and constants, and direct users to the documentation of mat64 and cmat128 for more explanation on how they can arise?

cmat128/doc.go
kortschak: How does this differ from he mat64 case which uses the same language s/complex128/float64/

btracey: It doesn't, sorry, I should have been more clear. Except for the 64 vs 128 comments, they apply to both.

kortschak: This is the language that we've used previously. Any particular reason for the change?

btracey: I think it's clearer, but the difference is minor. Feel free to ignore.

cmat128/doc.go
kortschak: I think this is more something for the wiki documentation when we write it.

btracey: Okay.

cmat128/doc.go
kortschak: Done - without parens.

cmat128/doc.go
kortschak: Done.

doc.go
kortschak: I'll see if I can come up with some words that reflect this.

btracey: Okay. 

The problem I want solved is that the current size of the documentation makes it seem like the package does much more than it does. This might be the first thing users see, so it does need to direct them to the right place, but otherwise should just say what it does.

btracey:
Looks good minus the commenting on matrix.

kortschak:
PTAL

I'm not completely happy with the `Errors` section for matrix, but I think it's a start.

doc.go
```diff
@@ -0,0 +1,89 @@
...
+//  - Error recovery mechanisms
+//  - Common constants used by mat64 and cmat128
+//
+// Errors
+//
+// The mat64 and cmat128 matrix packages share a common set of errors
+// provided by matrix via the matrix.Error type. Errors are either
+// returned in cases where the function may meaningfully fail such as
+// the mat64 solve methods, or may be the parameter for a panic call
+// when there is a mismatch between the provided parameter and the
```
btracey: mismatch between the inputs of the function or method and the expectations, for example...

btracey:
I'm still not convinced that the Invariants and Shadowing sections need to be present in the matrix documentation, but it's probably not harmful to have them.

LGTM with the minor changes.

doc.go
kortschak: That messes with my head. For me, inputs can be the theoretical inputs or the concrete values passed to the function, and the expectations are not defined with respect to whose.

This is the sentence that I don't like, but I don't think that fixes it. I'll think more.

btracey: Okay. I don't think the use of the word "parameter" as both the input to panic and input to function. Maybe just call "input" the thing provided to panic, i.e. "may be the input to a panic call when there is a mismatch between the provided parameter(s) and ..."

kortschak: I'm breaking it into smaller sentences (always a good idea). I'll send again as a follow-up commit in this PR and add it to the first commit when you agree to it.

kortschak:
PTAL

kortschak:
Fixes #282.

doc.go
```diff
@@ -0,0 +1,95 @@
...
+//  - Error recovery mechanisms
+//  - Common constants used by mat64 and cmat128
+//
+// Errors
+//
+// The mat64 and cmat128 matrix packages share a common set of errors
+// provided by matrix via the matrix.Error type.
+//
+// Errors are either returned directly or used as the parameter of a panic
+// depending on the class of error encountered.
```
btracey: Maybe it'll help to say when we panic.

"Errors are either returned directly or used as the parameter of a panic depending on the class of error encountered. Examples of each type are found in the mat64 Solve methods, which find x such that A *x = b.  

Panics:
A function will panic when the input parameters are inappropriate for the function or when correctness is easily verified before calling the function. In Solve, for example, the columns of each matrix must be equal. Solve will panic with ErrShape if this is violated, as column length is easily verified before the call to solve

Error Returns:
An error value is returned from the function or method when the operation can meaningfully fail. The Solve operation is cannot complete if A is singular. However, determining the singularity of A is most easily discovered during the Solve procedure itself, and so in this case an error is returned."

Not great, but maybe something along those lines?

kortschak: I'm happy with that without the headings.

kortschak: Uhp, no it makes some claims I'm not happy with. The issue for panics is really that we are mimicking static types at runtime for the most part. This is essentially the same as the use of panic by reflect. I'll think more, but this helps. I also want to include that the receiver can cause a panic if the wrong shape.

kortschak:
PTAL

doc.go
```diff
@@ -0,0 +1,107 @@
...
+//  - Common constants used by mat64 and cmat128
+//
+// Errors
+//
+// The mat64 and cmat128 matrix packages share a common set of errors
+// provided by matrix via the matrix.Error type.
+//
+// Errors are either returned directly or used as the parameter of a panic
+// depending on the class of error encountered. As is usually the case in
+// Go code, returned errors indicate that a call was not able to complete
```
btracey: Is it usually the case in Go code? Probably just delete the introductory clause, or change go to gonum.

doc.go
```diff
@@ -0,0 +1,107 @@
...
+// successfully while panics generally indicate a programmer error or an
+// unrecoverable error.
+//
+// Examples of each type are found in the mat64 Solve methods, which find
+// x such that A*x = b.
+//
+// An error value is returned from the function or method when the operation
+// can meaningfully fail. The Solve operation is cannot complete if A is
+// singular. However, determining the singularity of A is most easily
+// discovered during the Solve procedure itself and is a valid result from
```
btracey: I would probably delete "is a valid result from the operation".

doc.go
```diff
@@ -0,0 +1,107 @@
...
+// the operation, so in this case an error is returned.
+//
+// A function will panic when the input parameters are inappropriate for
+// the function. In Solve, for example, the number of rows of each input
+// matrix must be equal because of the rules of matrix multiplication.
+// Similarly, for solving A*x = b, a non-zero receiver must have the same
+// number of rows as A has columns and must have the same number of columns
+// as b. In all cases where a function will panic, conditions that would
```
btracey: I would add something like "Incorrect sizes are easily discovered and handled prior to the call to Solve, so Solve panics with ErrShape if there is a mismatch. More generally, inappropriate input conditions lead to a panic if they can be easily checked prior to a call."

btracey:
Made a couple of suggestions, but your call. 

LGTM

Thanks for the expanded documentation.

doc.go
kortschak: It is, but I can delete.

doc.go
kortschak: Why? I want to highlight that a panic happens when an unexpected thing happens and that a singular matrix is something that is a reasonable expectation.

btracey: I just don't like the word "valid". Errors happen, but that doesn't make the other result valid. The returned matrix isn't a "valid solution to A*x = b", it's just an unpreventable error without trying. Maybe sub valid with "plausible"?

doc.go
btracey: It just seems like a more controversial statement than it's worth.

doc.go
kortschak: I don't think this is the basis - it puts the cart before the horse. We are mimicking a type system (we are actually implementing a type system), just one that is determined at run time. We preclude certain sets of behaviours by prohibiting some user actions. This is the foremost goal. The corollary of that prohibition is that we should (and do) make it possible for the user to determine whether they would be breaking the rules (if, for example, they were given untrusted input). The wording here reflects that priority. The converse, e.g. trying to solve a system which has a singular input, is a perfectly legitimate thing to do in the rules of linear algebra, it just fails. This is the difference IMO.

btracey: I see. That's a good way of putting it.

doc.go
kortschak: This is where my biology background (and interaction with a theoretical physicist for more than a decade comes in). Here the validity is the system behaviour, not the solution. Note that the text does not say that the solution is valid, but that the result is. This may need clarification, but the result is the ensemble of changes to `m` and the returned `error` - together in the Go/gonum context the result of a call to `m.Solve(a, b)` where a represents a singular matrix is perfectly valid - the result need to be interpreted as "The A matrix is singular and this is how badly, so I cannot solve this for you." This is a valid result.

btracey: Okay, we can leave it.

doc.go
kortschak: Done.

PR #286, opened 2015-10-28
mat64: Vector clean up and some additions

vladimir-ch:
@btracey @kortschak PTAL

### DIFF
--- a/mat64/vector.go
+++ b/mat64/vector.go
@@ -210,7 +210,7 @@ func (v *Vector) AddScaledVec(a *Vector, alpha float64, b *Vector) {
 	}
 }
 
-// AddVec adds a and b element-wise, placing the result in the receiver.
+// AddVec adds the vectors a and b, placing the result in the receiver.
 func (v *Vector) AddVec(a, b *Vector) {
 	ar := a.Len()
 	br := b.Len()
--- a/mat64/vector.go
+++ b/mat64/vector.go
@@ -145,6 +145,15 @@ func (v *Vector) ScaleVec(alpha float64, a *Vector) {
 
 // AddScaledVec adds the vectors a and alpha*b, placing the result in the receiver.
 func (v *Vector) AddScaledVec(a *Vector, alpha float64, b *Vector) {
+	if alpha == 1 {
+		v.AddVec(a, b)
+		return
+	}
+	if alpha == -1 {
+		v.SubVec(a, b)
+		return
+	}
+
 	ar := a.Len()
 	br := b.Len()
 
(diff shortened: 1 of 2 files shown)


mat64/vector.go
```diff
@@ -71,11 +71,11 @@ func NewVector(n int, data []float64) *Vector {
 }
 
 // ViewVec returns a sub-vector view of the receiver starting at element i and
-// extending n columns. If i is out of range, or if n is zero or extend beyond the
-// bounds of the Vector ViewVec will panic with ErrIndexOutOfRange. The returned
-// Vector retains reference to the underlying vector.
+// extending n rows. If i is out of range, or if n is zero, or if the view
```
kortschak: "If i is out of range, n is zero, or the view"...

mat64/vector.go
```diff
@@ -71,11 +71,11 @@ func NewVector(n int, data []float64) *Vector {
...
+// extends beyond the bounds of the Vector, ViewVec will panic with ErrIndexOutOfRange.
+// The returned Vector retains reference to the underlying vector.
 func (v *Vector) ViewVec(i, n int) *Vector {
-	if i+n > v.n {
+	if i < 0 || n <= 0 || i+n > v.n {
 		panic(matrix.ErrIndexOutOfRange)
 	}
 	return &Vector{
```
kortschak: Can you add an upper bound to the Data slice. When we add the overlap detection to Vector operations it will detect things as overlapping if they are not properly truncated. (We'll probably detect other cases of this when I wire that in).

mat64/vector.go
```diff
@@ -116,16 +116,32 @@ func (v *Vector) Reset() {
...
 func (v *Vector) RawVector() blas64.Vector {
 	return v.mat
 }
 
 // CopyVec makes a copy of elements of a into the receiver. It is similar to the
-// built-in copy; it copies as much as the overlap between the two matrices and
-// returns the number of rows and columns it copied.
+// built-in copy; it copies as much as the overlap between the two vectors and
+// returns the number of elements it copied.
 func (v *Vector) CopyVec(a *Vector) (n int) {
```
kortschak: We can drop the n in the signature here.

kortschak:
LGTM with minor comments.

mat64/vector.go
vladimir-ch: I added a tight upper bound `(i+n-1)*v.mat.Inc+1` but now I realize that perhaps it may clash with some BLAS or other assumptions so I'd like to confirm that it is ok. If not, I'll change it to the simple `(i+n)*v.mat.Inc` and submit in a PR.

kortschak: The tight upper bound was what I was expecting. I'd rather have a catastrophic failure with this than have to figure out where an incorrect bound was coming from when I add overlap detection.

If any blas routine fails from this, I think it's probably wrong.

I guess it's probably worth adding tests for viewed vector operations.

vladimir-ch: Ok, that's what I was thinking.

> I guess it's probably worth adding tests for viewed vector operations.

I already started looking at how to best add them.

PR #289, opened 2015-11-04
mat64: remove unused interface checks and add Reseter check

kortschak:
The dead checks are not valid and are never going to be valid under the current design approach, so delete them. *Vector is however a Reseter, so check that.

@btracey Please take a look.


btracey:
LGTM

PR #294, opened 2015-11-06
mat64: special-case Dot for RawVectorers

vladimir-ch:
At the moment not for submission because oddly enough, this change makes fail one (and only one) test for the updated optimize.BFGS (that uses mat64 instead of floats). If I run the tests in optimize with `-tags noasm`, the test passes. I suspect that the assembler and Go versions of asm.Ddot have different order of the operations leading to different results and eventually to the failed test, something described for example in http://blog.nag.com/2011/02/wandering-precision.html. What would be a proper action? Loosen the tolerance for that test?

### DIFF
--- a/mat64/matrix.go
+++ b/mat64/matrix.go
@@ -368,9 +368,17 @@ func Dot(a, b Matrix) float64 {
 	if r != rb || c != cb {
 		panic(matrix.ErrShape)
 	}
-	var sum float64
 	aU, aTrans := untranspose(a)
 	bU, bTrans := untranspose(b)
+	if rma, ok := aU.(RawVectorer); ok {
+		if rmb, ok := bU.(RawVectorer); ok {
+			if c > r {
+				r = c
+			}
+			return blas64.Dot(r, rma.RawVector(), rmb.RawVector())
+		}
+	}
+	var sum float64
 	if rma, ok := aU.(RawMatrixer); ok {
 		if rmb, ok := bU.(RawMatrixer); ok {
 			ra := rma.RawMatrix()


btracey:
Assuming the toleranace doesn't need to be increased by more than an order of magnitude, seems okay to raise.

mat64/matrix.go
```diff
@@ -368,9 +368,19 @@ func Dot(a, b Matrix) float64 {
...
 	}
-	var sum float64
 	aU, aTrans := untranspose(a)
 	bU, bTrans := untranspose(b)
+	if rma, ok := aU.(RawVectorer); ok {
+		if rmb, ok := bU.(RawVectorer); ok {
+			if c > r {
+				r = c
+			}
+			ra := rma.RawVector()
```
btracey: Just move these into the blas64 function call.

btracey:
LGTM with the minor comment.

kortschak:
We are going to see similar issue with the addition of go1.6 to the test matrix. I just ran the test for my proposal for this with the tip version of go I was working on this morning to bring @sbinet's fortran cgo CL back to life and TestInvers and TestCond both fail - running the unmodified tip also fails with these.

mat64/matrix.go
vladimir-ch: Good point. Done.

vladimir-ch:
A tiny reorganization in one of BFGS formulas made the test pass. Crazy.

kortschak:
@vladimir-ch If you hold up on the merge until after travis has finished (or at least started) and the same for the deletion of the branch, then it won't fail like this.

PR #295, opened 2015-11-06
mat64: add fast path for *Vector pair to Dot

kortschak:
@btracey Please take a look.


kortschak:
Abandoned in favour of #294.

PR #297, opened 2015-11-09
Add SubsetSym to symmetric

btracey:


btracey:
PTAL @kortschak 

I'm not sure this is the right name, but I couldn't think of a better one. Documenting the function is a bit tricky even though the behavior is simple. 

This will be used in distmv for computing marginal and conditional distributions of Normal

btracey:
The fast path seems wrong. Not sure why it's passing. Fixing.

btracey:
Fixed. Tests were passing because I forgot to return early so the answer was being overwritten anyway.

kortschak:
I think this need a really clear example. Indexing always does my head in.

btracey:
Example added.

mat64/symmetric.go
```diff
@@ -326,3 +335,42 @@ func (s *SymDense) ScaleSym(f float64, a Symmetric) {
 		}
 	}
 }
+
+// SubsetSym extracts a subset of the rows and columns of the matrix a and stores
+// the result in-place into the receiver. The resulting matrix size is
+// len(set)×len(set). Specifically, at the conclusion of SubsetSym,
+// s.At(i, j) equals a.At(set[i], set[j]). Note that the supplied set does not
```
kortschak: Add "for all (i, j) of the receiver" or something similar after "a.At....".

mat64/symmetric_example_test.go
```diff
@@ -0,0 +1,49 @@
...
+	s := mat64.NewSymDense(5, nil)
+	count := 1.0
+	for i := 0; i < n; i++ {
+		for j := i; j < n; j++ {
+			s.SetSym(i, j, count)
+			count++
+		}
+	}
+	fmt.Println("Original matrix:")
+	fmt.Printf("%0.4v\n", mat64.Formatted(s))
```
kortschak: Wandered off into commit land again.

Extra newline at end of each fmt.Printf format (except last).

mat64/symmetric_example_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+	// Take the subset {0, 2, 4}
+	var sub mat64.SymDense
+	sub.SubsetSym(s, []int{0, 2, 4})
+
```
kortschak: Delete this line.

kortschak:
LGTM with minor comments.

mat64/symmetric_example_test.go
btracey: Done.

mat64/symmetric_example_test.go
btracey: Done.

PR #300, opened 2015-12-01
Add ViewSym

btracey:


mat64/symmetric_test.go
```diff
@@ -533,3 +533,50 @@ func TestSubsetSym(t *testing.T) {
...
+	} {
+		n := test.n
+		s := NewSymDense(n, nil)
+		for i := 0; i < n; i++ {
+			for j := i; j < n; j++ {
+				s.SetSym(i, j, float64((i+1)*n+j+1))
+			}
+		}
+
+		// Take a subset and check the view matches.
```
kortschak: Make a check that a write to the view reflects in the original or vice versa.

kortschak:
At some stage this will presumably gain a Grow as well? This is addable at a later stage, but do you want to do it now?

LGTM after additional test added.

mat64/symmetric_test.go
btracey: Oh, yea, thanks.

btracey:
I can do it now, sure. I'll do it in the next PR.

btracey:
Actually, I wonder if the signature should really be 
````
ViewSquare(i,n int) Matrix
````
as the same signature works for Triangular. We don't lose anything with the interface return, since Symmetric isn't a viewer anyway, and an interface assertion is needed to use advanced operations either way.

**Comment by btracey** [comment]
Oh, and I see I forgot how view worked. I'll add Grow here as they really are complementary.

**Comment by kortschak** [comment]
> `ViewSquare(i,n int) Matrix`

SGTM

**Comment by btracey** [comment]
This will be a bit, as a cap stride is needed to correctly support grow. Consider a view starting at {1,1}, and then growing by 1. The stride number looks like there is space, but there is not really. I don't have time to fix this now, but I'll try to get to it later tonight.

**Comment by kortschak** [comment]
Understood. It's not strictly necessary to go in now, but I'm happy to wait for this.

**Comment by btracey** [comment]
No, it should go in now, because it's the proper way to make View work. 

Is it okay to call the field "cap" or should it be avoided because of the built-in?

**Comment by kortschak** [comment]
cap is fine.

**Comment by btracey** [comment]
That wasn't as bad as I thought it would be. PTAL.

Triangular capacity added for consistency. Grow and View can be added when needed.

File: mat64/dense.go [hunk 2]
```diff
@@ -238,7 +239,7 @@ func (m *Dense) rowView(r int) []float64 {
 
 // View returns a new Matrix that shares backing data with the receiver.
 // The new matrix is located from row i, column j extending r rows and c
-// columns.
+// columns. View panicks if the view is outside the bounds of the receiver.
```
kortschak: panics

mat64/dense.go
```diff
@@ -253,9 +254,10 @@ func (m *Dense) View(i, j, r, c int) Matrix {
 	return &t
 }
 
-// Grow returns an expanded copy of the receiver. The copy is expanded
-// by r rows and c columns. If the dimensions of the new copy are outside
-// the caps of the receiver a new allocation is made, otherwise not.
+// Grow returns the receiver expanded by r rows and c columns. If the dimensions
+// of the expanded matrix are outside the capacities of the receiver a new
+// allocation is made, otherwise not. Note the receiver itself is not modified
+// during the call to Grow.
```
kortschak: At some point when we start writing wiki-type docs this probably needs expanding. The whole "slice is a reference type" thing will rear its head here too - people wondering why the receiver sometimes reflects changes in the returned matrix and sometimes not.

Probably what we need to do is write a gonum-tour set of articles (I think I might have mentioned this previously).

mat64/dense.go
```diff
@@ -291,10 +293,18 @@ func (m *Dense) Grow(r, c int) Matrix {
 		t.capRows = cr
 		t.capCols = cc
 		// Copy the complete matrix over to the new matrix.
-		// Including elements not currently visible.
-		r, c, m.mat.Rows, m.mat.Cols = m.mat.Rows, m.mat.Cols, m.capRows, m.capCols
-		t.Copy(m)
-		m.mat.Rows, m.mat.Cols = r, c
+		// Including elements not currently visible. Use a temporary structure
```
kortschak: Do we know if tmp escapes? The assumption that it did is probably why I wrote it the way I did.

mat64/pool.go
```diff
@@ -112,6 +112,7 @@ func getWorkspaceSym(n int, clear bool) *SymDense {
 	}
 	s.mat.N = n
 	s.mat.Stride = n
+	s.cap = n
```
kortschak: We need to make sure that we never call Grow on a workspace.

mat64/symmetric.go
```diff
@@ -136,16 +147,7 @@ func (s *SymDense) AddSym(a, b Symmetric) {
...
-		s.mat = blas64.Symmetric{
-			N:      n,
-			Stride: n,
-			Data:   use(s.mat.Data, n*n),
-			Uplo:   blas.Upper,
-		}
-	} else if s.mat.N != n {
-		panic(matrix.ErrShape)
-	}
+	s.reuseAs(n)
```
kortschak: Odd that we weren't using this before.

mat64/dense.go
btracey: Done.

mat64/dense.go
btracey: I definitely agree on the gonum-tour article set. It would also be a lot more fun to do than lapack translations and reviews :)

Yea, these are another of those weird cases. It's also odd that they don't behave like slices, i.e. View is only shorter. It's fine behavior for a matrix package, but it's something worth noting.

mat64/dense.go
btracey: Oh, I don't know. I think this way is better even if it allocates. With these changes, Grow can be called concurrently with other operations. We should support that as much as reasonable.

mat64/symmetric.go
btracey: This code predates reuseAs for Dense, and especially for SymDense.

mat64/symmetric.go
```diff
@@ -374,3 +359,62 @@ func (s *SymDense) SubsetSym(a Symmetric, set []int) {
 		}
 	}
 }
+
+// ViewSquare returns a view of the submatrix starting at {i, i} and extending
+// for n rows and columns. ViewSquare panicks if the view is outside the bounds
```
btracey: Fixed spelling.

mat64/symmetric.go
```diff
@@ -112,8 +119,12 @@ func (s *SymDense) reuseAs(n int) {
 			Data:   use(s.mat.Data, n*n),
 			Uplo:   blas.Upper,
 		}
+		s.cap = n
 		return
 	}
+	if s.mat.N > s.cap {
```
kortschak: Move this above the assignment to *s, as written this will never fail. The intention is to catch a case where something previously incorrectly set cap (or failed to set it at all more likely), rather than that the current reuseAs called failed.

kortschak:
LGTM after typo and reuseAs fixed.

btracey:
PTAL

mat64/dense.go
kortschak: > With these changes, Grow can be called concurrently with other operations. We should support that as much as reasonable.

Agreed, that trumps allocation cost.

mat64/symmetric.go
btracey: I moved it, but I don't understand why this would never fail. If s.mat.N is non-zero and s.cap is zero this would panic, right?

kortschak: You are right. The set is small, but non-zero. I feel safer though that we catch any case where this has happened.

kortschak:
LGTM

kortschak:
Can you fix the commit message?

btracey:
Okay... apparently when they let you edit the commit message when you press the "merge" button, that only edits the merge commit text, and not the commit message itself. Looking how to fix.

btracey:
Nope... force pushing a new commit to master is not allowed. Is there another way? Sorry...

kortschak:
Not fixable - by design.

kortschak:
I should have looked before I wrote LGTM.

btracey:
No, it's my fault. I assumed that editing the commit message in the box was the same as editing the commit message on git. I should have checked.

PR #305, opened 2016-01-15
mat64: use new DscalInc signature in Vector.ScaleVec()

vladimir-ch:
Depends on gonum/internal#23. PTAL @kortschak


kortschak:
LGTM

PR #310, opened 2016-01-25
Add Dim method to cholesky

btracey:


mat64/cholesky.go
```diff
@@ -74,6 +74,11 @@ func (c *Cholesky) Det() float64 {
 	return math.Exp(c.LogDet())
 }
 
+// Dim returns the dimension of the factorized matrix.
```
kortschak: Move this above Det.

kortschak:
Is Dim the best name? I can't think of a better one, but it's very close to Dims which is one of our interface methods.

kortschak:
LGTM

btracey:
I'm not sure I like Dim either. Other options:
- Size
- N
- Cholesky (in the spirit of Symmetric and Triangular)

kortschak:
Cholesky as a method name is misleading. Size works for me though I was also considering N.

btracey:
I'll change to Size. I also thought about Dims, but I thought Cholesky even starting to look like it implements Matrix is a bad idea.

PR #315, opened 2016-01-31
Some cleanup in mat64 docs

vladimir-ch:


mat64/matrix.go
```diff
@@ -409,7 +409,7 @@ func Dot(a, b Matrix) float64 {
 	return sum
 }
 
-// Equal returns whether element a and b have the same size and contain all equal
+// Equal returns whether the matrices a and b have the same size and contain all equal
```
kortschak: The wording of this is weird. An interpretation could be that all elements of a and b must be the same.

mat64/solve.go
```diff
@@ -11,15 +11,15 @@ import (
 	"github.com/gonum/matrix"
 )
 
-// Solve solves a minimum-norm solution to a system of linear equations defined
-// by the matrices a and b. If a is singular or near-singular a Condition error
+// Solve finds a minimum-norm solution to a system of linear equations defined
+// by the matrices A and b. If A is singular or near-singular a Condition error
```
kortschak: This was intentional. Though would could change the decision. `a` is the name of the variable in the signature while A is the convention for the matrix notation. These are in conflict.

vladimir-ch: I know that we distinguish between the two but it is true that I'm not sure which one we choose when either is possible (as in this case). I went with the common linear algebra notation because SolveLU et al use it (true, they don't have the matrix explicitly among the arguments) but also because `Dense.Solve` has `... a and b. ...` while `Vector.Solve` has `... A and B. ...`.

Do you mean that it should be: small a and b in this paragraph and in the comment for `Vector.Solve`, and capital A and B in the next paragraph?

mat64/matrix.go
vladimir-ch: What would be a better wording?

```
Equal returns whether the matrices a and be have the same size
and are element-wise equal.
```
? Also maybe I prefer to replace `size` with `shape`.

kortschak: That wording is good. I'm not sure about size v shape.

mat64/solve.go
kortschak: a when you're writing about the variable label and A when you are writing about the mathematical entity.

mat64/matrix.go
vladimir-ch: Done, kept `size`.

vladimir-ch:
Removed some more comment lint. PTAL

kortschak:
LGTM but wait for @btracey.

mat64/index_bound_checks.go
```diff
@@ -40,7 +40,8 @@ func (m *Dense) set(r, c int, v float64) {
 	m.mat.Data[r*m.mat.Stride+c] = v
 }
 
-// At returns the element at row r, column c. It panics if c is not zero.
+// At returns the element at row r, column c.
```
btracey: While we're here, do we want to make this At(i,j int)? Frequently r and c and the number of rows and columns, and i and j are a specific index.

mat64/solve.go
```diff
@@ -11,16 +11,16 @@ import (
...
+// by the matrices a and b. If A is singular or near-singular, a Condition error
 // is returned. Please see the documentation for Condition for more information.
 //
 // The minimization problem solved depends on the input parameters.
-//  1. If m >= n and trans == false, find X such that ||a*X - b||_2 is minimized.
-//  2. If m < n and trans == false, find the minimum norm solution of a * X = b.
-//  3. If m >= n and trans == true, find the minimum norm solution of a^T * X = b.
-//  4. If m < n and trans == true, find X such that ||a*X - b||_2 is minimized.
-// The solution matrix, X, is stored in place into the receiver.
+//  1. If m >= n and trans == false, find X such that ||A*X - B||_2 is minimized.
```
btracey: There is no trans input, so we can delete conditions 3 and 4.

mat64/index_no_bound_checks.go
```diff
@@ -40,7 +40,8 @@ func (m *Dense) set(r, c int, v float64) {
 	m.mat.Data[r*m.mat.Stride+c] = v
 }
 
-// At returns the element at row r, column c. It panics if c is not zero.
```
btracey: Same here.

mat64/index_bound_checks.go
btracey: Same would go with Set.

vladimir-ch: Done.

mat64/solve.go
vladimir-ch: 3 and 4 deleted, 1 and 2 adapted.

mat64/index_no_bound_checks.go
vladimir-ch: Done.

vladimir-ch:
PTAL

btracey:
Could you change the Set routines to use i and j as well? 

LGTM after that.

vladimir-ch:
Changed everywhere in `index_*bound_checks.go`.

PR #318, opened 2016-02-01
mat64: fix documentation generation errors

### LINKED ISSUE #317: mat64 documentation refers to cblas128
Looks like an autogeneration error. Sorry for missing this @kortschak 

Current text
````
BLAS and LAPACK are the standard APIs for linear algebra routines. Many operations in mat64 are implemented using calls to the wrapper functions in gonum/blas/cblas128 and gonum/lapack/clapack128. By default, cblas128 and clapack128 call the native Go implementations of the routines. Alternatively, it is possible to use C-based implementations of the APIs through the respective cgo packages and "Use" functions. The Go implementation of LAPACK makes calls through cblas128, so if a cgo BLAS implementation is registered, the clapack128 calls will be partially executed in Go and partially executed in C. 
````

kortschak:
Fixes #317.

@btracey Please take a look.

### DIFF
--- a/gendoc.go
+++ b/gendoc.go
@@ -258,8 +258,8 @@ package {{.Name}}
 //  var c mat64.Dense // construct a new zero-sized matrix
 //  c.Mul(a, a)       // c is automatically adjusted to be 6×6`,
 
-		BLAS:   []string{"cblas128"},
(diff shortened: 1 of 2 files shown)


btracey:
LGTM

Issue #313, opened 2016-01-29
eigendecomposition is not usable?


idavydov:
6015d269f11686cecdb1622a6ac1060ca8201862 makes it impossible to use matrix eigendecomposition outside the library.
Change from Eigen to eigen makes the function non-exported.

btracey:
Could you explain what you mean?

````
var eigen mat64.Eigen
ok := eigen.Factorize(a)
if !ok {
   log.Fatal("bad decomp")
}
evs := eigen.Values(nil, false)
fmt.Println("eigenvalues = ", evs)
````

idavydov:
@btracey is it possible to get eigenvectors as well?

btracey:
Ah, no, not right at the moment. Have you been using that extraction? The plan is to put it in, but I don't know the resolution to https://github.com/gonum/matrix/issues/308 . If you need eigenvectors, we can temporarily make EigenFactors still public until that issue is resolved.

idavydov:
@btracey yes I've been using eigenvalues and eigenvectors in my code: https://bitbucket.org/Davydov/godon/src/c45db765c6f7b364454665151241cadc7572a259/cmodel/ematrix.go?at=master&fileviewer=file-view-default#ematrix.go-46
Having a way to access them will be very nice, thanks.

btracey:
PR submitted. Couple of side notes though

1) You are aware we have an Exp function in mat64 already? https://godoc.org/github.com/gonum/matrix/mat64#Dense.Exp
2) Do you know your matrix is symmetric? Are your sure your code works even if complex eigenvalues are generated?
3) It's almost always a bad idea to compute an explicit matrix inverse. The code
````
res := mat64.NewDense(cols, rows, nil)
res.Mul(m.v, cD)
res.Mul(res, m.iv)
`````

is better as 

````
res := mat64.NewDense(cols, rows, nil)
err := res.Solve(m.v.T(), cD.T())
if err != nil {
    log.Fatal("bad solve")
}
res.Mul(m.v, res)
````

idavydov:
@btracey thanks a lot.
In my problem I have to compute exp(t*Q) multiple times, where Q is a constant matrix and t is a variable scalar. So performing a single eigendecomposition followed by a number of dot products is much faster.
The matrix is non-symmetric. But since my Q matrix corresponds to the time-reversible continuous-time Markov proccess, there should be no complex eigenvalues. And the exponential should be strictly positive.
Thanks a lot.

[...]

PR #320, opened 2016-02-02
allow epsilon specification for the Factorize method, issue #313

idavydov:
Proposed a temporary fix.
It's good to have a control over epsilon, since it affects the eigen performance.

### DIFF
--- a/cmat128/matrix.go
+++ b/cmat128/matrix.go
@@ -9,9 +9,9 @@ type Matrix interface {
 	// Dims returns the dimensions of a Matrix.
 	Dims() (r, c int)
 
-	// At returns the value of a matrix element at (r, c). It will panic if r or c are
-	// out of bounds for the matrix.
-	At(r, c int) complex128
+	// At returns the value of a matrix element at row i, column j.
+	// It will panic if i or j are out of bounds for the matrix.
+	At(i, j int) complex128
 
 	// T returns the transpose of the Matrix. Whether T returns a copy of the
 	// underlying data is implementation dependent.
--- a/conv/conv.go
+++ b/conv/conv.go
@@ -49,15 +49,15 @@ func (m Complex) Dims() (r, c int) {
 	return m.r.Dims()
 }
 
-// At returns the element at row r, column c.
-func (m Complex) At(r, c int) complex128 {
+// At returns the element at row i, column j.
+func (m Complex) At(i, j int) complex128 {
 	if m.i == nil {
-		return complex(m.r.At(r, c), 0)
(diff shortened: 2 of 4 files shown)


kortschak:
I don't think we can have a method Factorize that takes an epsilon. We can provide a Factorize that hand over a reasonable epsilon (the lowest value that works for your failing case, multiplied by ~10 is probably reasonable). We could conceivable offer a temporary FactoriseEpsilon, thought I don't like that much.

idavydov:
Can we have something like this?

    type Eigen struct {
            vectorsComputed bool
    
            n int // The size of the factorized matrix.
    
            ef eigenFactors
            
            Epsilon float64
    }

    func NewEigen() *Eigen {
        return &Eigen{Epsilon: 1e-9}
    }

kortschak:
We can with similar qualms. The issue is that when the implementation is changed to be based on LAPACK, there is no such API. The exported Epsilon field would have to be marked as temporary. I'll leave it to @btracey to make this decision.

btracey:
Does the epsilon change at which your code hangs? I'd rather just set epsilon to 1e-8 or thereabouts.

btracey:
So you know @idavydov , part of the hesitancy to add the field is that I do not know when this code will get replaced with Dgeev. There are roughly 18 lapack functions that need to be implemented which is a lot of work.

idavydov:
@btracey the minimal epsilon value at which it works is 1e-310, 1e-311 hangs (I suppose it's zero in float64 representation?).
I think something like 1e-8 is a reasonable default choice. But I have to note that eigen with epsilon 1e-9 is 100 time faster compared to epsilon=1e-310. That's why I think it might be useful to be able to tweak this.
At least for now, while it's not LAPACK. 
For my code something like 1e-8 should be a reasonable value.

btracey:
Is 1e-9 100 times faster that 1e-16 (machine precision), or basically the same? 

We'd like API stability, and this function will lag that. When it eventually gets switched over, it would be nice to have similar (or faster) performance. The problem with an epsilon field is that not only will its removal break code, but it also allows users to run with a tolerance of 1e-1 or something to quickly get approximate eigenvalues. This is not possible to do with the Lapack implementation. In short, the tolerance field could break both compiling, but also break algorithms. If something like 1e-8 or 1e-16 performs reasonably, then it seems closer to what the eventual behavior will be, thus increasing stability.

idavydov:
@btracey 1e-16 is ~20% slower compared to 1e-9. I think that should be ok for the most applications.

btracey:
1e-16 sounds good to me.

PR #324, opened 2016-02-09
mat64: RawColView returns []float64

### LINKED ISSUE #316: RawColViewer should return []float64
RawColViewer currently returns a *Vector. It should be a []float64

vladimir-ch:
Closes #316


kortschak:
LGTM

PR #325, opened 2016-02-09
Rename Vectorer and VectorSetter interfaces

vladimir-ch:
`RowColer` and `RowColSetter` seem like the most natural options.


kortschak:
LGTM

PR #327, opened 2016-02-10
Update TestRow and TestCol

vladimir-ch:
Better reviewed on a per-commit basis.

### DIFF
--- a/mat64/matrix_test.go
+++ b/mat64/matrix_test.go
@@ -60,22 +60,54 @@ func eye(n int) *Dense {
 }
 
 func TestCol(t *testing.T) {
-	for i, af := range [][][]float64{
-		{{1, 2, 3}, {4, 5, 6}, {7, 8, 9}},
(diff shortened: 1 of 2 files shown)


mat64/matrix_test.go
```diff
@@ -58,15 +59,57 @@ func eye() *Dense {
...
+	}
+
+	denseComparison := func(a *Dense) interface{} {
+		r, c := a.Dims()
 		ans := make([][]float64, c)
 		for j := range ans {
-			ans[j] = Col(nil, j, a)
+			ans[j] = make([]float64, r)
+			for i := range ans[j] {
+				ans[j][i] = a.At(i, j)
```
vladimir-ch: It seemed to me that the comparison should not use the function that is being tested.

vladimir-ch:
PTAL

kortschak:
LGTM

PR #334, opened 2016-02-12
Fix docs in conv and cmat128.Conjugate.At

vladimir-ch:


cmat128/matrix.go
```diff
@@ -35,7 +37,7 @@ type Conjugate struct {
 // At returns the value of the element at row i and column j of the transposed
 // matrix, that is, row j and column i of the Matrix field.
 func (t Conjugate) At(i, j int) complex128 {
-	return t.Matrix.At(j, i) * complex(1, -1)
+	return cmplx.Conj(t.Matrix.At(j, i))
```
kortschak: I don't this this is worth the import, but it could be more sensibly written as 

```
v := t.Matrix.At(j, i)
return complex(real(v), -imag(v)) 
```

which is just what cmplx.Conj does.

kortschak:
LGTM after change to Conjugate.At body. Thank you.

cmat128/matrix.go
vladimir-ch: In fact this is what I originally had there but then I noticed that cmplx.Conj exists without checking what it does internally. Done.

kortschak: Presumably at a later date we will need to import cmplx, we can use that then.

Issue #337, opened 2016-02-25
Use in conjunction with gob


pcostanza:
I'm trying to send instances of `mat64.Vector` over a TCP/IP connection, and use `encoding/gob` from the Go standard library for this purpose. While it's possible to make it work, I believe it's currently not ideal.

Encoding (sending) is relatively straightforward:

    func EncodeVector(enc *gob.Encoder, v *mat64.Vector) {
        enc.Encode(v.RawVector().Data)
    }

Encoding the vector instance directly, like in `enc.Encode(v)` is not possible, because the instance variables of the `mat64.Vector` struct are not exported, and therefore the gob library does not (cannot?) access them. Encoding the raw `[]float64` slice instead is fine, though.

However, decoding is wasteful. In my version, it currently looks like this:

    func DecodeVector(dec *gob.Decoder, v *mat64.Vector) *mat64.Vector {
        floats := v.RawVector().Data
        dec.Decode(&floats)
        return mat64.NewVector(len(floats), floats)
    }

While I'm making some attempt to reuse the raw `[]float64` slice of an existing vector, there is actually a good chance a new slice will be created, and on top of that, a new `mat64.Vector` instance needs to be created in any case.

My preference would be to be able to say `dec.Decode(&v.RawVector().Data)`, but Go doesn't allow for taking the address of the field `Data` in the non-addressable instance returned by `RawVector()`, which is probably good, because the original `Data` field inside `v` would not be assigned to if a new slice needs to be allocated.

Do you see a possibility to make `gonum/matrix` interact better with `encoding/gob`? Or am I missing a way to have them work better together already? (I'm new to Go, so please accept my apologies if I'm missing something obvious here.)

Thanks,
Pascal

btracey:
If Vector et al were to implement the GobEncoder and GobDecoder interfaces would that fix the problm?

jonlawlor:
I think it just needs to implement MarshalBinary and UnmarshalBinary - see the EncodeDecode example in the gob library.  The example even uses a vector type with an unexported field!

btracey:
[...]

PR #339, opened 2016-02-26
Changed matrix.Vector to be a type alias of blas64.Vector

jonlawlor:
This is a proof of concept and I would not dig deeply into the changes
immediately, because if you don’t like the premise then the rest is
irrelevant.

At first I was thinking about adding a BinaryMarshal and
BinaryUnmarshal to matrix.Vector to satisfy #337, but those methods are
supposed to be composable, so I thought maybe it would make sense to
add to blas64.Vector and then call from matrix.

But I’m really lazy and I’d rather not have to call it from matrix at
all.  By aliasing the type of matrix.Vector to blas64.Vector, it will
automatically pick up any methods from the original type.  As a bonus,
calling blas funcs on matrix vectors is a snap.  I imagine this
approach could be extended to other matrix types, and would allow
library callers to get down to a lower implementation level whenever
they want to take the gloves off.

(github autocorrect keeps trying to call this “composable" idea
“compostable”, hopefully you won’t feel the same…)

### DIFF
--- a/mat64/cholesky.go
+++ b/mat64/cholesky.go
@@ -121,8 +121,8 @@ func (v *Vector) SolveCholeskyVec(chol *Cholesky, b *Vector) error {
 	if v != b {
 		v.CopyVec(b)
 	}
-	blas64.Trsv(blas.Trans, chol.chol.mat, v.mat)
-	blas64.Trsv(blas.NoTrans, chol.chol.mat, v.mat)
+	blas64.Trsv(blas.Trans, chol.chol.mat, blas64.Vector(*v))
+	blas64.Trsv(blas.NoTrans, chol.chol.mat, blas64.Vector(*v))
 	if chol.cond > matrix.ConditionTolerance {
 		return matrix.Condition(chol.cond)
 	}
--- a/mat64/dense.go
+++ b/mat64/dense.go
@@ -168,11 +168,8 @@ func (m *Dense) ColView(j int) *Vector {
 		panic(matrix.ErrColAccess)
 	}
 	return &Vector{
-		mat: blas64.Vector{
-			Inc:  m.mat.Stride,
-			Data: m.mat.Data[j : (m.mat.Rows-1)*m.mat.Stride+j+1],
-		},
-		n: m.mat.Rows,
+		Inc:  m.mat.Stride,
+		Data: m.mat.Data[j : (m.mat.Rows-1)*m.mat.Stride+j+1],
 	}
 }
 
@@ -214,11 +211,8 @@ func (m *Dense) RowView(i int) *Vector {
 		panic(matrix.ErrRowAccess)
 	}
 	return &Vector{
-		mat: blas64.Vector{
-			Inc:  1,
-			Data: m.mat.Data[i*m.mat.Stride : i*m.mat.Stride+m.mat.Cols],
-		},
-		n: m.mat.Cols,
+		Inc:  1,
+		Data: m.mat.Data[i*m.mat.Stride : i*m.mat.Stride+m.mat.Cols],
 	}
 }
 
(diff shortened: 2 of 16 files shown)


jonlawlor:
This is failing on index_bound_checks.go - somehow it tested on my machine without it.  All other tests passed.

jonlawlor:
So my question is basically, why aren't the basic types of mat64 defined as:
```go
type Dense blas64.General
type Symmetric blas64.Symmetric
type Triangular blas64.Triangular
type Vector blas64.Vector
```

Because then users get direct access to `blas` and `lapack` routines with the `matrix` types, and those routines return `matrix` types (after a type conversion).  @btracey, @vladimir-ch and @kortschak I know that you've taken steps to stabilize the interface (https://groups.google.com/forum/#!topic/gonum-dev/PJffDI7BVwc), so maybe this suggestion comes too late for consideration.  And callers can always use `RawMatrix()` or the equivalent to get the `blas` types out.  But you would gain a lot of consistency across packages.

kortschak:
I don't like this, for two reasons: 1) it exposes things to the user that should not be easily exposed - in particular Inc is not a dynamic field in most use cases, so the capacity to change it should not be easily available; less important 2) the cost of calculating the N of the vector is now significantly higher.

The extension to 2D objects is not going to happen since it is prohibited by the fact that they can be resized - we have cap fields that do not exist in the base types. Likewise, here Stride is not something that can be meaningfully arbitrarily modified in most cases. Making it public sends the message that modification is OK.

The fundamental difference between the mat64 types and the blas64 types is that blas64 type are not shape-dynamic, mat64 types are.

btracey:
I agree with Dan in the fundamental objection. Certain matrix fields should not be modified directly (like Inc or Stride), and thus should privately be contained within the type.

I also don't think we want users calling BLAS directly. I would rather have it be that mat64 satisfies the needs of users in the normal case. IMO, the only reason users should be calling BLAS or Lapack directly is to save memory and/or computation time, but not because of lacking functionality in mat64. The BLAS functions, and especially the Lapack functions, are notably harder to use. They don't make the same guarantees about immutability, they don't make the same guarantees about data overlap, and frequently they are not "type stable" -- a matrix enters as a Triangular but ends up as a Symmetric or General. It should be trickier to call them. The lack of composability is a feature here. blas64 is a nice wrapper around the raw BLAS calls, and mat64 serves as a nice wrapper around blas64 in many cases.

jonlawlor:
I figured this would be the answer, and it follows the same line of thought as matlab and numpy.  And you are right that some interfaces would not line up after blas or lapack operations.  However, go code tends to involve less hand holding, so things like modifying the stride when you aren't supposed to is mostly on the user, and it may have a benefit from both a simplicity standpoint (and maybe performance, but it would require work to be able to tell).

Anyway I guess the right way forward with BinaryEncoding is to implement it on the blas types and then call it from matrix, and tack on the additional fields from there.  I think that would be the first method on a blas numeric type (as opposed to an implementation) in the blas package.  Any thoughts as to where it would be appropriate to include it in the package?

kortschak:
I don't think this is the right thing to do. It plays badly with views - a view where col != stride will encode the non-viewed elements as well (unless we explicitly don't which is a bit magical).

btracey:
In addition, we've explicitly avoided adding methods to BLAS types. So far that has seemed like a good decision. Marshaling seems like a very weird place to break that since BLAS has nothing to do with serializing data.

kortschak:
Though to be fair, that may change for col major conversion handling.

jonlawlor:
The reason to put it in blas (and also in matrix) is so that people who are comfortable using blas or lapack can use them without matrix.  There is a need to have binarymarshaller in both matrix and blas.  If you don't want it in blas then people that need it can probably implement it themselves, but I think it should be there, or failing that some kind of blas-serialization package.

kortschak:
encoding/gob would do a perfectly reasonable job at encoding - assuming only Go interop.

jonlawlor:
Closing

PR #342, opened 2016-03-02
mat64: implement Binary(Un)Marshaler for Vector

sbinet:
This CL makes *mat64.Vector implement encoding.BinaryMarshaler and
encoding.BinaryUnmarshaler.

Fixes gonum/matrix#337.

### DIFF
--- a/mat64/dense.go
+++ b/mat64/dense.go
@@ -5,9 +5,6 @@
 package mat64
 
 import (
-	"bytes"
-	"encoding/binary"
-
 	"github.com/gonum/blas"
 	"github.com/gonum/blas/blas64"
 	"github.com/gonum/matrix"
@@ -452,74 +449,4 @@ func (m *Dense) Augment(a, b Matrix) {
 	m.Copy(a)
 	w := m.View(0, ac, br, bc).(*Dense)
 	w.Copy(b)
-}
-
-// MarshalBinary encodes the receiver into a binary form and returns the result.
-//
-// Dense is little-endian encoded as follows:
-//   0 -  8  number of rows    (int64)
-//   8 - 16  number of columns (int64)
-//  16 - ..  matrix data elements (float64)
-//           [0,0] [0,1] ... [0,ncols-1]
-//           [1,0] [1,1] ... [1,ncols-1]
-//           ...
-//           [nrows-1,0] ... [nrows-1,ncols-1]
-func (m Dense) MarshalBinary() ([]byte, error) {
-	buf := bytes.NewBuffer(make([]byte, 0, m.mat.Rows*m.mat.Cols*sizeFloat64+2*sizeInt64))
-	err := binary.Write(buf, defaultEndian, int64(m.mat.Rows))
-	if err != nil {
-		return nil, err
-	}
-	err = binary.Write(buf, defaultEndian, int64(m.mat.Cols))
-	if err != nil {
-		return nil, err
-	}
-
-	for i := 0; i < m.mat.Rows; i++ {
-		for _, v := range m.rowView(i) {
-			err = binary.Write(buf, defaultEndian, v)
-			if err != nil {
-				return nil, err
-			}
-		}
-	}
-	return buf.Bytes(), err
-}
-
-// UnmarshalBinary decodes the binary form into the receiver.
-// It panics if the receiver is a non-zero Dense matrix.
-//
-// See MarshalBinary for the on-disk layout.
-func (m *Dense) UnmarshalBinary(data []byte) error {
-	if !m.isZero() {
-		panic("mat64: unmarshal into non-zero matrix")
-	}
-
-	buf := bytes.NewReader(data)
-	var rows int64
-	err := binary.Read(buf, defaultEndian, &rows)
-	if err != nil {
-		return err
-	}
-	var cols int64
-	err = binary.Read(buf, defaultEndian, &cols)
-	if err != nil {
-		return err
-	}
-
-	m.mat.Rows = int(rows)
-	m.mat.Cols = int(cols)
-	m.mat.Stride = int(cols)
-	m.capRows = int(rows)
-	m.capCols = int(cols)
-	m.mat.Data = use(m.mat.Data, m.mat.Rows*m.mat.Cols)
-
-	for i := range m.mat.Data {
-		err = binary.Read(buf, defaultEndian, &m.mat.Data[i])
-		if err != nil {
-			return err
-		}
-	}
-
-	return err
 }
--- a/mat64/io.go
+++ b/mat64/io.go
@@ -5,14 +5,182 @@
 package mat64
 
 import (
+	"bytes"
 	"encoding/binary"
+	"errors"
+)
+
+const (
+	// maxLen is the biggest slice/array len one can create on a 32/64b platform.
+	maxLen = int64(int(^uint(0) >> 1))
 )
 
 var (
-	littleEndian  = binary.LittleEndian
-	bigEndian     = binary.BigEndian
-	defaultEndian = littleEndian
-
 	sizeInt64   = binary.Size(int64(0))
 	sizeFloat64 = binary.Size(float64(0))
+
+	errTooBig    = errors.New("mat64: resulting data slice too big")
+	errBadBuffer = errors.New("mat64: data buffer size mismatch")
+	errBadSize   = errors.New("mat64: invalid dimension")
 )
+
+// MarshalBinary encodes the receiver into a binary form and returns the result.
+//
+// Dense is little-endian encoded as follows:
+//   0 -  7  number of rows    (int64)
+//   8 - 15  number of columns (int64)
+//  16 - ..  matrix data elements (float64)
+//           [0,0] [0,1] ... [0,ncols-1]
+//           [1,0] [1,1] ... [1,ncols-1]
+//           ...
+//           [nrows-1,0] ... [nrows-1,ncols-1]
+func (m Dense) MarshalBinary() ([]byte, error) {
+	bufLen := int64(m.mat.Rows)*int64(m.mat.Cols)*int64(sizeFloat64) + 2*int64(sizeInt64)
+	if bufLen <= 0 {
+		// bufLen is too big and has wrapped around.
+		return nil, errTooBig
+	}
+
+	buf := bytes.NewBuffer(make([]byte, 0, bufLen))
+	err := binary.Write(buf, binary.LittleEndian, int64(m.mat.Rows))
+	if err != nil {
+		return nil, err
+	}
+	err = binary.Write(buf, binary.LittleEndian, int64(m.mat.Cols))
+	if err != nil {
+		return nil, err
+	}
+
+	r, c := m.Dims()
+	for i := 0; i < r; i++ {
+		for j := 0; j < c; j++ {
+			err = binary.Write(buf, binary.LittleEndian, m.at(i, j))
+			if err != nil {
+				return nil, err
+			}
+		}
+	}
+
+	return buf.Bytes(), nil
+}
+
+// UnmarshalBinary decodes the binary form into the receiver.
+// It panics if the receiver is a non-zero Dense matrix.
+//
+// See MarshalBinary for the on-disk layout.
+//
+// Limited checks on the validity of the binary input are performed:
+//  - matrix.ErrShape is returned if the number of rows or columns is negative,
+//  - an error is returned if the resulting Dense matrix is too
+//  big for the current architecture (e.g. a 16GB matrix written by a
+//  64b application and read back from a 32b application.)
+// UnmarshalBinary does not limit the size of the unmarshaled matrix, and so
+// it should not be used on untrusted data.
+func (m *Dense) UnmarshalBinary(data []byte) error {
+	if !m.isZero() {
+		panic("mat64: unmarshal into non-zero matrix")
+	}
+
+	buf := bytes.NewReader(data)
+	var rows int64
+	err := binary.Read(buf, binary.LittleEndian, &rows)
+	if err != nil {
+		return err
+	}
+	var cols int64
+	err = binary.Read(buf, binary.LittleEndian, &cols)
+	if err != nil {
+		return err
+	}
+
+	if rows < 0 || cols < 0 {
+		return errBadSize
+	}
+
+	size := rows * cols
+	if int(size) < 0 || size > maxLen {
+		return errTooBig
+	}
+
+	if len(data) != int(size)*sizeFloat64+2*sizeInt64 {
+		return errBadBuffer
+	}
+
+	m.mat.Rows = int(rows)
+	m.mat.Cols = int(cols)
+	m.mat.Stride = int(cols)
+	m.capRows = int(rows)
+	m.capCols = int(cols)
+	m.mat.Data = use(m.mat.Data, int(size))
+
+	return binary.Read(buf, binary.LittleEndian, m.mat.Data)
+}
+
+// MarshalBinary encodes the receiver into a binary form and returns the result.
+//
+// Vector is little-endian encoded as follows:
+//   0 -  7  number of elements     (int64)
+//   8 - ..  vector's data elements (float64)
+func (v Vector) MarshalBinary() ([]byte, error) {
+	bufLen := int64(sizeInt64) + int64(v.n)*int64(sizeFloat64)
+	if bufLen <= 0 {
+		// bufLen is too big and has wrapped around.
+		return nil, errTooBig
+	}
+
+	buf := bytes.NewBuffer(make([]byte, 0, bufLen))
+	err := binary.Write(buf, binary.LittleEndian, int64(v.n))
+	if err != nil {
+		return nil, err
+	}
+
+	for i := 0; i < v.n; i++ {
+		err = binary.Write(buf, binary.LittleEndian, v.at(i))
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return buf.Bytes(), nil
+}
+
+// UnmarshalBinary decodes the binary form into the receiver.
+// It panics if the receiver is a non-zero Vector.
+//
+// See MarshalBinary for the on-disk layout.
+//
+// Limited checks on the validity of the binary input are performed:
+//  - matrix.ErrShape is returned if the number of rows is negative,
+//  - an error is returned if the resulting Vector is too
+//  big for the current architecture (e.g. a 16GB vector written by a
+//  64b application and read back from a 32b application.)
+// UnmarshalBinary does not limit the size of the unmarshaled vector, and so
+// it should not be used on untrusted data.
+func (v *Vector) UnmarshalBinary(data []byte) error {
+	if !v.isZero() {
+		panic("mat64: unmarshal into non-zero vector")
+	}
+
+	buf := bytes.NewReader(data)
+	var n int64
+	err := binary.Read(buf, binary.LittleEndian, &n)
+	if err != nil {
+		return err
+	}
+
+	if n < 0 {
+		return errBadSize
+	}
+	if n > maxLen {
+		return errTooBig
+	}
+	if len(data) != int(n)*sizeFloat64+sizeInt64 {
+		return errBadBuffer
+	}
+
+	v.n = int(n)
+	v.mat.Inc = 1
+	v.mat.Data = use(v.mat.Data, v.n)
+
+	return binary.Read(buf, binary.LittleEndian, v.mat.Data)
+}
--- a/mat64/io_test.go
+++ b/mat64/io_test.go
@@ -4,28 +4,134 @@
 
 package mat64
 
-import "testing"
+import (
+	"bytes"
+	"encoding"
+	"math"
+	"testing"
 
-func TestDenseRW(t *testing.T) {
-	for i, test := range []*Dense{
-		NewDense(0, 0, []float64{}),
-		NewDense(2, 2, []float64{1, 2, 3, 4}),
-		NewDense(2, 3, []float64{1, 2, 3, 4, 5, 6}),
-		NewDense(3, 2, []float64{1, 2, 3, 4, 5, 6}),
-		NewDense(3, 3, []float64{1, 2, 3, 4, 5, 6, 7, 8, 9}),
-		NewDense(3, 3, []float64{1, 2, 3, 4, 5, 6, 7, 8, 9}).View(0, 0, 2, 2).(*Dense),
-		NewDense(3, 3, []float64{1, 2, 3, 4, 5, 6, 7, 8, 9}).View(1, 1, 2, 2).(*Dense),
-		NewDense(3, 3, []float64{1, 2, 3, 4, 5, 6, 7, 8, 9}).View(0, 1, 3, 2).(*Dense),
-	} {
-		buf, err := test.MarshalBinary()
+	"github.com/gonum/blas/blas64"
+)
+
+var (
+	_ encoding.BinaryMarshaler   = (*Dense)(nil)
+	_ encoding.BinaryUnmarshaler = (*Dense)(nil)
+	_ encoding.BinaryMarshaler   = (*Vector)(nil)
+	_ encoding.BinaryUnmarshaler = (*Vector)(nil)
+)
+
+var denseData = []struct {
+	raw  []byte
+	want *Dense
+	eq   func(got, want Matrix) bool
+}{
+	{
+		raw:  []byte("\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"),
+		want: NewDense(0, 0, []float64{}),
+		eq:   Equal,
+	},
+	{
+		raw:  []byte("\x02\x00\x00\x00\x00\x00\x00\x00\x02\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xf0?\x00\x00\x00\x00\x00\x00\x00@\x00\x00\x00\x00\x00\x00\b@\x00\x00\x00\x00\x00\x00\x10@"),
+		want: NewDense(2, 2, []float64{1, 2, 3, 4}),
+		eq:   Equal,
+	},
+	{
+		raw:  []byte("\x02\x00\x00\x00\x00\x00\x00\x00\x03\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xf0?\x00\x00\x00\x00\x00\x00\x00@\x00\x00\x00\x00\x00\x00\b@\x00\x00\x00\x00\x00\x00\x10@\x00\x00\x00\x00\x00\x00\x14@\x00\x00\x00\x00\x00\x00\x18@"),
+		want: NewDense(2, 3, []float64{1, 2, 3, 4, 5, 6}),
+		eq:   Equal,
+	},
+	{
+		raw:  []byte("\x03\x00\x00\x00\x00\x00\x00\x00\x02\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xf0?\x00\x00\x00\x00\x00\x00\x00@\x00\x00\x00\x00\x00\x00\b@\x00\x00\x00\x00\x00\x00\x10@\x00\x00\x00\x00\x00\x00\x14@\x00\x00\x00\x00\x00\x00\x18@"),
+		want: NewDense(3, 2, []float64{1, 2, 3, 4, 5, 6}),
+		eq:   Equal,
+	},
+	{
+		raw:  []byte("\x03\x00\x00\x00\x00\x00\x00\x00\x03\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xf0?\x00\x00\x00\x00\x00\x00\x00@\x00\x00\x00\x00\x00\x00\b@\x00\x00\x00\x00\x00\x00\x10@\x00\x00\x00\x00\x00\x00\x14@\x00\x00\x00\x00\x00\x00\x18@\x00\x00\x00\x00\x00\x00\x1c@\x00\x00\x00\x00\x00\x00 @\x00\x00\x00\x00\x00\x00\"@"),
+		want: NewDense(3, 3, []float64{1, 2, 3, 4, 5, 6, 7, 8, 9}),
+		eq:   Equal,
+	},
+	{
+		raw:  []byte("\x02\x00\x00\x00\x00\x00\x00\x00\x02\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xf0?\x00\x00\x00\x00\x00\x00\x00@\x00\x00\x00\x00\x00\x00\x10@\x00\x00\x00\x00\x00\x00\x14@"),
+		want: NewDense(3, 3, []float64{1, 2, 3, 4, 5, 6, 7, 8, 9}).View(0, 0, 2, 2).(*Dense),
+		eq:   Equal,
+	},
+	{
+		raw:  []byte("\x02\x00\x00\x00\x00\x00\x00\x00\x02\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x14@\x00\x00\x00\x00\x00\x00\x18@\x00\x00\x00\x00\x00\x00 @\x00\x00\x00\x00\x00\x00\"@"),
+		want: NewDense(3, 3, []float64{1, 2, 3, 4, 5, 6, 7, 8, 9}).View(1, 1, 2, 2).(*Dense),
+		eq:   Equal,
+	},
+	{
+		raw:  []byte("\x03\x00\x00\x00\x00\x00\x00\x00\x02\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00@\x00\x00\x00\x00\x00\x00\b@\x00\x00\x00\x00\x00\x00\x14@\x00\x00\x00\x00\x00\x00\x18@\x00\x00\x00\x00\x00\x00 @\x00\x00\x00\x00\x00\x00\"@"),
+		want: NewDense(3, 3, []float64{1, 2, 3, 4, 5, 6, 7, 8, 9}).View(0, 1, 3, 2).(*Dense),
+		eq:   Equal,
+	},
+	{
+		raw:  []byte("\x01\x00\x00\x00\x00\x00\x00\x00\x04\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xf0\xff\x00\x00\x00\x00\x00\x00\xf0\u007f\x01\x00\x00\x00\x00\x00\xf8\u007f"),
+		want: NewDense(1, 4, []float64{0, math.Inf(-1), math.Inf(+1), math.NaN()}),
+		eq: func(got, want Matrix) bool {
+			for _, v := range []bool{
+				got.At(0, 0) == 0,
+				math.IsInf(got.At(0, 1), -1),
+				math.IsInf(got.At(0, 2), +1),
+				math.IsNaN(got.At(0, 3)),
+			} {
+				if !v {
+					return false
+				}
+			}
+			return true
+		},
+	},
+}
+
+func TestDenseMarshal(t *testing.T) {
+	for i, test := range denseData {
+		buf, err := test.want.MarshalBinary()
 		if err != nil {
-			t.Errorf("error encoding test #%d: %v\n", i, err)
+			t.Errorf("error encoding test-%d: %v\n", i, err)
+			continue
 		}
 
-		nrows, ncols := test.Dims()
+		nrows, ncols := test.want.Dims()
 		sz := nrows*ncols*sizeFloat64 + 2*sizeInt64
 		if len(buf) != sz {
-			t.Errorf("encoded size test #%d: want=%d got=%d\n", i, sz, len(buf))
+			t.Errorf("encoded size test-%d: want=%d got=%d\n", i, sz, len(buf))
+		}
+
+		if !bytes.Equal(buf, test.raw) {
+			t.Errorf("error encoding test-%d: bytes mismatch.\n got=%q\nwant=%q\n",
+				i,
+				string(buf),
+				string(test.raw),
+			)
+			continue
+		}
+	}
+}
+
+func TestDenseUnmarshal(t *testing.T) {
+	for i, test := range denseData {
+		var v Dense
+		err := v.UnmarshalBinary(test.raw)
+		if err != nil {
+			t.Errorf("error decoding test-%d: %v\n", i, err)
+			continue
+		}
+		if !test.eq(&v, test.want) {
+			t.Errorf("error decoding test-%d: values differ.\n got=%v\nwant=%v\n",
+				i,
+				&v,
+				test.want,
+			)
+		}
+	}
+}
+
+func TestDenseIORoundTrip(t *testing.T) {
+	for i, test := range denseData {
+		buf, err := test.want.MarshalBinary()
+		if err != nil {
+			t.Errorf("error encoding test #%d: %v\n", i, err)
 		}
 
 		var got Dense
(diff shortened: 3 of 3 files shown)


mat64/io_test.go
```diff
@@ -39,3 +49,36 @@ func TestDenseRW(t *testing.T) {
 		}
 	}
 }
+
+func TestVectorRW(t *testing.T) {
+	for i, test := range []*Vector{
+		NewVector(0, []float64{}),
```
jonlawlor: Could you add in cases with `math.NaN()` and `math.Inf(1), math.Inf(-1)` ?

sbinet: done.

jonlawlor: Also, what do we want to happen if the binary `n` is invalid, such as a negative value or larger than the size of the slice?

sbinet: is it really something that can happen?
https://github.com/gonum/matrix/blob/master/mat64/vector.go#L35 should insure that `n == len(data)`, always.

or are you wondering whether we should make sure the panic, when unmarshalling a corrupted binary file, is crystal clear?

jonlawlor: It is really just a question.  I think it would return an EOF error if `n > len(data)`, and panic when `n < 0`.  That sounds good to me, but I wanted it to be considered.

btracey: I don't think I/O should panic. I/O is a file error, it's not a programmer error. It is not a knowable error before runtime and it is possible to handle it, so it should be an error.

jonlawlor: Then you'll need to put in a check for `n < 0` because otherwise the `Data: make([]float64, v.n),` line will panic.

sbinet: fair enough.

sbinet:
PTAL

mat64/dense.go
```diff
@@ -512,7 +512,13 @@ func (m *Dense) UnmarshalBinary(data []byte) error {
 	m.mat.Stride = int(cols)
 	m.capRows = int(rows)
 	m.capCols = int(cols)
-	m.mat.Data = use(m.mat.Data, m.mat.Rows*m.mat.Cols)
+
```
btracey: They need to individually checked for being negative. 
Move the check before assigning to m. 
I'm not sure that check for overflow is correct, on 64 bit systems the matrix could be larger. The check can be something like 
````
size := rows * cols
if size < 0 || int64(int(size)) != size { ...}
````

Comment (btracey) [inline mat64/dense.go]: Oh, sorry, misunderstood. Still though, if it overflows maxint it's not going to be > maxInt on 64 bit systems.

File: mat64/vector.go [hunk 3]
```diff
@@ -357,6 +360,61 @@ func (v *Vector) MulVec(a Matrix, b *Vector) {
 	}
 }
 
+// MarshalBinary encodes the receiver into a binary form and returns the result.
```
btracey: Do you think it's worth moving this code and the similar code for *Dense into io.go? This way all of the constants are in closer view.

mat64/dense.go
sbinet: of course not but it would wrap around and become negative :)
I'll add a check on rows and cols.

mat64/vector.go
```diff
@@ -357,6 +360,61 @@ func (v *Vector) MulVec(a Matrix, b *Vector) {
...
+//   0 -  8  number of elements     (int64)
+//   8 - ..  vector's data elements (float64)
+func (v Vector) MarshalBinary() ([]byte, error) {
+	buf := bytes.NewBuffer(make([]byte, 0, sizeInt64+v.n*sizeFloat64))
+	err := binary.Write(buf, defaultEndian, int64(v.n))
+	if err != nil {
+		return nil, err
+	}
+
+	for _, val := range v.mat.Data {
```
btracey: This isn't right, because Vectors can have non-uniform inc. See Vector.MulElemVec for appropriate indexing.

mat64/vector.go
```diff
@@ -357,6 +360,61 @@ func (v *Vector) MulVec(a Matrix, b *Vector) {
...
+	}
+	return buf.Bytes(), err
+}
+
+// UnmarshalBinary decodes the binary form into the receiver.
+// It panics if the receiver is a non-zero Vector.
+//
+// See MarshalBinary for the on-disk layout.
+func (v *Vector) UnmarshalBinary(data []byte) error {
+	if v.n != 0 || len(v.mat.Data) != 0 {
```
btracey: use v.isZero()

mat64/io_test.go
```diff
@@ -39,3 +50,86 @@ func TestDenseRW(t *testing.T) {
...
+		math.IsInf(got.At(0, 2), +1),
+		math.IsNaN(got.At(0, 3)),
+	} {
+		if !v {
+			t.Errorf("r/w test d.At(0,%d) failed\nwant=%#v\n got=%#v\n", i, d.At(0, i), got.At(0, i))
+		}
+	}
+}
+
+func TestVectorRW(t *testing.T) {
```
btracey: Add a test with non-uniform Inc please.

mat64/vector.go
sbinet: oh! right.

alternatively, perhaps write the whole `v.mat.Data` + `v.mat.Inc` (and read them back) ?
it's perhaps less optimized, but introduces less hysteresis effects...

mat64/io_test.go
sbinet: right.

mat64/vector.go
btracey: MarshalBinary and UnmarshalBinary are about writing the values in the matrix. They are intentionally cross-platform, and not about replicating the Go type. I think this is for the best, even though it does create hysteresis effects.

sbinet: it could be surprising from outside and break invariants of users' code. Consider

```go
package my

type View struct {
    vec mat64.Vector
    inc int
}

func (view View) MarshalBinary(...) {...}
func (view *View) UnmarshalBinary(...) {...}
```

that could potentially break users' marshalling code.
we could of course document it, but it seems to me like an unnecessary hurdle (and it would just move the problem to someone else having to also properly document this gotcha)

btracey: Dense has the same problem. It doesn't marshal all of the data, just what is returned by At. Dense.MarshalBinary can be considered as doing 
````
r,c := d.Dims()
for i := 0; i < r; i++ {
   for j := 0; j < c; j++ {
        marshal(d.At(i,j))
   }
}
````
and similarly for a Vector
````
for i := 0; i < v.Len(); i++ {
    marshal(v.At(i,0))
}
````

I agree there are good uses for marshaling code that is hysteresis free. Gob would be a good candidate for using a marshaling that is specific for the internal representation. MarshalBinary is the best candidate for marshaling in a way that is clear and easy to use cross-language. The best way to keep it easily cross-platform is to only marshal the viewable data in the type.

I'm not exactly sure what your example illustrates. Why would the user want to individually track the inc of the vector? Non-unity incs only arise in mat64 when there is a ColView of a Dense. Knowing the number of columns in the Dense is not sufficient to know the Inc, because the user also has to know the Dense has never been viewed.

Comment (sbinet) [inline mat64/vector.go]: I guess my main worry is that breaking the invariant `v === unmarshal(marshal(v))` where `===` is the identity (not just mere equality) could break people's code in interesting ways and bite us in the rear area down the road.

besides, the current implementation *is* cross-platform and cross-language friendly.
a `python` module can easily read the resulting binary file and discard, if deemed unrelevant, the extra `Inc` int64 (that I would, of course, need to add to the binary layout)

File: mat64/vector.go (hunk 5 above)
Comment (sbinet) [inline mat64/vector.go]: done.

File: mat64/vector.go (hunk 4 above)
Comment (btracey) [inline mat64/vector.go]: Ensuring that ````v === unmarshal(marshal(v))```` depends on the internal representation. Holding that invariant either means never changing the internal representation, or means changing the binary format in the future. Neither of those seem in the spirit of BinaryMarshal. For example, in the future we may need to add a "Cap" field to Vector like we have in Dense.

I understand your concern, I just think it's not the point of BinaryMarshal. If the internal representation were to change, I would say that BinaryMarshal shouldn't change, while GobEncode should. For example, maintaining i/o equality means writing the entirety of v.mat.Data, which could be much bigger than v.Len() in the presence of inc and views.

Comment (kortschak) [inline mat64/vector.go]: The Dense marshaling is only done on the visible data. Vector should do the same. If it becomes useful to serialise computational state for checkpointing (I guess it will), then as @btracey says, that can be done by gob, though even that it fraught; if you gob encode two views of the same backing matrix do you expect on deserialisation for the views to still be backed by the same data? You might want that and the behaviour might suggest it, but I don't think we will even bother trying to achieve that.

Comment (sbinet) [inline mat64/vector.go]: fine. data reduction it is :) (I'll push something tomorrow)

File: mat64/io_386.go [hunk 7]
```diff
@@ -0,0 +1,11 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package mat64
+
+import "math"
+
+const (
+	maxInt = math.MaxInt32
```
vladimir-ch: How is maxInt defined on other architectures?

mat64/io_386.go
```diff
@@ -0,0 +1,11 @@
+// Copyright ©2015 The gonum Authors. All rights reserved.
```
vladimir-ch: 2016? Also elsewhere.

mat64/dense.go
```diff
@@ -512,7 +512,13 @@ func (m *Dense) UnmarshalBinary(data []byte) error {
...
+	size := rows * cols
+	if rows < 0 || cols < 0 || int(size) < 0 || size > maxInt {
+		return matrix.ErrShape
+	}
+
+	m.mat.Data = use(m.mat.Data, int(size))
 
 	for i := range m.mat.Data {
 		err = binary.Read(buf, defaultEndian, &m.mat.Data[i])
```
vladimir-ch: Why not use `binary.LittleEndian` directly instead of `defaultEndian` if the documentation explicitly says `little-endian encoded`?

mat64/io_386.go
vladimir-ch: Rather, why not use `const maxInt = int(^uint(0) >> 1)` directly in `UnmarshallBinary` instead of package-level const?

mat64/dense.go
vladimir-ch: Alternatively, there could be a limit on the number of elements less than or equal to MaxInt32. `gob` has 1GB of memory limit, MaxInt32 of float64 is 16GB of memory.

mat64/vector.go
```diff
@@ -357,6 +360,61 @@ func (v *Vector) MulVec(a Matrix, b *Vector) {
 	}
 }
 
+// MarshalBinary encodes the receiver into a binary form and returns the result.
+//
+// Vector is little-endian encoded as follows:
+//   0 -  8  number of elements     (int64)
```
vladimir-ch: `0 -  7` ?

mat64/vector.go
```diff
@@ -357,6 +360,61 @@ func (v *Vector) MulVec(a Matrix, b *Vector) {
...
+	if n < 0 || n > maxInt {
+		return matrix.ErrShape
+	}
+
+	v.n = int(n)
+	v.mat.Inc = 1
+	v.mat.Data = use(v.mat.Data, v.n)
+
+	for i := range v.mat.Data {
+		err = binary.Read(buf, defaultEndian, &v.mat.Data[i])
```
vladimir-ch: The third argument could be `v.mat.Data` directly. The documentation says `... or a slice of fixed-size values`.

mat64/dense.go
sbinet: why would you artificially limit the number of elements an array/slice/vector could hold?
if you bought a high-end machine with loads of memory, it's to be used :)

unless you disagree, I'll keep the check on the number of elements in the slice.

mat64/vector.go
sbinet: fine by me.

mat64/dense.go
kortschak: I think it's OK to not limit size, but that should be noted with a warning that this should not be used from untrusted sources. (A similar warning should be placed on Dense's method.)

mat64/io_386.go
sbinet: done. (use const-maxInt)

mat64/dense.go
sbinet: shorter to type and easier to document/enforce in the code, plus: consistency?

actually, I might add a little decoder/encoder helper:

```go
func binWrite(w io.Writer, val interface{}) error {
  return binary.Write(w, binary.LittleEndian, val)
}

func binRead(r io.Reader, valptr interface{}) error {
  return binary.Read(r, binary.LittleEndian, valptr)
}
```

what do you think ?

or, even:

```go
type encoder struct {
  w io.Writer
  err error
}

func (enc *encoder) write(val interface{}) {
  if enc.err != nil {
    return
  }
  enc.err = binary.Write(enc.w, binary.LittleEndian, val)
}

type decoder struct {
  r io.Reader
  err error
}

func (dec *decoder) read(valptr interface{}) {
  if dec.err != nil { return }
  dec.err = binary.Read(dec.r, binary.LittleEndian, valptr)
}
```

to streamline error handling.

mat64/vector.go
sbinet: what about `[0 - 8)` ?

mat64/vector.go
sbinet: ack. but that won't work well with `v.mat.Inc` :)

mat64/dense.go
vladimir-ch: I'm also ok with not limiting size, it was just a side remark.

mat64/vector.go
vladimir-ch: Why? `Inc` is 1 here. Or am I missing something?

mat64/vector.go
vladimir-ch: I don't have a preference, please pick one :-)

mat64/dense.go
vladimir-ch: I still think that
```go
err = binary.Read(buf, binary.LittleEndian, m.mat.Data)
```
is the simplest option.

sbinet:
PTAL

mat64/dense.go
kortschak: The warning is important though. Otherwise someone will use it for a wire, and then get clobbered by a malicious sender.

mat64/io.go
```diff
@@ -5,14 +5,152 @@
...
+	m.mat.Stride = int(cols)
+	m.capRows = int(rows)
+	m.capCols = int(cols)
+
+	// maxLen is the biggest slice/array len one can create on a 32/64b platform.
+	const maxLen = int64(int(^uint(0) >> 1))
+
+	size := rows * cols
+	if rows < 0 || cols < 0 || int(size) < 0 || size > maxLen {
+		return matrix.ErrShape
```
kortschak: I don't think this is ErrShape, you can define a private error along the lines of "mat64: too big".

mat64/io.go
```diff
@@ -5,14 +5,152 @@
...
+	err := binary.Read(buf, binLE, &n)
+	if err != nil {
+		return err
+	}
+
+	// maxLen is the biggest slice/array len one can create on a 32/64b platform.
+	const maxLen = int64(int(^uint(0) >> 1))
+
+	if n < 0 || n > maxLen {
+		return matrix.ErrShape
```
kortschak: Same here.

mat64/io.go
```diff
@@ -5,14 +5,152 @@
...
+	return err
+}
+
+// MarshalBinary encodes the receiver into a binary form and returns the result.
+//
+// Vector is little-endian encoded as follows:
+//   0 -  8  number of elements     (int64)
+//   8 - ..  vector's data elements (float64)
+func (v Vector) MarshalBinary() ([]byte, error) {
+	buf := bytes.NewBuffer(make([]byte, 0, sizeInt64+v.n*sizeFloat64))
```
kortschak: Can we check that this is in bounds?

mat64/io.go
```diff
@@ -5,14 +5,152 @@
...
+// Dense is little-endian encoded as follows:
+//   0 -  8  number of rows    (int64)
+//   8 - 16  number of columns (int64)
+//  16 - ..  matrix data elements (float64)
+//           [0,0] [0,1] ... [0,ncols-1]
+//           [1,0] [1,1] ... [1,ncols-1]
+//           ...
+//           [nrows-1,0] ... [nrows-1,ncols-1]
+func (m Dense) MarshalBinary() ([]byte, error) {
+	buf := bytes.NewBuffer(make([]byte, 0, m.mat.Rows*m.mat.Cols*sizeFloat64+2*sizeInt64))
```
kortschak: Can we check that this is in bounds?

mat64/io.go
```diff
@@ -5,14 +5,152 @@
...
-
+	binLE       = binary.LittleEndian
 	sizeInt64   = binary.Size(int64(0))
 	sizeFloat64 = binary.Size(float64(0))
 )
+
+// MarshalBinary encodes the receiver into a binary form and returns the result.
+//
+// Dense is little-endian encoded as follows:
+//   0 -  8  number of rows    (int64)
```
kortschak: Agree with @vladimir-ch, make these 0 - 7 etc. This is a format doc, not a mathematics statement.

mat64/io.go
```diff
@@ -5,14 +5,152 @@
...
+		}
+	}
+
+	return err
+}
+
+// MarshalBinary encodes the receiver into a binary form and returns the result.
+//
+// Vector is little-endian encoded as follows:
+//   0 -  8  number of elements     (int64)
```
kortschak: Same here.

mat64/io.go
```diff
@@ -5,14 +5,152 @@
...
+
+	"github.com/gonum/matrix"
 )
 
 var (
-	littleEndian  = binary.LittleEndian
-	bigEndian     = binary.BigEndian
-	defaultEndian = littleEndian
-
+	binLE       = binary.LittleEndian
```
kortschak: Can we delete this and just use binary.LittleEndian throughout please.

mat64/io.go
sbinet: sorry, what should we check is in bounds?

mat64/io.go
```diff
@@ -5,14 +5,152 @@
...
+			err = binary.Write(buf, binLE, m.at(i, j))
+			if err != nil {
+				return nil, err
+			}
+		}
+	}
+	return buf.Bytes(), err
+}
+
+// UnmarshalBinary decodes the binary form into the receiver.
```
kortschak: "UnmarshalBinary does not limit the size of the marshaled matrix, and so UnmarshalBinary should not be used on untrusted data."

mat64/io.go
```diff
@@ -5,14 +5,152 @@
...
+	for i := 0; i < v.n; i++ {
+		err = binary.Write(buf, binLE, v.mat.Data[i*v.mat.Inc])
+		if err != nil {
+			return nil, err
+		}
+	}
+	return buf.Bytes(), err
+}
+
+// UnmarshalBinary decodes the binary form into the receiver.
```
kortschak: "UnmarshalBinary does not limit the size of the marshaled matrix, and so UnmarshalBinary should not be used on untrusted data."

mat64/io.go
```diff
@@ -5,14 +5,152 @@
...
+
+	buf := bytes.NewReader(data)
+	var n int64
+	err := binary.Read(buf, binLE, &n)
+	if err != nil {
+		return err
+	}
+
+	// maxLen is the biggest slice/array len one can create on a 32/64b platform.
+	const maxLen = int64(int(^uint(0) >> 1))
```
kortschak: This can be package-level - deleting the other instance.

mat64/io.go
kortschak: `sizeInt64+v.n*sizeFloat64`

mat64/io.go
kortschak: `m.mat.Rows*m.mat.Cols*sizeFloat64+2*sizeInt64`

mat64/io.go
```diff
@@ -5,14 +5,152 @@
...
+
+	size := rows * cols
+	if rows < 0 || cols < 0 || int(size) < 0 || size > maxLen {
+		return matrix.ErrShape
+	}
+
+	m.mat.Data = use(m.mat.Data, int(size))
+
+	for i := range m.mat.Data {
+		err = binary.Read(buf, binLE, &m.mat.Data[i])
```
vladimir-ch: I think passing the slice gives the same result. I would replace this loop with:
```go
err = binary.Read(buf, binary.LittleEndian, m.mat.Data)
if err != nil {
	return err
}
```

mat64/io.go
```diff
@@ -5,14 +5,152 @@
...
+	if n < 0 || n > maxLen {
+		return matrix.ErrShape
+	}
+
+	v.n = int(n)
+	v.mat.Inc = 1
+	v.mat.Data = use(v.mat.Data, v.n)
+
+	for i := range v.mat.Data {
+		err = binary.Read(buf, binLE, &v.mat.Data[i])
```
vladimir-ch: And also here.

mat64/io.go
kortschak: Agreed.

mat64/io.go
sbinet: done

mat64/dense.go
sbinet: done.

mat64/vector.go
sbinet: done

mat64/vector.go
sbinet: right. done.

mat64/io.go
sbinet: sorry, what should be checked for being in-bounds ?

mat64/io.go
vladimir-ch: Agreed, I didn't realize there were two of these.

mat64/io.go
```diff
@@ -5,14 +5,159 @@
...
-	defaultEndian = littleEndian
-
 	sizeInt64   = binary.Size(int64(0))
 	sizeFloat64 = binary.Size(float64(0))
 )
+
+// MarshalBinary encodes the receiver into a binary form and returns the result.
+//
+// Dense is little-endian encoded as follows:
+//  [ 0 -  8)  number of rows    (int64)
```
kortschak: I'd really prefer 0 - 7 etc.

mat64/io.go
sbinet: ok, done.

mat64/io.go
sbinet: ditto

mat64/io.go
sbinet: done.

mat64/io.go
```diff
@@ -5,14 +5,159 @@
...
+//   [8 - ..)  vector's data elements (float64)
+func (v Vector) MarshalBinary() ([]byte, error) {
+	buf := bytes.NewBuffer(make([]byte, 0, sizeInt64+v.n*sizeFloat64))
+	err := binary.Write(buf, binary.LittleEndian, int64(v.n))
+	if err != nil {
+		return nil, err
+	}
+
+	for i := 0; i < v.n; i++ {
+		err = binary.Write(buf, binary.LittleEndian, v.mat.Data[i*v.mat.Inc])
```
vladimir-ch: Replace `v.mat.Data[i*v.mat.Inc]` with `v.at(i)`.

mat64/io.go
```diff
@@ -5,14 +5,159 @@
...
+
+	if n < 0 || n > maxLen {
+		return matrix.ErrShape
+	}
+
+	v.n = int(n)
+	v.mat.Inc = 1
+	v.mat.Data = use(v.mat.Data, v.n)
+
+	err = binary.Read(buf, binary.LittleEndian, &v.mat.Data)
```
vladimir-ch: Is the `&` necessary?

mat64/io.go
```diff
@@ -5,14 +5,159 @@
...
+	if err != nil {
+		return err
+	}
+
+	return err
```
vladimir-ch: `return nil`

mat64/io.go
```diff
@@ -5,14 +5,159 @@
...
+		return nil, err
+	}
+
+	for i := 0; i < v.n; i++ {
+		err = binary.Write(buf, binary.LittleEndian, v.mat.Data[i*v.mat.Inc])
+		if err != nil {
+			return nil, err
+		}
+	}
+	return buf.Bytes(), err
```
vladimir-ch: `return buf.Bytes(), nil`

mat64/io.go
sbinet: done.

mat64/io.go
sbinet: done.

mat64/io.go
sbinet: done

mat64/io.go
sbinet: done

mat64/io.go
sbinet: done

mat64/io.go
```diff
@@ -5,14 +5,159 @@
...
+	const maxLen = int64(int(^uint(0) >> 1))
+
+	size := rows * cols
+	if rows < 0 || cols < 0 || int(size) < 0 || size > maxLen {
+		return matrix.ErrShape
+	}
+
+	m.mat.Data = use(m.mat.Data, int(size))
+
+	err = binary.Read(buf, binary.LittleEndian, &m.mat.Data)
```
vladimir-ch: Is the '&' necessary?

Also, these 6 lines could be simplified to
```go
return binary.Read(buf, binary.LittleEndian, m.mat.Data)
```

mat64/io.go
vladimir-ch: Could also be simplified to a single return.

mat64/io.go
kortschak: If I have a large matrix on a 386 (16kx16k elements), I could conceivably cause a panic here because m.mat.Rows*m.mat.Cols*sizeFloat64+2*sizeInt64 > 2G. So check that m.mat.Rows*m.mat.Cols*sizeFloat64+2*sizeInt64 <= maxInt and return an error otherwise. Equivalent for Vector.

It's not going to catch much, but if we miss one that's bad.

mat64/io.go
```diff
@@ -5,14 +5,169 @@
...
+	nrows, ncols := m.Dims()
+	for i := 0; i < nrows; i++ {
+		for j := 0; j < ncols; j++ {
+			err = binary.Write(buf, binary.LittleEndian, m.at(i, j))
+			if err != nil {
+				return nil, err
+			}
+		}
+	}
+	return buf.Bytes(), err
```
vladimir-ch: `return buf.Bytes(), nil`

mat64/io.go
sbinet: done (but that's an additional bound check + function call...)

mat64/io.go
sbinet: turns out it isn't:
 https://golang.org/src/encoding/binary/binary.go#L206

done.

mat64/io.go
sbinet: done

mat64/io.go
sbinet: done.

mat64/io.go
sbinet: done.

mat64/io.go
sbinet: right. done.

mat64/io.go
kortschak: We don't do internal bounds checks unless the bounds tag is set, and the
function call will be inlined.

sbinet:
rebased against `origin/master`. PTAL.

mat64/io.go
```diff
@@ -5,14 +5,171 @@
...
+		return matrix.ErrShape
+	}
+
+	if int(size) < 0 || size > maxLen {
+		return errDataTooBig
+	}
+
+	m.mat.Data = use(m.mat.Data, int(size))
+
+	return binary.Read(buf, binary.LittleEndian, m.mat.Data)
```
jonlawlor: Do we want a check to see if the `data []byte` is too long compared to the expected result matrix data?  It won't return an error from `binary.Read`, but it would be an indication that the matrix was not encoded correctly.  See https://play.golang.org/p/k5adVwofft for an example of what I mean.

Ditto with `(v *Vector) UnmarshalBinary(...)`.

sbinet: something like (just before `m.mat.Data = use(m.mat.Data, int(size))`):

```go
if len(data) != (int(size)*sizeFloat64+2*sizeInt64) {
    return errDataBufferTooBig
}
```

right?

fine by me.

sbinet: done.

mat64/io.go
```diff
@@ -5,14 +5,179 @@
...
+	m.capRows = int(rows)
+	m.capCols = int(cols)
+
+	size := rows * cols
+	if rows < 0 || cols < 0 {
+		return matrix.ErrShape
+	}
+
+	if int(size) < 0 || size > maxLen {
+		return errDataTooBig
```
btracey: Shouldn't these errors be public and in matrix? They can be handled so it should be easy to do so. @kortschak ?

mat64/io.go
jonlawlor: Yes, something like that.

mat64/io.go
```diff
@@ -5,14 +5,179 @@
...
+	err := binary.Write(buf, binary.LittleEndian, int64(m.mat.Rows))
+	if err != nil {
+		return nil, err
+	}
+	err = binary.Write(buf, binary.LittleEndian, int64(m.mat.Cols))
+	if err != nil {
+		return nil, err
+	}
+
+	nrows, ncols := m.Dims()
```
vladimir-ch: We normally call these variables `r` and `c`.

mat64/io.go
```diff
@@ -5,14 +5,179 @@
...
+// maxLen is the biggest slice/array len one can create on a 32/64b platform.
+const maxLen = int64(int(^uint(0) >> 1))
 
-	sizeInt64   = binary.Size(int64(0))
-	sizeFloat64 = binary.Size(float64(0))
+var (
+	sizeInt64           = binary.Size(int64(0))
+	sizeFloat64         = binary.Size(float64(0))
+	errDataTooBig       = errors.New("mat64: resulting data slice too big")
+	errDataBufferTooBig = errors.New("mat64: data buffer size mismatch")
```
kortschak: s/errDataTooBig/errTooBig/g
s/errDataBufferTooBig/errBadBuffer/g

mat64/io.go
```diff
@@ -5,14 +5,179 @@
...
+	}
+
+	m.mat.Rows = int(rows)
+	m.mat.Cols = int(cols)
+	m.mat.Stride = int(cols)
+	m.capRows = int(rows)
+	m.capCols = int(cols)
+
+	size := rows * cols
+	if rows < 0 || cols < 0 {
```
vladimir-ch: This check can be done earlier before modifying the receiver.

mat64/io.go
```diff
@@ -5,14 +5,179 @@
...
+		return matrix.ErrShape
+	}
+
+	if int(size) < 0 || size > maxLen {
```
vladimir-ch: This one too.

mat64/io.go
```diff
@@ -5,14 +5,179 @@
...
+		return errDataTooBig
+	}
+
+	if len(data) != int(size)*sizeFloat64+2*sizeInt64 {
```
vladimir-ch: And this one.

mat64/io.go
sbinet: done. (left unexported)

mat64/io.go
sbinet: done

mat64/io.go
sbinet: done

mat64/io.go
sbinet: done

mat64/io.go
sbinet: it seems @kortschak is fine with them left unexported :)

mat64/io.go
sbinet: done.

sbinet:
rebased and squashed.

mat64/io.go
```diff
@@ -5,14 +5,180 @@
...
+	if err != nil {
+		return err
+	}
+	var cols int64
+	err = binary.Read(buf, binary.LittleEndian, &cols)
+	if err != nil {
+		return err
+	}
+
+	size := rows * cols
```
vladimir-ch: Move this line before the next `if`.

mat64/io.go
```diff
@@ -5,14 +5,180 @@
...
+	if rows < 0 || cols < 0 {
+		return matrix.ErrShape
```
vladimir-ch: I'm note sure `ErrShape` is the correct error. Negative rows, cols can come only from corrupted data. This is not doing an operation on matrices of incompatible size.

kortschak: I think you are correct. errBadSize sounds like a good name for whatever this is.

mat64/io.go
```diff
@@ -5,14 +5,180 @@
...
+	if len(data) != int(size)*sizeFloat64+2*sizeInt64 {
+		return errBadBuffer
+	}
+
+	m.mat.Rows = int(rows)
+	m.mat.Cols = int(cols)
+	m.mat.Stride = int(cols)
+	m.capRows = int(rows)
+	m.capCols = int(cols)
+
```
vladimir-ch: Delete line

mat64/io.go
```diff
@@ -5,14 +5,180 @@
...
+	buf := bytes.NewReader(data)
+	var n int64
+	err := binary.Read(buf, binary.LittleEndian, &n)
+	if err != nil {
+		return err
+	}
+
+	if n < 0 {
+		return matrix.ErrShape
```
vladimir-ch: Also here.

mat64/io.go
```diff
@@ -5,14 +5,180 @@
...
+//   0 -  7  number of rows    (int64)
+//   8 - 15  number of columns (int64)
+//  16 - ..  matrix data elements (float64)
+//           [0,0] [0,1] ... [0,ncols-1]
+//           [1,0] [1,1] ... [1,ncols-1]
+//           ...
+//           [nrows-1,0] ... [nrows-1,ncols-1]
+func (m Dense) MarshalBinary() ([]byte, error) {
+	bufLen := int64(m.mat.Rows*m.mat.Cols*sizeFloat64 + 2*sizeInt64)
+	if bufLen > maxLen {
```
vladimir-ch: Can this ever be true?

mat64/io_test.go
```diff
@@ -4,7 +4,20 @@
...
+)
+
+var (
+	_ encoding.BinaryMarshaler   = (*Dense)(nil)
+	_ encoding.BinaryUnmarshaler = (*Dense)(nil)
+	_ encoding.BinaryMarshaler   = (*Vector)(nil)
+	_ encoding.BinaryUnmarshaler = (*Vector)(nil)
+)
 
 func TestDenseRW(t *testing.T) {
```
vladimir-ch: Rename to something more predictable? `TestDenseIO` or `TestDenseMarshalUnmarshall`?

mat64/io_test.go
vladimir-ch: Better name also here?

mat64/io.go
kortschak: Good catch. The conversions should happen prior to the calculation: `int64(m.mat.Rows)*int64(m.mat.Cols)*int64(sizeFloat64) + int64(2*sizeInt64))`.

Also below.

mat64/io_test.go
kortschak: TestDenseIORoundtrip

mat64/io_test.go
kortschak: TestVectorIORoundtrip

mat64/io_test.go
```diff
@@ -39,3 +52,93 @@ func TestDenseRW(t *testing.T) {
...
+			t.Errorf("encoded size test #%d: want=%d got=%d\n", i, sz, len(buf))
+		}
+
+		var got Vector
+		err = got.UnmarshalBinary(buf)
+		if err != nil {
+			t.Errorf("error decoding test #%d: %v\n", i, err)
+		}
+
+		if !Equal(&got, test) {
```
vladimir-ch: If this check fails, we know that there is a bug somewhere but not where. There should be tests that test Marshal and Unmarshal individually.

mat64/io_test.go
```diff
@@ -39,3 +52,93 @@ func TestDenseRW(t *testing.T) {
...
+			t.Errorf("r/w test #%d failed\nwant=%#v\n got=%#v\n", i, test, &got)
+		}
+	}
+}
+
+func TestVectorRWNaNInf(t *testing.T) {
```
vladimir-ch: Can't these NaN and Inf tests be made just a test case within the main test routines?

mat64/io.go
vladimir-ch: The parenthesis are wrong at the end, it should be
`int64(m.mat.Rows)*int64(m.mat.Cols)*int64(sizeFloat64) + 2*int64(sizeInt64)`.

Even better, `sizeFloat64` and `sizeInt64` could be constants.

mat64/io.go
```diff
@@ -5,14 +5,180 @@
...
 
-var (
-	littleEndian  = binary.LittleEndian
-	bigEndian     = binary.BigEndian
-	defaultEndian = littleEndian
+// maxLen is the biggest slice/array len one can create on a 32/64b platform.
+const maxLen = int64(int(^uint(0) >> 1))
 
+var (
 	sizeInt64   = binary.Size(int64(0))
```
vladimir-ch: Do these two have to be variables? The types have their size in their name, we don't need to call `binary.Size`.

mat64/io.go
kortschak: Thanks. They can't be constants of the right type - unsafe.Sizeof returns a uintptr (and would require importing unsafe) and binary.Size cannot be used to set a constant.

vladimir-ch: Alternatively, we could avoid all this and rely on `bytes.Buffer` method panicking with `ErrTooLarge` if the buffer becomes too large. That would probably be the cleanest option.

vladimir-ch: Or take inspiration from `bytes/buffer.go`:
```go
// makeSlice allocates a slice of size n. If the allocation fails, it panics
// with ErrTooLarge.
func makeSlice(n int) []byte {
    // If the make fails, give a known error.
    defer func() {
        if recover() != nil {
            panic(ErrTooLarge)
        }
    }()
    return make([]byte, n)
}
```

vladimir-ch: > They can't be constants of the right type - unsafe.Sizeof returns a uintptr (and would require importing unsafe) and binary.Size cannot be used to set a constant.

I meant hardcoding them :-)

kortschak: >I meant hardcoding them :-)

Yeah, I intentionally ignored that.

kortschak: The failure mode in bytes/buffer is different though. It is a failure for when a contiguous allocation cannot be made, while this failure here would be for spuriously large size descriptors. I'd like to keep the check.

vladimir-ch: > It is a failure for when a contiguous allocation cannot be made, while this failure here would be for spuriously large size descriptors.

Isn't it the same? We care about large size descriptors only because we need to allocate the destination buffer. As long as we can get the buffer and write to it, Marshal does what it is supposed to do. In other words, what is the purpose of this check? To avoid panic in `make` below? Does this check guarantee that it won't panic?

kortschak: There are two ways make can panic. One of them is not recoverable, that one is the one that the pre-check finds.

vladimir-ch: Ok, now it makes sense. Thanks.

vladimir-ch: Is this behavior of `make` described somewhere?

vladimir-ch: I didn't find a description but now I know what you mean with those two panics.

kortschak: There are some posts on golang-nuts by Ian Taylor that explain the situation.

vladimir-ch: If the posts I found are those that you referred to, then I was maybe confused about the unrecoverable panic. What do you mean by the unrecoverable panic?

vladimir-ch: Ok, now I think that I finally (partly) understand. I made a tiny program that tries to allocate a large slice of bytes. When I cross compile with GOARCH=386 and run, the program terminates with unrecoverable panic. When I compile it on amd64 and run, it terminates with a recoverable panic. I don't seem to be able to get the unrecoverable panic on amd64. The length of the slice is 60% of `maxLen`, so I still wonder what is the purpose of this check. On 64bit platforms the condition is never true because we comparse to the _maximum_ integer. On 32bit platforms it can evaluate to true but `make` can obviously crash with slice sizes much lower than maxLen anyway. Please correct me if I am being dense and missing something obvious.

mat64/io_test.go
sbinet: done

mat64/io.go
sbinet: done

mat64/io.go
sbinet: done

mat64/io.go
sbinet: done

mat64/io.go
sbinet: done

mat64/io.go
sbinet: on linux-32b, the maximum addressable slice is a slice of size maxInt32 (ie: 4GB)
the problem is, the linux kernel will reserve 1GB for itself so an application can't use all of the 4GB theoretical address space.
I *believe* this is the issue you are seeing.

mat64/io_test.go
sbinet: done

mat64/io_test.go
sbinet: checking for NaN and Infs is a pain in the neck and muddles a bit the code.
I'd rather keep the happy code path clean and restrict the NaN/Inf-ness ugly checks localized.

mat64/io.go
sbinet: done.

mat64/io_test.go
vladimir-ch: I was thinking that there should be test routines for testing only MarshalBinary (that compare to explicit slice of bytes) and UnmarshalBinary (that get explicit slice of bytes and compare the resulting matrix). I'm not sure if it is worth the effort.

Anyway, Infs can be compared directly and instead of using `Equal` you could use a modified `equal` from `list_test.go`.

mat64/io.go
vladimir-ch: I'm not sure if @kortschak was in favor of changing this. I think it is ok to leave as is.

kortschak: Yes, I'd like this to be legible without a comment.

sbinet: it now reads:

```go
const (
	// maxLen is the biggest slice/array len one can create on a 32/64b platform.
	maxLen = int64(int(^uint(0) >> 1))

	// space binary.Write will require for an int64
	sizeInt64 = 8

	// space binary.Write will require for a float64
	sizeFloat64 = 8
)
```

`binary.(Read|Write)` are very straightforward.
The size required to store `int64`s and `float64`s won't change anytime soon.

sbinet: @kortschak ah. ok.

mat64/io.go
kortschak: Brain is weak. We should be comparing against the maximum architecture-sized int. This still doesn't really protect us that much (unless we make the size smaller). Maybe you're right, we should just let the thing crash. I don't know anymore.

kortschak: @sbinet the issue is described nicely [here](https://groups.google.com/forum/#!topic/golang-dev/wUEWhk2jtHM). Also note that the size of a slice that can be made is 2Gelements rather than 4GB.

kortschak: A possible solution is just to write in the docs that the Unmarshal methods should not be used as part of a wire protocol where the sender is not trusted.

vladimir-ch: I understand but my question was rather what this `if` actually tests. On 64 bits `bufLen > maxLen` is never true. So it seems to be relevant only on 32 bit architectures in cases where the destination slice size is more than 2GB. But even then this check gives no guarantees. If this is well known and everybody is fine with it, then please just document it somehow in a comment.

kortschak: You are of course correct. Sorry.

vladimir-ch: Then I think it is fine to drop this check and replace it with recovering from a panic. After all, we are marshalling a matrix that we already have which means that it is not extremely big so that allocating the destination slice should fail.
We could avoid calling `make`, let `bytes.Buffer` handle growing the buffer and recover `bytes.ErrTooLarge`.

kortschak: So long as we say the source must be trusted that's OK. The payload otherwise could easily be [ff ff ff ff ff ff ff 7f ff ff ff ff ff ff 7f]; there doesn't need to be a backing data slice, just a request for space.

vladimir-ch: But here were are creating the payload and I hope that we can trust the `Dense` receiver upon which `MarshalBinary` is being called.

jonlawlor: How would you know what size of matrix to create if you haven't unmarshalled it yet?

sbinet: sorry but I am now completely confused.

`bufLen` is there to check that `[]float64` slice won't translate into a non-addressable `[]byte` slice (also taking into account the number of rows/columns `int64` descriptors)

handling this error mode at the `(Vector|Dense).MarshalBinary` level is the right place.
and as we *know* what the final slice size ought to be, it's good practice to tell `bytes.Buffer` instead of letting it dynamically grow (which could load the garbage allocator and severely impair performances)

here is what my code currently looks like:

```go
func (v Vector) MarshalBinary() ([]byte, error) {
	bufLen := int64(sizeInt64) + int64(v.n)*int64(sizeFloat64)
	if bufLen <= 0 {
		// bufLen is too big and has wrapped around.
		return nil, errTooBig
	}
// ...
```

kortschak: I'm just going to sit in the corner for a while. I think I should take a break. Sorry.

On 07/03/2016, at 8:58 PM, "Vladimír Chalupecký" <notifications@github.com<mailto:notifications@github.com>> wrote:

But here were are creating the payload and I hope that we can trust the Dense receiver upon which MarshalBinary is being called.

mat64/io_test.go
sbinet: done.
I actually went the extra mile and incorporated the NaN/Inf checks into a global test-case suit (the uglyness was somewhat limited)

sbinet:
PTAL.
modulo the infamous `bufLen` check, I believe I have addressed all concerns/issues.

vladimir-ch:
Thanks for the PR and for bearing with me :-) LGTM!

jonlawlor:
Yes, thanks for going through the gauntlet!  After this maybe we can fuzz the matrix library...

btracey:
I have the start for a fuzzer for blas. It found a few bugs in just the
level one, so I'm sure there's more to find.

sbinet:
ok, 1 explicit LGTM and 1.5 implicit LGTM.
as this has been quit a long winded PR, I think I'd wait for a tally of at least 2 explicit LGTMs :)

kortschak:
I'd change the text of errBadSize to something like ��mat64: invalid dimension" to avoid making a claim that may not be true about cause. After that...

LGTM

PR #345, opened 2016-03-08
mat64: implement binary streamers

### LINKED ISSUE #344: mat64: implement a stream-oriented binary (de)serialization
the whole discussion about #342 and the `[]byte` size implications made me think about implementing a more memory-wise efficient approach.

why not also provide:

```go
// pseudo-code
func (d *Dense) MarshalBinaryStream(w io.Writer) error {
      binary.Write(w, littleEndian, nrows)
      binary.Write(w, littleEndian, ncols)
      for i := range nrows {
        for j := range ncols {
           binary.Write(w, littleEndian, data.at(i,j))
        }
      }
      return err
}

func (d *Dense) UnmarshalBinaryStream(r io.Reader) error {
   // ...
}
```

and then implement `MarshalBinary` in terms of `MarshalBinaryStream`.

people not wanting to get beaten by the possibly large temporary `[]byte` slice could just use the stream oriented version.

sbinet:
Introduce:
- MarshalBinaryTo(w io.Writer) (int, error) and
- UnmarshalBinaryFrom(r io.Reader) (int, error)

for Dense and Vector.
Re-wire MarshalBinary and UnmarshalBinary to use the above, with slight
modifications.

Fixes #344.

### DIFF
--- a/mat64/io.go
+++ b/mat64/io.go
@@ -7,6 +7,7 @@ package mat64
 import (
 	"encoding/binary"
 	"errors"
+	"io"
 	"math"
 )
 
@@ -60,6 +61,41 @@ func (m Dense) MarshalBinary() ([]byte, error) {
 	return buf, nil
 }
 
+// MarshalBinaryTo encodes the receiver into a binary form and writes it into w.
+// MarshalBinaryTo returns the number of bytes written into w and an error, if any.
+//
+// See MarshalBinary for the on-disk layout.
+func (m Dense) MarshalBinaryTo(w io.Writer) (int, error) {
+	var n int
+	var buf [8]byte
+	binary.LittleEndian.PutUint64(buf[:], uint64(m.mat.Rows))
+	nn, err := w.Write(buf[:])
+	n += nn
+	if err != nil {
+		return n, err
+	}
+	binary.LittleEndian.PutUint64(buf[:], uint64(m.mat.Cols))
+	nn, err = w.Write(buf[:])
+	n += nn
+	if err != nil {
+		return n, err
+	}
+
+	r, c := m.Dims()
+	for i := 0; i < r; i++ {
+		for j := 0; j < c; j++ {
+			binary.LittleEndian.PutUint64(buf[:], math.Float64bits(m.at(i, j)))
+			nn, err = w.Write(buf[:])
+			n += nn
+			if err != nil {
+				return n, err
+			}
+		}
+	}
+
+	return n, nil
+}
+
 // UnmarshalBinary decodes the binary form into the receiver.
 // It panics if the receiver is a non-zero Dense matrix.
 //
@@ -99,19 +135,70 @@ func (m *Dense) UnmarshalBinary(data []byte) error {
 		return errBadBuffer
 	}
 
-	m.mat.Rows = int(rows)
-	m.mat.Cols = int(cols)
-	m.mat.Stride = int(cols)
-	m.capRows = int(rows)
-	m.capCols = int(cols)
-	m.mat.Data = use(m.mat.Data, int(size))
-
+	m.reuseAs(int(rows), int(cols))
 	for i := range m.mat.Data {
 		m.mat.Data[i] = math.Float64frombits(binary.LittleEndian.Uint64(data[p : p+sizeFloat64]))
 		p += sizeFloat64
 	}
 
 	return nil
+}
+
+// UnmarshalBinaryFrom decodes the binary form into the receiver and returns
+// the number of bytes read and an error if any.
+// It panics if the receiver is a non-zero Dense matrix.
+//
+// See MarshalBinary for the on-disk layout.
+//
+// Limited checks on the validity of the binary input are performed:
+//  - matrix.ErrShape is returned if the number of rows or columns is negative,
+//  - an error is returned if the resulting Dense matrix is too
+//  big for the current architecture (e.g. a 16GB matrix written by a
+//  64b application and read back from a 32b application.)
+// UnmarshalBinary does not limit the size of the unmarshaled matrix, and so
+// it should not be used on untrusted data.
+func (m *Dense) UnmarshalBinaryFrom(r io.Reader) (int, error) {
+	if !m.isZero() {
+		panic("mat64: unmarshal into non-zero matrix")
+	}
+
+	var (
+		n   int
+		buf [8]byte
+	)
+	nn, err := readFull(r, buf[:])
+	n += nn
+	if err != nil {
+		return n, err
+	}
+	rows := int64(binary.LittleEndian.Uint64(buf[:]))
+
+	nn, err = readFull(r, buf[:])
+	n += nn
+	if err != nil {
+		return n, err
+	}
+	cols := int64(binary.LittleEndian.Uint64(buf[:]))
+	if rows < 0 || cols < 0 {
+		return n, errBadSize
+	}
+
+	size := rows * cols
+	if int(size) < 0 || size > maxLen {
+		return n, errTooBig
+	}
+
+	m.reuseAs(int(rows), int(cols))
+	for i := range m.mat.Data {
+		nn, err = readFull(r, buf[:])
+		n += nn
+		if err != nil {
+			return n, err
+		}
+		m.mat.Data[i] = math.Float64frombits(binary.LittleEndian.Uint64(buf[:]))
+	}
+
+	return n, nil
 }
 
 // MarshalBinary encodes the receiver into a binary form and returns the result.
@@ -137,6 +224,35 @@ func (v Vector) MarshalBinary() ([]byte, error) {
 	}
 
 	return buf, nil
+}
+
+// MarshalBinaryTo encodes the receiver into a binary form, writes it to w and
+// returns the number of bytes written and an error if any.
+//
+// See MarshalBainry for the on-disk format.
+func (v Vector) MarshalBinaryTo(w io.Writer) (int, error) {
+	var (
+		n   int
+		buf [8]byte
+	)
+
+	binary.LittleEndian.PutUint64(buf[:], uint64(v.n))
+	nn, err := w.Write(buf[:])
+	n += nn
+	if err != nil {
+		return n, err
+	}
+
+	for i := 0; i < v.n; i++ {
+		binary.LittleEndian.PutUint64(buf[:], math.Float64bits(v.at(i)))
+		nn, err = w.Write(buf[:])
+		n += nn
+		if err != nil {
+			return n, err
+		}
+	}
+
+	return n, nil
 }
 
 // UnmarshalBinary decodes the binary form into the receiver.
@@ -169,13 +285,76 @@ func (v *Vector) UnmarshalBinary(data []byte) error {
 		return errBadBuffer
 	}
 
-	v.n = int(n)
-	v.mat.Inc = 1
-	v.mat.Data = use(v.mat.Data, v.n)
+	v.reuseAs(int(n))
 	for i := range v.mat.Data {
 		v.mat.Data[i] = math.Float64frombits(binary.LittleEndian.Uint64(data[p : p+sizeFloat64]))
 		p += sizeFloat64
 	}
 
 	return nil
+}
+
+// UnmarshalBinaryFrom decodes the binary form into the receiver, from the
+// io.Reader and returns the number of bytes read and an error if any.
+// It panics if the receiver is a non-zero Vector.
+//
+// See MarshalBinary for the on-disk layout.
+// See UnmarshalBinary for the list of sanity checks performed on the input.
+func (v *Vector) UnmarshalBinaryFrom(r io.Reader) (int, error) {
+	if !v.isZero() {
+		panic("mat64: unmarshal into non-zero vector")
+	}
+
+	var (
+		n   int
+		buf [8]byte
+	)
+	nn, err := readFull(r, buf[:])
+	n += nn
+	if err != nil {
+		return n, err
+	}
+	sz := int64(binary.LittleEndian.Uint64(buf[:]))
+	if sz < 0 {
+		return n, errBadSize
+	}
+	if sz > maxLen {
+		return n, errTooBig
+	}
+
+	v.reuseAs(int(sz))
+	for i := range v.mat.Data {
+		nn, err = readFull(r, buf[:])
+		n += nn
+		if err != nil {
+			return n, err
+		}
+		v.mat.Data[i] = math.Float64frombits(binary.LittleEndian.Uint64(buf[:]))
+	}
+
+	if n != sizeInt64+int(sz)*sizeFloat64 {
+		return n, io.ErrUnexpectedEOF
+	}
+
+	return n, nil
+}
+
+// readFull reads from r into buf until it has read len(buf).
+// It returns the number of bytes copied and an error if fewer bytes were read.
+// If an EOF happens after reading fewer than len(buf) bytes, io.ErrUnexpectedEOF is returned.
+func readFull(r io.Reader, buf []byte) (int, error) {
+	var n int
+	var err error
+	for n < len(buf) && err == nil {
+		var nn int
+		nn, err = r.Read(buf[n:])
+		n += nn
+	}
+	if n == len(buf) {
+		return n, nil
+	}
+	if err == io.EOF {
+		return n, io.ErrUnexpectedEOF
+	}
+	return n, err
 }
--- a/mat64/io_test.go
+++ b/mat64/io_test.go
@@ -7,6 +7,8 @@ package mat64
 import (
 	"bytes"
 	"encoding"
+	"io"
+	"io/ioutil"
 	"math"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


sbinet:
just a first stab.

- need to document the new methods

also: should they return `(int, error)` or `(int64, error)` ?

`io.Reader.Read([]byte)` and `io.Writer.Write([]byte)` return `(int, error)`

but:
`io.ReaderFrom.ReadFrom(r io.Reader)` and `io.WriterTo.WriteTo(w io.Writer)` return `(int64, error)`

jonlawlor:
Then I would say `(int64, error)`

kortschak:
I'm not entirely convinced that int64 is needed; io.ReaderFrom and io.WriterTo are generic pass through mechanisms that may take their bytes from arbitrarily long streams. We are marshaling and unmarshaling values that are in memory and so must be able to fit in a slice with len < maxInt for the architecture (and must already be smaller than maxInt for the architecture when considering \sum(sizeof(parts))). If the integer value is larger than maxInt (say unmarshaling a matrix on 386 that was marshaled on amd64) then the result is a non-nil error and the number of bytes that were actually read.

jonlawlor:
I agree that it basically doesn't matter in practice.  I'm just in favor of using the same convention as the standard lib whenever possible, and I think the io.Reader{From|To} cases are sufficiently close to the proposed MarshalBinary{From|To}.

kortschak:
I don't think they are. The fundamental difference is that WriterTo and ReaderFrom deal with arbitrarily long streams; from thr ReaderFrom docs, "ReadFrom reads data from r until EOF or error." UnmarshelFrom adds an additional constraint - or until enough data have been read to build the matrix. Semantically they are quite different despite the similarity of signature.

jonlawlor:
Yes on second thought you are right.

mat64/io.go
```diff
@@ -82,28 +68,73 @@ func (m *Dense) UnmarshalBinary(data []byte) error {
...
+	n += sizeInt64
+	err = binary.Write(w, binary.LittleEndian, int64(m.mat.Cols))
+	if err != nil {
+		return n, err
+	}
+	n += sizeInt64
+	r, c := m.Dims()
+	for i := 0; i < r; i++ {
+		for j := 0; j < c; j++ {
+			err = binary.Write(w, binary.LittleEndian, m.at(i, j))
```
kortschak: You'll use `binary.LittleEndian.PutUint64(buf[:], math.Float64bits(m.At(i, j)))` and the read equivalent `m.Set(i, j, math.Float64frombits(binary.LittleEndian.Uint64(buf[:])))` below?

kortschak: Actually, for the read operation don't use set, just iterate across the Data slice since we know the rows are contiguous: `m.mat.Data[i] = math.Float64frombits(binary.LittleEndian.Uint64(buf[:]))`

mat64/io.go
```diff
@@ -162,25 +193,64 @@ func (v *Vector) UnmarshalBinary(data []byte) error {
 	}
 
 	buf := bytes.NewReader(data)
-	var n int64
-	err := binary.Read(buf, binary.LittleEndian, &n)
+	_, err := v.UnmarshalBinaryFrom(buf)
 	if err != nil {
 		return err
 	}
+	if len(data) != int(v.n)*sizeFloat64+sizeInt64 {
```
kortschak: Use n from the `v.UnmarshalBinaryFrom(buf)` call against `len(data)`.

mat64/io.go
```diff
@@ -82,28 +68,73 @@ func (m *Dense) UnmarshalBinary(data []byte) error {
...
-	var rows int64
-	err := binary.Read(buf, binary.LittleEndian, &rows)
+	n, err := m.UnmarshalBinaryFrom(buf)
 	if err != nil {
 		return err
 	}
+	if n != len(data) {
+		return errBadBuffer
+	}
+	if len(data) != len(m.mat.Data)*sizeFloat64+2*sizeInt64 {
```
kortschak: Why check both?

mat64/io.go
kortschak: When you do this, if you have an EOF before completion of the expected size, return an `io.UnexpectedEOF`.

sbinet:
I'll wait for #354 to land before coming back to this, just to make sure I won't introduce performance regressions.

sbinet:
PTAL.

went duplicating the streamers functions (it's not like the format will change anyways)
added tests comparing streamers-based and non-streamers-based functions
added tests for the error returned by the `UnmarshalBinaryFrom`

mat64/io.go
```diff
@@ -60,6 +62,42 @@ func (m Dense) MarshalBinary() ([]byte, error) {
 	return buf, nil
 }
 
+// MarshalBinaryTo encodes the receiver into a binary form and writes it into w.
+// MarshalBinaryTo returns the number of bytes written into w and an error, if any.
+//
+// See MarshalBinary for the on-disk layout.
+func (m Dense) MarshalBinaryTo(w io.Writer) (int, error) {
+	// please keep Dense.MarshalBinaryTo in sync with Dense.MarshalBinary
```
kortschak: Don't make a comment, just add a test that checks that the []byte-based and streaming methods give the same result - marshal to a bytes buffer and to a [] and round-trip that to a dense/vector via a bytes reader and the []byte.

All locations below as well.

kortschak:
The code looks OK, but we should have a test to guarantee congruity.

I'm also wondering if there should be a type mark at the beginning of the stored bytes.

mat64/io.go
sbinet: done

mat64/io.go
```diff
@@ -60,6 +61,41 @@ func (m Dense) MarshalBinary() ([]byte, error) {
...
+// See MarshalBinary for the on-disk layout.
+func (m Dense) MarshalBinaryTo(w io.Writer) (int, error) {
+	var n int
+	var buf [8]byte
+	binary.LittleEndian.PutUint64(buf[:], uint64(m.mat.Rows))
+	nn, err := w.Write(buf[:])
+	if err != nil {
+		return n, err
+	}
+	n += nn
```
kortschak: Move these n adjustments above the returns, nn may not be zero.

mat64/io.go
```diff
@@ -60,6 +61,41 @@ func (m Dense) MarshalBinary() ([]byte, error) {
...
+	binary.LittleEndian.PutUint64(buf[:], uint64(m.mat.Cols))
+	nn, err = w.Write(buf[:])
+	if err != nil {
+		return n, err
+	}
+	n += nn
```
kortschak: Here too.

mat64/io.go
```diff
@@ -60,6 +61,41 @@ func (m Dense) MarshalBinary() ([]byte, error) {
...
+
+	r, c := m.Dims()
+	for i := 0; i < r; i++ {
+		for j := 0; j < c; j++ {
+			binary.LittleEndian.PutUint64(buf[:], math.Float64bits(m.at(i, j)))
+			nn, err = w.Write(buf[:])
+			if err != nil {
+				return n, err
+			}
+			n += nn
```
kortschak: And here.

mat64/io.go
```diff
@@ -114,6 +150,69 @@ func (m *Dense) UnmarshalBinary(data []byte) error {
...
+
+	var (
+		n   int
+		buf [8]byte
+	)
+	nn, err := readFull(r, buf[:])
+	if err != nil {
+		return n, err
+	}
+	n += nn
```
kortschak: And here.

mat64/io.go
```diff
@@ -114,6 +150,69 @@ func (m *Dense) UnmarshalBinary(data []byte) error {
...
+	rows := int64(binary.LittleEndian.Uint64(buf[:]))
+
+	nn, err = readFull(r, buf[:])
+	if err != nil {
+		return n, err
+	}
+	n += nn
```
kortschak: And here.

mat64/io.go
```diff
@@ -139,6 +238,35 @@ func (v Vector) MarshalBinary() ([]byte, error) {
...
+		n   int
+		buf [8]byte
+	)
+
+	binary.LittleEndian.PutUint64(buf[:], uint64(v.n))
+	nn, err := w.Write(buf[:])
+	if err != nil {
+		return n, err
+	}
+	n += nn
```
kortschak: And here.

mat64/io.go
```diff
@@ -139,6 +238,35 @@ func (v Vector) MarshalBinary() ([]byte, error) {
...
+
+	for i := 0; i < v.n; i++ {
+		binary.LittleEndian.PutUint64(buf[:], math.Float64bits(v.at(i)))
+		nn, err = w.Write(buf[:])
+		if err != nil {
+			return n, err
+		}
+		n += nn
```
kortschak: And here.

mat64/io.go
kortschak: Same for all the reads.

sbinet:
done.

mat64/io_test.go
```diff
@@ -274,6 +487,28 @@ func TestVectorIORoundTrip(t *testing.T) {
...
+			t.Errorf("error encoding test #%d: %v\n", i, err)
+		}
+
+		if !bytes.Equal(buf, wbuf.Bytes()) {
+			t.Errorf("encoding via MarshalBinary and MarshalBinaryTo differ:\nwith-stream: %q\n  no-stream: %q\n",
+				i, wbuf.Bytes(), buf,
+			)
+		}
+
+		var wgot Vector
```
kortschak: You need to add UnmarshalBinary as well here and check the congruence.

mat64/io_test.go
```diff
@@ -259,6 +403,75 @@ func TestVectorUnmarshal(t *testing.T) {
...
+		},
+	} {
+		buf := bytes.NewReader(test.raw[tt.beg:tt.end])
+		var v Vector
+		_, err := v.UnmarshalBinaryFrom(buf)
+		if err != io.ErrUnexpectedEOF {
+			t.Errorf("test #%d: error decoding. got=%v. want=%v\n", i, err, io.ErrUnexpectedEOF)
+		}
+	}
+}
```
kortschak: blank below.

mat64/io_test.go
```diff
@@ -141,7 +237,29 @@ func TestDenseIORoundTrip(t *testing.T) {
...
+		}
+
+		if !bytes.Equal(buf, wbuf.Bytes()) {
+			t.Errorf("encoding via MarshalBinary and MarshalBinaryTo differ:\nwith-stream: %q\n  no-stream: %q\n",
+				i, wbuf.Bytes(), buf,
+			)
+		}
+
+		var wgot Dense
+		_, err = wgot.UnmarshalBinaryFrom(wbuf)
```
kortschak: Need a UnmarshalBinary as well and test for congruence.

mat64/io.go
```diff
@@ -114,6 +150,69 @@ func (m *Dense) UnmarshalBinary(data []byte) error {
...
+	if rows < 0 || cols < 0 {
+		return n, errBadSize
+	}
+
+	size := rows * cols
+	if int(size) < 0 || size > maxLen {
+		return n, errTooBig
+	}
+
+	m.mat.Rows = int(rows)
```
kortschak: Can this whole stanza be replaced with m.reuseAs(int(rows), int(cols))?

The same thing in UnmarshalBinary.

mat64/io_test.go
sbinet: do we need it?
we test all these properties:
buf == wbuf
wbuf -> wgot and wgot == test.want
buf->got and got == test.want

I don't really see what else could be tested.

kortschak: Sorry, I missed that since it already exists.

sbinet: no worries :)

mat64/io.go
sbinet: is the additional function call worth it?
`reuseAs` makes a few sanity checks that we either already do or know by construction that we don't need.

well, I did a quick bench for Dense w/ and w/o `reuseAs`:

```
$ benchstat no-reuseas.txt ./reuseas.txt 
name                       old time/op    new time/op    delta
UnmarshalDense10-4            262ns ± 1%     269ns ± 4%    ~             (p=0.206 n=5+5)
UnmarshalDense100-4          1.70µs ± 3%    1.71µs ± 8%    ~             (p=0.841 n=5+5)
UnmarshalDense1000-4         15.5µs ± 2%    15.1µs ± 2%  -2.65%          (p=0.016 n=5+5)
UnmarshalDense10000-4         139µs ± 2%     138µs ± 6%    ~             (p=0.222 n=5+5)
UnmarshalFromDense10-4        619ns ± 2%     622ns ± 1%    ~             (p=0.468 n=5+5)
UnmarshalFromDense100-4      4.49µs ± 1%    4.55µs ± 2%    ~             (p=0.095 n=5+5)
UnmarshalFromDense1000-4     42.7µs ± 3%    43.0µs ± 3%    ~             (p=0.841 n=5+5)
UnmarshalFromDense10000-4     419µs ± 3%     426µs ± 4%    ~             (p=0.310 n=5+5)

name                       old alloc/op   new alloc/op   delta
UnmarshalDense10-4            80.0B ± 0%     80.0B ± 0%    ~     (all samples are equal)
UnmarshalDense100-4            896B ± 0%      896B ± 0%    ~     (all samples are equal)
UnmarshalDense1000-4         8.19kB ± 0%    8.19kB ± 0%    ~     (all samples are equal)
UnmarshalDense10000-4        81.9kB ± 0%    81.9kB ± 0%    ~     (all samples are equal)
UnmarshalFromDense10-4        88.0B ± 0%     88.0B ± 0%    ~     (all samples are equal)
UnmarshalFromDense100-4        904B ± 0%      904B ± 0%    ~     (all samples are equal)
UnmarshalFromDense1000-4     8.20kB ± 0%    8.20kB ± 0%    ~     (all samples are equal)
UnmarshalFromDense10000-4    81.9kB ± 0%    81.9kB ± 0%    ~     (all samples are equal)

name                       old allocs/op  new allocs/op  delta
UnmarshalDense10-4             1.00 ± 0%      1.00 ± 0%    ~     (all samples are equal)
UnmarshalDense100-4            1.00 ± 0%      1.00 ± 0%    ~     (all samples are equal)
UnmarshalDense1000-4           1.00 ± 0%      1.00 ± 0%    ~     (all samples are equal)
UnmarshalDense10000-4          1.00 ± 0%      1.00 ± 0%    ~     (all samples are equal)
UnmarshalFromDense10-4         2.00 ± 0%      2.00 ± 0%    ~     (all samples are equal)
UnmarshalFromDense100-4        2.00 ± 0%      2.00 ± 0%    ~     (all samples are equal)
UnmarshalFromDense1000-4       2.00 ± 0%      2.00 ± 0%    ~     (all samples are equal)
UnmarshalFromDense10000-4      2.00 ± 0%      2.00 ± 0%    ~     (all samples are equal)
```

it seems the extra function call isn't impacting too much.
it's definitely in the noise.

kortschak: It makes the code here cleaner and means if there are additional field added to Dense we get them initialised here for free, and we get the cap sanity check. The cost is close to zero.

sbinet: ok.
perhaps do that in a different CL ?

sbinet: ok, done.

mat64/io.go
```diff
@@ -179,3 +293,71 @@ func (v *Vector) UnmarshalBinary(data []byte) error {
...
+	}
+	sz := int64(binary.LittleEndian.Uint64(buf[:]))
+	if sz < 0 {
+		return n, errBadSize
+	}
+	if sz > maxLen {
+		return n, errTooBig
+	}
+
+	v.reuseAs(int(n))
```
kortschak: This should be `v.reuseAs(int(sz))`.

sbinet: good catch. sorry about that.

mat64/io.go
```diff
@@ -169,9 +285,7 @@ func (v *Vector) UnmarshalBinary(data []byte) error {
 		return errBadBuffer
 	}
 
-	v.n = int(n)
-	v.mat.Inc = 1
-	v.mat.Data = use(v.mat.Data, v.n)
+	v.reuseAs(int(sz))
```
kortschak: Not here though.

sbinet:
*sigh*... I need holidays.

thanks.
done.

mat64/io.go
```diff
@@ -179,3 +293,71 @@ func (v *Vector) UnmarshalBinary(data []byte) error {
...
+	)
+	for n < max && err == nil {
+		var nn int
+		nn, err = r.Read(buf[n:])
+		n += nn
+	}
+	if n != max {
+		err = io.ErrUnexpectedEOF
+	}
+	if err == io.EOF {
```
kortschak: I don't believe this is correct. It is valid for an io.Reader to successfully complete a read, filling the provided byte slice and ending at the end of file to also return an io.EOF. That would mean here we get the expected number of bytes and an io.ErrUnexpectedEOF.

The correct code looks like this:

```
func readFull(r Reader, buf []byte) (int, error) {
	var n int
	var err error
	for n < len(buf) && err == nil {
		var nn int
		nn, err = r.Read(buf[n:])
		n += nn
	}
	if n == len(buf) {
		return n, nil
	}
	if n > 0 && err == io.EOF {
		return n, io.ErrUnexpectedEOF
	}
	return n, err
}
```

sbinet: done.

mat64/io.go
```diff
@@ -179,3 +293,71 @@ func (v *Vector) UnmarshalBinary(data []byte) error {
...
+	return n, nil
+}
+
+// readFull reads from r into buf until it has read len(buf).
+// It returns the number of bytes copied and an error if fewer bytes were read.
+// If an EOF happens after reading fewer than len(buf) bytes, io.ErrUnexpectedEOF is returned.
+func readFull(r io.Reader, buf []byte) (int, error) {
+	var (
+		n   int
+		max = len(buf)
```
kortschak: This isn't really necessary, but we can leave it.

kortschak:
LGTM

mat64/io.go
sbinet: fixed anyway.

Issue #336, opened 2016-02-19
mat64: extend aliasing detection to types other than RawMatrixer


kortschak:
Currently we perform alias detection for RawMatrixer types. We should cover:

- [x] RawSymmetricer
- [x] RawTriangular
- [x] RawVectorer
- [x] RawMatrixer against RawVectorer (special case).

We have alias detection code in shadow.go, but it needs to be wired in to the user-facing calls.

There is currently no direct way to do comparison for RawMatrixer against Vector, but a small helper to convert a Vector to a blas64.General will suffice.

kortschak:
A question that I don't remember the answer to, though I think it has been discussed. Why are we not checking overlap during Copy calls? I don't see any obvious reason why we don't.

btracey:
Depends on how much we want to match the builtin. The builtin copy is fine with overlap. Copy is the weird function in mat64, so it's not clear it must obey the rules of the other functions.

kortschak:
The builtin is fine with overlap because it has an understanding of direction (see [here](http://play.golang.org/p/J-H6SgvCHA)). We don't, because BLAS doesn't. We could, though then we'd need more unsafe offset support. At the moment it seems to me like it's being special-cased for no real reason except for similarity to copy.

kortschak:
Fixed by #347, #349, #350.

PR #347, opened 2016-03-09
mat64: add support for vector shadowing detection

kortschak:
Updates #336.

### DIFF
--- a/mat64/lq.go
+++ b/mat64/lq.go
@@ -213,7 +213,5 @@ func (v *Vector) SolveLQVec(lq *LQ, trans bool, b *Vector) error {
 	} else {
 		v.reuseAs(c)
 	}
-	m := vecAsDense(v)
-	bm := vecAsDense(b)
-	return m.SolveLQ(lq, trans, bm)
+	return v.asDense().SolveLQ(lq, trans, b.asDense())
 }
--- a/mat64/lq.go
+++ b/mat64/lq.go
@@ -202,6 +202,9 @@ func (m *Dense) SolveLQ(lq *LQ, trans bool, b Matrix) error {
 // SolveLQVec finds a minimum-norm solution to a system of linear equations.
 // Please see Dense.SolveLQ for the full documentation.
 func (v *Vector) SolveLQVec(lq *LQ, trans bool, b *Vector) error {
+	if v != b {
+		v.checkOverlap(b.mat)
+	}
 	r, c := lq.lq.Dims()
 	// The Solve implementation is non-trivial, so rather than duplicate the code,
 	// instead recast the Vectors as Dense and call the matrix code.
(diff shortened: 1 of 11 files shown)


kortschak:
I am reasonably sure I got all the Vector/Dense candidates, but please take extra looking for cases I've missed.

mat64/lq.go
```diff
@@ -210,7 +213,7 @@ func (v *Vector) SolveLQVec(lq *LQ, trans bool, b *Vector) error {
 	} else {
 		v.reuseAs(c)
 	}
-	m := vecAsDense(v)
-	bm := vecAsDense(b)
+	m := v.asDense()
```
kortschak: Considering moving these into the call line as `v.asDense().SolveLQ(lq, trans, b.asDense())`.

mat64/qr.go
```diff
@@ -209,21 +212,7 @@ func (v *Vector) SolveQRVec(qr *QR, trans bool, b *Vector) error {
 	} else {
 		v.reuseAs(c)
 	}
-	m := vecAsDense(v)
-	bm := vecAsDense(b)
+	m := v.asDense()
+	bm := b.asDense()
```
kortschak: Considering moving these into the call line as `return v.asDense().SolveQR(qr, trans, b.asDense())`.

mat64/solve.go
```diff
@@ -107,17 +107,20 @@ func (m *Dense) Solve(a, b Matrix) error {
...
 	v.reuseAs(c)
-	m := vecAsDense(v)
+	m := v.asDense()
 	// We conditionally create bm as m when b and v are identical
 	// to prevent the overlap detection code from identifying m
 	// and bm as overlapping but not identical.
 	bm := m
 	if v != b {
-		bm = vecAsDense(b)
+		bm = b.asDense()
```
kortschak: Not here because of conditionality.

vladimir-ch:
Not sure: check overlap with `lu.lu` in `LU.RankOne`?

kortschak:
I don't think it's possible for there to be an overlap there because the caller can't modify that or recover it. @btracey?

vladimir-ch:
Oh, sorry for the noise, it makes a copy of x and y into xs and ys.

vladimir-ch:
Your reason being the relevant one, of course.

vladimir-ch:
What about in `Dense.Mul`? There are checks when a and b are RawMatrixer but not when RawVectorer or *Vector.

kortschak:
Thanks.

kortschak:
Related to that, we don't actually use RawVectorer much - the concrete type is used in all cases. Why?

vladimir-ch:
I have wondered that myself before, I don't know. Maybe history, maybe because RawVector doesn't carry the length. In Mul it is not needed but at several places in matrix.go the `n` field is used (could be avoided though). I think we should use RawVectorer instead.

I don't see any other missed overlap checks and I'm fine with the one-liners. LGTM but needless to say, wait for @btracey .

kortschak:
I think my question was more about why RawVectorer exists.

btracey:
I agree that our position with Vector is weird. However, why shouldn't RawVectorer exist when all all of the other RawX types do?

btracey:
I agree that lu shouldn't need it since it is impossible for the user to access the data directly.

btracey:
LGTM

PR #349, opened 2016-03-10
mat64: add support for TriDense shadowing detection

kortschak:
This only happens in one place. It seems to me that it would make more sense for the check to be done in TriDense.Copy, but we're not doing that for reasons that I don't recall (possibly because copy doesn't panic?).


kortschak:
Updates #336.

vladimir-ch:
LGTM

btracey:
LGTM

PR #352, opened 2016-03-10
Minor cleanup in mat64

vladimir-ch:


btracey:
LGTM

kortschak:
Would you rename rowView to rawRowView while you're here?

btracey:
Good idea

kortschak:
LGTM after that.

I think there is an opportunity to put bounds checks in rawRowView (in two versions - check and nocheck) for column width. This goes part of the way towards my slice bounds checking issue. The use of rawRowView can then be extended in use.

PR #353, opened 2016-03-11
mat64: remove stale TODO

kortschak:
@btracey / @vladimir-ch Please take a look.

### DIFF
--- a/mat64/dense.go
+++ b/mat64/dense.go
@@ -327,7 +327,8 @@ func (m *Dense) Reset() {
 }
 
 // Clone makes a copy of a into the receiver, overwriting the previous value of
-// the receiver. The clone operation does not make any restriction on shape.
+// the receiver. The clone operation does not make any restriction on shape and
+// will not cause shadowing.
 //
 // See the Cloner interface for more information.
 func (m *Dense) Clone(a Matrix) {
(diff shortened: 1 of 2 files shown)


vladimir-ch:
For my better understanding: why is it stale?

kortschak:
I need to think more about this.

kortschak:
OK. Done thinking. It's clear that we have decided that Clone is something that does not preserve the receiver's state in any way (for better or worse). The original design in my head when I started the shadowing code was to engineer safe work places when there was shadowing. On discussion with @btracey we decided that restriction was more sensible. The TODO here was written when overlap detection would have allowed Clone to work somewhere else in case of an overlap. That can't happen anymore.

I'll add some extra documentation to Cloner and Dense.Clone, and I also will add the *Vector path.

mat64/dense.go
```diff
@@ -360,6 +356,12 @@ func (m *Dense) Clone(a Matrix) {
 				copy(mat.Data[i*c:(i+1)*c], amat.Data[i*amat.Stride:i*amat.Stride+c])
 			}
 		}
+	case *Vector:
+		amat := aU.mat
+		mat.Data = make([]float64, aU.n)
```
vladimir-ch: One more question: the default case reuses the slice because there is no way (unlike with *Vector or RawMatrixer) that such matrix could shadow a Dense, right?

kortschak: It actually could if someone constructed it that way. A matrix that stores its elements in a blas64.General but does not have a RawMatrix method would allow this. Probably I should allocate a new slice there since I am now saying that Clone **will not** cause shadowing.

kortschak: Done.

vladimir-ch: Sounds reasonable and consistent, and reduces my doubt.

vladimir-ch:
LGTM

PR #355, opened 2016-03-14
mat64: removed unused labels

kortschak:
Identified by honnef.co/go/unused/cmd/unused.

Note that two other unused labels are present, but I thought I would wait for input on those:

- github.com/gonum/matrix/mat64/dense_test.go:18:6: asDense is unused (my blame)
- github.com/gonum/matrix/mat64/triangular.go:319:6: getBlasTriangular is unused (@btracey)


vladimir-ch:
LGTM

kortschak:
Waiting for @btracey, particularly wrt getBlasTriangular (but also wrt asDense).

btracey:
I was framed! Curse you commit history!

I don't understand what the asDense function was ever supposed to do, so if it's unused it's fine to delete it.

It looks like `getBlasTriangular` is now done with 
````
func (t *TriDense) InverseTri(a Triangular) error {
	n, _ := a.Triangle()
	t.reuseAs(a.Triangle())
	t.Copy(a)
````

The latter is a lot nicer, so seems fine to delete.

kortschak:
FWIW asDense was probably used in a range []func(*Dense) Matrix loop prior to to invention of list test.

Thanks. I'll delete both.

PR #360, opened 2016-03-17
mat64: fix shadow boundary case

### LINKED ISSUE #359: mat64: shadowing detection has false negatives
The following test fails:

```
func TestIssue359(t *testing.T) {
	for xi := 0; xi < 2; xi++ {
		for xj := 0; xj < 2; xj++ {
			for yi := 0; yi < 2; yi++ {
				for yj := 0; yj < 2; yj++ {
					a := NewDense(3, 3, []float64{
						1, 2, 3,
						4, 5, 6,
						7, 8, 9,
					})
					x := a.View(xi, xj, 2, 2).(*Dense)
					y := a.View(yi, yj, 2, 2).(*Dense)

					panicked, _ := panics(func() { x.checkOverlap(y.mat) })
					if !panicked {
						t.Errorf("expected panic for aliased with offsets x(%d,%d) y(%d,%d):\nx:\n%v\ny:\n%v",
							xi, xj, yi, yj, Formatted(x), Formatted(y),
						)
					}
				}
			}
		}
	}
}
```

It's pretty clear this is an off-by-one - which is mirrored in the test expectations that do exist.

```
$ go test -run 359
--- FAIL: TestIssue359 (0.00s)
	shadow_test.go:115: expected panic for aliased with offsets x(0,0) y(0,1):
		x:
		⎡1  2⎤
		⎣4  5⎦
		y:
		⎡2  3⎤
		⎣5  6⎦
	shadow_test.go:115: expected panic for aliased with offsets x(0,0) y(1,1):
		x:
		⎡1  2⎤
		⎣4  5⎦
		y:
		⎡5  6⎤
		⎣8  9⎦
	shadow_test.go:115: expected panic for aliased with offsets x(0,1) y(0,0):
		x:
		⎡2  3⎤
		⎣5  6⎦
		y:
		⎡1  2⎤
		⎣4  5⎦
	shadow_test.go:115: expected panic for aliased with offsets x(1,0) y(1,1):
		x:
		⎡4  5⎤
		⎣7  8⎦
		y:
		⎡5  6⎤
		⎣8  9⎦
	shadow_test.go:115: expected panic for aliased with offsets x(1,1) y(0,0):
		x:
		⎡5  6⎤
		⎣8  9⎦
		y:
		⎡1  2⎤
		⎣4  5⎦
	shadow_test.go:115: expected panic for aliased with offsets x(1,1) y(1,0):
		x:
		⎡5  6⎤
		⎣8  9⎦
		y:
		⎡4  5⎤
		⎣7  8⎦
FAIL
```

kortschak:
Fixes #359.

### DIFF
--- a/mat64/shadow.go
+++ b/mat64/shadow.go
@@ -191,15 +191,35 @@ func (v *Vector) checkOverlap(a blas64.Vector) bool {
 // rectanglesOverlap returns whether the strided rectangles a and b overlap
 // when b is offset by off elements after a but has at least one element before
 // the end of a. a and b have aCols and bCols respectively.
+//
+// rectanglesOverlap works by shifting both matrices left such that the left
+// column of a is at 0. The column indexes are flattened by obtaining the shifted
+// relative left and right column positions modulo the common stride. This allows
+// direct comparison of the column offsets when the matrix backing data slices
+// are known to overlap.
 func rectanglesOverlap(off, aCols, bCols, stride int) bool {
 	if stride == 1 {
+		// Unit stride means overlapping data
+		// slices must overlap as matrices.
 		return true
 	}
+
+	// Flatten the shifted matrix column positions
+	// so a starts at 0, modulo the common stride.
+	const aFrom = 0
 	aTo := aCols
+	// The mod stride operations here make the from
+	// and to indexes comparable between a and b when
+	// the data slices of a and b overlap.
 	bFrom := off % stride
 	bTo := (bFrom + bCols) % stride
-	if bFrom < bTo {
-		return aTo > bFrom
+
+	if bTo == 0 || bFrom < bTo {
+		// b matrix is not wrapped: compare for
+		// simple overlap.
+		return bFrom < aTo
 	}
-	return bTo > 0
+
+	// b strictly wraps and so must overlap with a.
+	return true
 }
--- a/mat64/shadow_test.go
+++ b/mat64/shadow_test.go
@@ -89,3 +89,29 @@ type interval struct{ from, to int }
 func intervalsOverlap(a, b interval) bool {
 	return a.to > b.from && b.to > a.from
 }
+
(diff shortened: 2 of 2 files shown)


btracey:
I don't understand. Could you elaborate on what the problem was? This seems to be subtler than I originally realized, so I think the best place to elaborate is in a function comment. Thanks.

kortschak:
Added commentary. PTAL

kortschak:
To be explicit here about the problem, previously rectanglesOverlap did not properly handle the case where the later matrix (b) has a right column that is at the far right of the strided backing slice. This meant it was handled by the wrapped case and caused the overlap to be missed. The comments in the code now should allow you to draw what things will look like (the only way to explain it properly is graphically, which I obviously can't do here).

*ping* also @vladimir-ch

mat64/shadow.go
```diff
@@ -191,15 +191,44 @@ func (v *Vector) checkOverlap(a blas64.Vector) bool {
...
+	// of b aligns with the stride of the backing
+	// data slice.
+	if bTo == 0 {
+		return bFrom < aTo
+	}
+
+	// Compare the case where the right column of
+	// b wraps around to the left again, modulo
+	// stride.
+	return bTo > aFrom
```
btracey: This is the same as `return true`, right? It's clearer to me to read something like "b wraps, so must overlap".

btracey:
LGTM, but I'd appreciate a glance from @vladimir-ch given I've already been wrong once about this snippet.

mat64/shadow.go
kortschak: Good find. Yes, it is always true. I'll change this.

kortschak:
Agreed.

vladimir-ch:
Just four ints and so difficult to reason about.
LGTM

kortschak:
Yes. The comments make it much clearer and @btracey 's observation has significantly simplified the test.

PR #361, opened 2016-03-17
mat64: make Copy direction-aware

### LINKED ISSUE #351: mat64: consider protecting Copy with shadow detection
See comment https://github.com/gonum/matrix/issues/336#issuecomment-194589199.

Either we detect the orientation of the copy (~~we can't guarantee this since the BLAS implementation may not be aware of this - and it's hard to get in pure Go~~) or we prohibit overlapping copies. At the moment we allow silent corruption.

kortschak:
Fixes #351.

~~Depends on #360.~~

~~Vector case not yet complete (nor tests for this): needs proper handling of transpose - requires thought.~~

### DIFF
--- a/mat64/dense.go
+++ b/mat64/dense.go
@@ -379,7 +379,9 @@ func (m *Dense) Clone(a Matrix) {
 
 // Copy makes a copy of elements of a into the receiver. It is similar to the
 // built-in copy; it copies as much as the overlap between the two matrices and
-// returns the number of rows and columns it copied.
+// returns the number of rows and columns it copied. If a aliases the receiver
+// and is a transposed Dense or Vector, with a non-unitary increment, Copy will
+// panic.
 //
 // See the Copier interface for more information.
 func (m *Dense) Copy(a Matrix) (r, c int) {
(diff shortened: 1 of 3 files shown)


kortschak:
This is ready for review. The main issue I think is documentation.

@btracey @vladimir-ch

mat64/dense.go
```diff
@@ -379,7 +379,9 @@ func (m *Dense) Clone(a Matrix) {
 
 // Copy makes a copy of elements of a into the receiver. It is similar to the
 // built-in copy; it copies as much as the overlap between the two matrices and
-// returns the number of rows and columns it copied.
+// returns the number of rows and columns it copied. If a aliases the receiver
+// and is a transposed Dense or Vector, with a non-unitary increment, Copy will
+// panic.
```
btracey: I realize that it can be done, but is there a particular case that makes copy with transpose and aliasing useful? The rule is more clear if it's "aliasing is okay except for transposes", and not "aliasing is okay except for transposes except when m has one dimension equal to one and hasn't been viewed". A function can't tell in general if the matrix has been viewed, and so it seems more prone to error to allowing certain special cases (i.e. this function worked fine until I sent in a viewed matrix)

kortschak: Initially I was going to prohibit aliased transpose copies completely, but then realised that this prevents doing row vector copies (where the row is copied from one part of a matrix to another that overlaps via a vector). This problem arises because vectors are column vectors and so thr copy must bee dont with v.T() as the source.

kortschak:
*ping*

mat64/dense.go
```diff
@@ -398,29 +400,57 @@ func (m *Dense) Copy(a Matrix) (r, c int) {
 	case RawMatrixer:
 		amat := aU.RawMatrix()
 		if trans {
+			if m.mat.Stride != 1 {
```
vladimir-ch: `amat`?

vladimir-ch:
I think I'll leave this for @btracey .

mat64/dense.go
btracey: Maybe we should explicity add a CopyRow then? If I wanted to copy a row from one part of a matrix to another, I would use a view. This does not require using a transpose. 
````
    r := a.View(r, 0, 1, cols)
    a.View(dst,0,1,cols).(*mat.Dense).Copy(r)
````
If I explicitly know it's a row, then I could call CopyRow (which can be added). There's overlap with SetRow, but they're slightly different.

File: mat64/dense.go [hunk 3]
```diff
@@ -398,29 +400,57 @@ func (m *Dense) Copy(a Matrix) (r, c int) {
...
 			for i := 0; i < r; i++ {
 				blas64.Copy(c,
 					blas64.Vector{Inc: amat.Stride, Data: amat.Data[i : i+(c-1)*amat.Stride+1]},
 					blas64.Vector{Inc: 1, Data: m.mat.Data[i*m.mat.Stride : i*m.mat.Stride+c]})
 			}
 		} else {
-			for i := 0; i < r; i++ {
-				copy(m.mat.Data[i*m.mat.Stride:i*m.mat.Stride+c], amat.Data[i*amat.Stride:i*amat.Stride+c])
+			switch o := offset(m.mat.Data, amat.Data); {
+			case o < 0:
```
btracey: Does this access ordering actually matter? I know it does for o == 0, but otherwise?

mat64/dense.go
```diff
@@ -398,29 +400,57 @@ func (m *Dense) Copy(a Matrix) (r, c int) {
...
 			stride = 1
 		} else {
 			n = r
 			stride = m.mat.Stride
 		}
-		amat := aU.mat
-		blas64.Copy(n,
-			blas64.Vector{Inc: amat.Inc, Data: amat.Data},
-			blas64.Vector{Inc: stride, Data: m.mat.Data})
+		if amat.Inc == 1 && stride == 1 {
```
btracey: m.mat.Inc == 1

mat64/dense.go
kortschak: Absolutely. http://play.golang.org/p/FXsd-exVVJ

mat64/dense.go
kortschak: I don't see the need to need for this proliferation. Copy does the work properly. An alternative approach which I also dislike, but less, is to rename Vector to Column and make another type Row that is defined `type Row Column` but satisfies Matrix.

The view approach you have above is fine for copying a row from one matrix to another, but doesn't deal with the case where you are building rows of a matrix from operations that produce a vector.

mat64/dense.go
kortschak: Done

mat64/dense.go
kortschak: Really the basis for this problem is that RowView returns a column.

mat64/dense.go
btracey: Ah, thanks.

mat64/dense.go
btracey: Thanks. Just wanted to be sure that we need the complexity here. Seems so.

mat64/dense.go
btracey: Yep, thanks.

btracey:
LGTM

PR #363, opened 2016-03-22
mat64: implement triangle shadow detection

### LINKED ISSUE #357: mat64: consider making symmetric and triangular overlap detection more precise
There is a couple of TODOs in shadow.go that suggest this. We need to understand whether it is worth doing. The intention of the TODO for each case is to only flag an overlap if the actual triangles in use overlap; this is consistent with the wording of the documentation (which would otherwise be changed), but is more work and more complicated code than appears in `rectanglesOverlap` (a brief survey suggests 6 cases instead of the current 2 - and I'm still not entirely sure of the best approach).

Is the cost of implementing this worth the effort?

kortschak:
Fixes #357.

Needs proper tests for triangle cases. Still thinking.

### DIFF
--- a/mat64/shadow.go
+++ b/mat64/shadow.go
@@ -4,7 +4,10 @@
 
 package mat64
 
-import "github.com/gonum/blas/blas64"
+import (
+	"github.com/gonum/blas"
+	"github.com/gonum/blas/blas64"
+)
 
 const (
 	// regionOverlap is the panic string used for the general case
@@ -66,10 +69,6 @@ func (m *Dense) checkOverlap(a blas64.General) bool {
 	return false
 }
 
-// BUG(kortschak): Overlap detection for symmetric and triangular matrices is not
-// precise; currently overlap is detected if the bounding rectangles overlap rather
-// than exact overlap between visible elements.
-
 func (s *SymDense) checkOverlap(a blas64.Symmetric) bool {
 	mat := s.RawSymmetric()
 	if cap(mat.Data) == 0 || cap(a.Data) == 0 {
@@ -100,12 +99,15 @@ func (s *SymDense) checkOverlap(a blas64.Symmetric) bool {
 		panic(mismatchedStrides)
 	}
 
-	// TODO(kortschak) Make this analysis more precise.
-	if off > 0 {
+	if off < 0 {
 		off = -off
 		mat.N, a.N = a.N, mat.N
+		// If we created the matrix it will always
+		// be in the upper triangle, but don't trust
+		// that this is the case.
+		mat.Uplo, a.Uplo = a.Uplo, mat.Uplo
 	}
-	if rectanglesOverlap(off, mat.N, a.N, mat.Stride) {
+	if trianglesOverlap(off, mat.N, a.N, mat.Stride, mat.Uplo == blas.Upper, a.Uplo == blas.Upper) {
 		panic(regionOverlap)
 	}
 	return false
@@ -141,12 +143,12 @@ func (t *TriDense) checkOverlap(a blas64.Triangular) bool {
 		panic(mismatchedStrides)
 	}
 
-	// TODO(kortschak) Make this analysis more precise.
-	if off > 0 {
+	if off < 0 {
 		off = -off
 		mat.N, a.N = a.N, mat.N
+		mat.Uplo, a.Uplo = a.Uplo, mat.Uplo
 	}
-	if rectanglesOverlap(off, mat.N, a.N, mat.Stride) {
+	if trianglesOverlap(off, mat.N, a.N, mat.Stride, mat.Uplo == blas.Upper, a.Uplo == blas.Upper) {
 		panic(regionOverlap)
 	}
 	return false
@@ -190,7 +192,7 @@ func (v *Vector) checkOverlap(a blas64.Vector) bool {
 
 // rectanglesOverlap returns whether the strided rectangles a and b overlap
 // when b is offset by off elements after a but has at least one element before
-// the end of a. a and b have aCols and bCols respectively.
+// the end of a. off must be positive. a and b have aCols and bCols respectively.
 //
 // rectanglesOverlap works by shifting both matrices left such that the left
 // column of a is at 0. The column indexes are flattened by obtaining the shifted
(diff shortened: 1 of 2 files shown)


mat64/shadow.go
```diff
@@ -223,3 +220,55 @@ func rectanglesOverlap(off, aCols, bCols, stride int) bool {
...
+		proj := off - rows*(stride+1)
+		if proj >= 0 && proj < aSize {
+			// Top left corner of b projects to element in top
+			// row of a.
+			return true
+		}
+		if bUpper {
+			// Project from upper right corner of b.
+			proj += bSize
+		} else {
```
vladimir-ch: Can a lower triangle overlap upper triangle if the upper left corner of the lower one lies outside of the upper one?

kortschak: This depends whether the lower triangle is a or b. If it's a (the data with the lower address), yes because all corners may be outside of the upper triangl. If it's b, no.

kortschak: So in this case, I think not.

kortschak: Done.

mat64/shadow.go
```diff
@@ -223,3 +220,62 @@ func rectanglesOverlap(off, aCols, bCols, stride int) bool {
...
+	// left corner, and upper right or lower left corner onto
+	// the left column of a.
+	cols := off % stride
+	proj := off - cols*(stride+1)
+	if proj >= 0 && proj < aSize*stride {
+		// Top left corner of b projects to element in left
+		// column of a.
+		return true
+	}
+	if bUpper {
```
vladimir-ch: I think that the situation here is the same, a is lower triangle, b is upper triangle, so there can't be overlap if the top left corner is not in a.

kortschak: I don't think so here.

```
     a
     aa
     a a
     a  a
     a   a
     a    a
  bbbbbbb  a
   b a  b   a
    baaaaaaaaa
     b  b
      b b
       bb
        b
```

vladimir-ch:
The notion of projecting the corners on the top row (or left column) of a is making the reasoning more difficult, at least for me. I'm not sure about this but, taking the situation aUpper and bUpper as an example, wouldn't testing the upper right corner of b against the diagonal of a and the upper left corner against the right column of a on a _given_ row of a work equally well? In other words, translate the top corners of b into the index space of a and test it there. For me it would make the test clearer.

Thinking about it, since you check overlap of the rectangles, in the above situation we could even skip testing the upper left corner of b. Or skip the rectangle overlap, and just check the two corners.

Isn't there a catch that I might be missing?

mat64/shadow.go
vladimir-ch: Right, sorry.

kortschak:
>The notion of projecting the corners on the top row (or left column) of a is making the reasoning more difficult, at least for me. I'm not sure about this but, taking the situation aUpper and bUpper as an example, wouldn't testing the upper right corner of b against the diagonal of a and the upper left corner against the right column of a on a given row of a work equally well? In other words, translate the top corners of b into the index space of a and test it there. For me it would make the test clearer.

I find this very difficult to reason about. Can you show me what you're saying in code?

>Thinking about it, since you check overlap of the rectangles, in the above situation we could even skip testing the upper left corner of b. Or skip the rectangle overlap, and just check the two corners.

That is true. This is the converse of the other case. Nice.

kortschak:
>That is true. This is the converse of the other case. Nice.

Sorry. No.

```
 a
 aa
 a a
 a  a  bbbbbbbbbb
 a   a  b       b
 a    a  b      b
 a     a  b     b
 a      a  b    b
 aaaaaaaaa  b   b
             b  b
              b b
               bb
                b
```

vladimir-ch:
> Can you show me what you're saying in code?

I'll try to but I think that in the end it will boil down to a similar set of checks.

>>That is true. This is the converse of the other case. Nice.

> Sorry. No.

In the comment I was considering just the case of aUpper to simplify the reasoning. At any case, don't take me too seriously.

kortschak:
>In the comment I was considering just the case of aUpper to simplify the reasoning.

Ahh, sorry. Yes, I think you are right.

kortschak:
Though I'm not sure it simplifies anything.

vladimir-ch:
Ok, no code is necessary. We are not thinking the same thing but almost and my concept does not simplify anything. But maybe we can remove the rectangles overlap check without _complicating_ anything. I'll comment in the code.

mat64/shadow.go
```diff
@@ -223,3 +220,62 @@ func rectanglesOverlap(off, aCols, bCols, stride int) bool {
...
+		return false
+	}
+
+	if aUpper {
+		// Test overlap by projecting diagonals from the b upper
+		// left corner, and upper right or lower left corner onto
+		// the top row of a.
+		rows := off / stride
+		proj := off - rows*(stride+1)
+		if proj >= 0 && proj < aSize {
```
vladimir-ch: If there is no `rectanglesOverlap` then this condition will be true in this situation:
```
aaa.
.aab
..a.
```
because
`aSize = 3`, `stride = 4`, `off = 7`.
Then
`rows = 1`, `proj = 7 - 1*5 = 2`.
And `2 >= 0 && 2 < 3` is true.
But `proj` in fact is the distance from a's diagonal. So if the condition is changed to
`if proj >= 0 && proj < aSize - rows`
it will be correctly false even without rectanglesOverlap.

Maybe something similar would work also in the !aUpper case but I haven't considered it.

I hope I'm not complicating the matter unnecessarily but for me it's a good way to understand this better. And getting the notion of projection to the top row away would simplify the reasoning for me.

kortschak: I'll have to think about this.

vladimir-ch: This is what I have in mind:
```go
func trianglesOverlap(off, aSize, bSize, stride int, aUpper, bUpper bool) bool {
	if aUpper {
		row := off / stride
		d := off - row*(stride+1) // Signed number of columns (modulo wrapping) between
		                          // b's upper left corner and a's diagonal.
		if d >= 0 && d < aSize - row {
			return true // b's upper left corner lies inside a.
		}
		d += bSize // Shift to b's upper right corner.
		if bUpper && d >= 0 && d < aSize - row {
			return true // b's upper right corner lies inside a.
		}
		return false
	}
	// !aUpper is more complicated.
}
```
Please be skeptical, it might be wrong. And not worth it.

vladimir-ch:
Being a true pioneer of dead ends, I've come to the conclusion that projecting onto the first row or column makes more sense, especially because of the `!aUpper` case. Sorry for the noise :-)

kortschak:
@vladimir-ch I have complete tests now. It's clear now that the conditions are not quite right, though the geometry is subtle. I'm going to leave it for a while, but if you feel like having a play and suggesting changes, please do.

kortschak:
The issue seems to be that I'm not properly accounting for the modular nature of the space I'm working in, but I've not figured out a good way of handling it.

kortschak:
The idea is to only work in mod space to get relative positions and then do all the comparisons in normal space. The stanza required to get offsets is as follows:

```
	// Find location of b relative to a.
	rowOffset := off / stride
	colOffset := off % stride
	if (off+bSize)%stride < colOffset {
		// We have wrapped, so readjust offsets.
		rowOffset++
		colOffset -= stride
	}
```

kortschak:
This works now. The conditions are inelegant and I'm sure that there are redundant checks, but this is reviewable.

PTAL

mat64/shadow.go
```diff
@@ -223,3 +225,55 @@ func rectanglesOverlap(off, aCols, bCols, stride int) bool {
...
+		// is in the triangle of a.
+		return bUpper && rowOffset < colOffset+bSize
+	}
+
+	// Check whether the upper left of b
+	// is in the triangle of a
+	if colOffset >= 0 && rowOffset >= colOffset {
+		return true
+	}
+	if !bUpper {
```
vladimir-ch: If you tested here `if bUpper` instead, there could be one indentation level less. Just a cosmetic detail.

I'm wondering if the following arrangement (and similar) is being tested somewhere:
```
a
aa
a.a b
a..abb
a...b.b
aaaab..b
    bbbbb
```

kortschak: I'll do that change.

Yes, that conformation is tested (lower left corner of b in a), though the comment doesn't say so. I'll fix that.

kortschak:
This fails on tip. I'll bisect that tomorrow.

kortschak:
Failure issue filed: https://github.com/golang/go/issues/15042.

mat64/shadow.go
```diff
@@ -223,3 +225,56 @@ func rectanglesOverlap(off, aCols, bCols, stride int) bool {
 	// b strictly wraps and so must overlap with a.
 	return true
 }
+
+// trianglesOverlap returns whether the strided triangles a and b overlap
+// when b is offset by off elements after a but has at least one element before
+// the end of a. a and b have aSize and bSize respectively.
```
btracey: This needs a note that "off" must be positive, right?

btracey: Could you say that they're aSize x aSize, it's not clear in this context that aSize is a dimension and not a total number of elements.

mat64/shadow_test.go
```diff
@@ -115,3 +258,27 @@ func TestIssue359(t *testing.T) {
 		}
 	}
 }
+
+// denseAsTriDense returns a triangular matrix derived from the
+// square matrix m, with the orientation specified by kind.
+func denseAsTriDense(m *Dense, kind matrix.TriKind) *TriDense {
```
btracey: Now who's being tricky :-)

btracey:
This looks correct to me. I'll let @vladimir-ch give the official goahead

mat64/shadow.go
kortschak: Done.

mat64/shadow_test.go
kortschak: I'm leading the way in bad behaviour because that's what we are trying to catch.

vladimir-ch:
LGTM

PR #367, opened 2016-04-08
fix issue #313 by increasing epsilon

idavydov:
In some cases eigen hangs.
https://gist.github.com/idavydov/eb4c88b4663a7240c92f
It works ok with epsilon=1e-15 works, 1e-14 to be safe, as this code is going to be replaced in future.


btracey:
Lgtm. If only dgeev wasn't so much work...

btracey:
Look good to you as well @kortschak ?

kortschak:
Yes, there was not problem here for me, I just like to leave the finer points of numerical things to you since you have very much more experience.

LGTM

PR #369, opened 2016-04-14
mat64: use lapack64.Ormqr in Dense.QFromQR

vladimir-ch:


mat64/qr.go
```diff
@@ -86,7 +86,7 @@ func (m *Dense) RFromQR(qr *QR) {
 
 // QFromQR extracts the m×m orthonormal matrix Q from a QR decomposition.
 func (m *Dense) QFromQR(qr *QR) {
-	r, c := qr.qr.Dims()
+	r, _ := qr.qr.Dims()
 	m.reuseAs(r, r)
```
kortschak: I think the elements of m need to be zeroed for this use.

vladimir-ch: It is, on lines 93-97.

kortschak: Ah yes, hidden in the fold. As you'll have noticed, Exp uses an initial I as well and it's probably worth sharing the reuseAsZeroed between them.

kortschak: I'll send that after this goes in.

kortschak:
LGTM

vladimir-ch:
I'll wait for @btracey .

btracey:
Woo lapack!

LGTM

PR #371, opened 2016-04-14
mat64: add reuseAsZeroed and use in Exp

kortschak:
~~WIP: Waiting to add use in QFromQR.~~

### DIFF
--- a/mat64/dense.go
+++ b/mat64/dense.go
@@ -62,6 +62,8 @@ func NewDense(r, c int, mat []float64) *Dense {
 
 // reuseAs resizes an empty matrix to a r×c matrix,
 // or checks that a non-empty matrix is r×c.
+//
+// reuseAs must be kept in sync with reuseAsZeroed.
 func (m *Dense) reuseAs(r, c int) {
 	if m.mat.Rows > m.capRows || m.mat.Cols > m.capCols {
 		// Panic as a string, not a mat64.Error.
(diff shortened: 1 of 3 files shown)


btracey:
I guess this is okay, but I would be happier if there was less copying from reuseAs. I don't see a good way around it though.

LGTM

kortschak:
This PR reduces copying when complete (after the QFromQR use is in). There is the option to merge reuseAs and reuseAsZeroed by passing a bool, but I don't like the readability impact of that (which is why both use and useZeroed exist).

btracey:
Yea, agreed on both accounts. Maybe add a comment to reuseAs that if that implementation changes, reuseAsZeroed should change as well?

kortschak:
Will do.

kortschak:
OK, this is ready.

PTAL

mat64/dense.go
```diff
@@ -83,6 +85,35 @@ func (m *Dense) reuseAs(r, c int) {
...
+		}
+		m.capRows = r
+		m.capCols = c
+		return
+	}
+	if r != m.mat.Rows || c != m.mat.Cols {
+		panic(matrix.ErrShape)
+	}
+	for i := 0; i < r; i++ {
+		zero(m.mat.Data[i*m.mat.Stride : i*m.mat.Stride+r])
```
btracey: i*m.mat.Stride + c, right?

btracey:
LGTM with fix.

mat64/dense.go
kortschak: Indeed.

kortschak: Interestingly it doesn't show in tests because both uses are for creating I, so r == c.

Issue #374, opened 2016-04-15
mat64: should Dense.SolveCholesky/Vec really return an error?


vladimir-ch:
If a matrix is positive definite, its Cholesky decomposition exists and the matrix is non-singular.
Cholesky.Factorize returns a bool that indicates that. Dense.SolveCholesky returns an error if the matrix is ill-conditioned, but 1) the decomposition exists so it is not strictly speaking an error to use for a solve, only the result will be useless 2) it makes it seem that the error somehow depends on the combination of a and b which is not true. What is the reasoning behind the current design?

btracey:
Yes it should. The cholesky factorization only fails if the matrix is exactly singular. This may not happen even if the matrix is exactly singular given floating point error. This is different for linear solve where it may not be exactly singular, but dangerously close. Secondly, the future may add functions like rank one to update the factorization without going through the original matrix. This could cause the matrix to become approximately singular

vladimir-ch:
Thanks, makes sense especially with respect to the updates. Speaking about updates, I have Cholesky.RankOne almost finished which is also the reason why I'm looking at Cholesky. I hope to submit sometime next week.

PR #376, opened 2016-04-19
mat64: add Cholesky.SymRankOne

vladimir-ch:


mat64/cholesky.go
```diff
@@ -178,3 +178,155 @@ func (s *SymDense) InverseCholesky(chol *Cholesky) error {
...
+	blas64.Scal(n, alpha, blas64.Vector{1, work})
+	// Solve U^T * p = x storing the result into work.
+	ok = lapack64.Trtrs(blas.Trans, c.chol.RawTriangular(), blas64.General{
+		Rows:   n,
+		Cols:   1,
+		Stride: 1,
+		Data:   work,
+	})
+	if !ok {
+		// The original matrix is singular.
```
vladimir-ch: We get here if the factorization has zeros on the diagonal, that is, when the matrix was not positive definite. In this situation Factorize should have returned false. It seems to me that we don't say clearly what the user should do in such situation. It is also slightly related to #374.

mat64/cholesky_test.go
```diff
@@ -210,6 +210,123 @@ func TestInverseCholesky(t *testing.T) {
...
+			x := NewVector(n, xdata)
+
+			var chol Cholesky
+			ok := chol.Factorize(&a)
+			if !ok {
+				t.Errorf("Bad random test, Cholesky factorization failed")
+				continue
+			}
+
+			alpha := rand.Float64()
```
vladimir-ch: Doing factorization on random matrices with random negative alpha is very unpredictable, so here the tests use only alpha > 0.

mat64/cholesky.go
```diff
@@ -178,3 +178,155 @@ func (s *SymDense) InverseCholesky(chol *Cholesky) error {
 	s.SymOuterK(1, &t)
 	return err
 }
+
+// RankOne performs a rank-1 update (or downdate when alpha is negative) of the
```
kortschak: Remove parenthetic; both are updates and "downdate" is not a word outside urbandictionary.

mat64/cholesky.go
```diff
@@ -178,3 +178,155 @@ func (s *SymDense) InverseCholesky(chol *Cholesky) error {
...
+// Cholesky factorization orig as if the update had been applied to the original
+// matrix A, storing the result into the reciever. That is, if in the original
+// Cholesky factorization
+//  U^T * U = A,
+// in the updated factorization
+//  U'^T * U' = A + alpha * x * x^T = A'.
+//
+// Note that when alpha is negative, the downdating problem may be
```
kortschak: s/downdating // or s/down/up/

mat64/cholesky.go
vladimir-ch: "Downdate/ing" sounds strange but after having studied the related literature for the past two weeks "downdate" is commonly used in this field to distinguish doing A - v v' from doing A + v v'. I will remove it here, splashing the word in the first line of the comment is not good.

kortschak: That's horrible. OK.

vladimir-ch: Done

mat64/cholesky.go
vladimir-ch: Down

kortschak: Nice word slip :)

mat64/cholesky.go
```diff
@@ -178,3 +178,154 @@ func (s *SymDense) InverseCholesky(chol *Cholesky) error {
 	s.SymOuterK(1, &t)
 	return err
 }
+
+// RankOne performs a rank-1 update of the Cholesky factorization orig as if the
```
btracey: original matrix A and refactorized

mat64/cholesky.go
```diff
@@ -178,3 +178,154 @@ func (s *SymDense) InverseCholesky(chol *Cholesky) error {
...
+//  U^T * U = A,
+// in the updated factorization
+//  U'^T * U' = A + alpha * x * x^T = A'.
+//
+// Note that when alpha is negative, the updating problem may be ill-conditioned
+// and the results may be inaccurate, or the updated matrix A' may not be
+// positive definite and not have a Cholesky factorization. RankOne returns
+// whether the updated matrix A' is positive definite.
+//
+// The utility of RankOne comes from the fact that updating the factorization is
```
btracey: Too wordy.

RankOne updates a Cholesky factorization in O(n^2) time. The Cholesky factorization computation from scratch is O(n^3).

mat64/cholesky.go
```diff
@@ -178,3 +178,154 @@ func (s *SymDense) InverseCholesky(chol *Cholesky) error {
...
+// an O(n²) algorithm while computing it from scratch is an O(n³) algorithm.
+func (c *Cholesky) RankOne(orig *Cholesky, alpha float64, x *Vector) (ok bool) {
```
btracey: This should be called SymRankOne. It's the equivalent of https://godoc.org/github.com/gonum/matrix/mat64#SymDense.SymRankOne

mat64/cholesky.go
```diff
@@ -178,3 +178,154 @@ func (s *SymDense) InverseCholesky(chol *Cholesky) error {
...
+	n := orig.Size()
+	if x.Len() != n {
+		panic(matrix.ErrShape)
+	}
+	if orig != c {
+		if c.chol == nil {
```
btracey: I know this isn't done in LU (it probably should be), but could you make an isZero() function? It's a lot clearer code-wise to see it that way.

mat64/cholesky.go
```diff
@@ -178,3 +178,154 @@ func (s *SymDense) InverseCholesky(chol *Cholesky) error {
...
+	// - Matthias Seeger: Low rank updates for the Cholesky decomposition.
+	//   EPFL Technical Report 161468 (2004)
+	//   http://infoscience.epfl.ch/record/161468
+
+	work := make([]float64, n)
+	blas64.Copy(n, x.RawVector(), blas64.Vector{1, work})
+
+	if alpha > 0 {
+		// Compute rank-1 update.
+
```
btracey: remove newline

mat64/cholesky.go
```diff
@@ -178,3 +178,154 @@ func (s *SymDense) InverseCholesky(chol *Cholesky) error {
...
+					blas64.Vector{1, work[i+1 : n]},
+					c, s)
+			}
+		}
+		c.updateCond(-1)
+		return true
+	}
+
+	// Compute rank-1 downdate.
+
```
btracey: remove newline

mat64/cholesky.go
btracey: If this case should never happen (because it implies a bug in Factorize), then maybe it should panic? It's not obvious in the general case what a user should do if the matrix is not positive definite. Get a new matrix? Add a term to the diagonal? This is why it's an error :).

mat64/cholesky.go
```diff
@@ -178,3 +178,154 @@ func (s *SymDense) InverseCholesky(chol *Cholesky) error {
...
+	if alpha > 0 {
+		// Compute rank-1 update.
+
+		blas64.Scal(n, math.Sqrt(alpha), blas64.Vector{1, work})
+		umat := c.chol.mat.Data
+		for i := 0; i < n; i++ {
+			// Compute parameters of the Givens matrix that zeroes
+			// the i-th element of x.
+			c, s, r, _ := blas64.Rotg(umat[i*n+i], work[i])
```
btracey: please use umat.Stride here instead of n.

mat64/cholesky.go
```diff
@@ -178,3 +178,154 @@ func (s *SymDense) InverseCholesky(chol *Cholesky) error {
...
+				c *= -1
+				s *= -1
+			}
+			umat[i*n+i] = r
+			if i < n-1 {
+				// Multiply the extended factorization matrix by
+				// the Givens matrix from the left. Only
+				// the i-th row and x are modified.
+				blas64.Rot(n-i-1,
+					blas64.Vector{1, umat[i*n+i+1 : i*n+n]},
```
btracey: same. umat.Stride.

mat64/cholesky.go
```diff
@@ -178,3 +178,154 @@ func (s *SymDense) InverseCholesky(chol *Cholesky) error {
...
+			norm *= -1
+			cos[i] *= -1
+			sin[i] *= -1
+		}
+	}
+	umat := c.chol.mat.Data
+	ok = true
+	for i := n - 1; i >= 0; i-- {
+		// Apply Givens matrices to U.
+		blas64.Rot(n-i, blas64.Vector{1, work[i:n]}, blas64.Vector{1, umat[i*n+i : i*n+n]}, cos[i], sin[i])
```
btracey: Stride

btracey:
Nice! I especially love non-trivial functionality that distinguishes us.

mat64/cholesky.go
vladimir-ch: Done. Please check whether I parsed your intention correctly.

mat64/cholesky.go
vladimir-ch: Thanks. Done.

mat64/cholesky.go
vladimir-ch: Done.

mat64/cholesky.go
vladimir-ch: Done.

mat64/cholesky.go
vladimir-ch: Done.

mat64/cholesky.go
vladimir-ch: Done.

mat64/cholesky.go
vladimir-ch: Done.

mat64/cholesky.go
```diff
@@ -178,3 +178,162 @@ func (s *SymDense) InverseCholesky(chol *Cholesky) error {
...
+		if c.isZero() {
+			c.chol = NewTriDense(n, matrix.Upper, nil)
+		} else if c.chol.mat.N != n {
+			panic(matrix.ErrShape)
+		}
+		c.chol.Copy(orig.chol)
+	}
+
+	if alpha == 0 {
+		return true
```
vladimir-ch: This is incorrect at the moment. When Factorize returns false, the original matrix is not positive definite while here by returning true we pretend it is. Again it boils down to how Cholesky should behave when Factorize returns false. Should we add a bool field to Cholesky indicating whether the matrix is PD and return it here? Or can we rely on cond being infinity? Should we panic and document it? The solve routines should behave similarly (now they return an error based on cond _after_ performing the solve).

mat64/cholesky.go
vladimir-ch: It can happen if the underlying matrix is not positive definite. I think Factorize is ok, it does not finish the factorization and returns a false. The question is how Cholesky should behave afterwards. The user tried to factorize, we found out the matrix is not PD, the factorization is left unfinished, and we tell him 'not PD'. He calls for example SolveCholesky anyway. At the moment it dutifully performs the solve even if the matrix is not PD which I think that will likely get a division by zero. This is why I asked if returning an error based on the condition number is the right thing to do. It seems that if Factorize returns false, it makes no sense to try to use Cholesky and we should be more explicit about it. If it returns true, then solving and returning an error based on the condition number does make sense.

I'm starting to think that the right behaviour would be a panic in SolveCholesky, SolveCholeskyVec, SymRankOne, LogDet, etc after Factorize returns false. Some or all of them may already panic with Log(0) or division by zero but panicking earlier with a more concrete message is desirable I guess.

btracey: https://github.com/gonum/matrix/issues/377

mat64/cholesky_example_test.go
```diff
@@ -68,3 +68,55 @@ func ExampleCholesky() {
...
+	// Rank-1 update the matrix a.
+	a.SymRankOne(a, 1, x)
+
+	var au mat64.SymDense
+	au.FromCholesky(&chol)
+
+	// Print the matrix that was updated directly.
+	fmt.Printf("\nA' =        %0.4v\n", mat64.Formatted(a, mat64.Prefix("            ")))
+	// Print the matrix recovered from the factorization.
+	fmt.Printf("U'^T * U' = %0.4v\n", mat64.Formatted(&au, mat64.Prefix("            ")))
```
btracey: Could you add a newline in here?

btracey:
LGTM

mat64/cholesky.go
```diff
@@ -178,3 +178,162 @@ func (s *SymDense) InverseCholesky(chol *Cholesky) error {
...
+			sin[i] *= -1
+		}
+	}
+	umat := c.chol.mat
+	stride := umat.Stride
+	ok = true
+	for i := n - 1; i >= 0; i-- {
+		// Apply Givens matrices to U.
+		blas64.Rot(n-i, blas64.Vector{1, work[i:n]}, blas64.Vector{1, umat.Data[i*stride+i : i*stride+n]}, cos[i], sin[i])
+		if umat.Data[i*stride+i] == 0 {
```
btracey: Actually, shouldn't this be <= 0? If it's less than zero then it's not positive definite.

mat64/cholesky_example_test.go
vladimir-ch: Done.

mat64/cholesky.go
vladimir-ch: I think not. The decisive check for whether the updated matrix is PD or not is done at line 293. What happens here is that cos and sin are determined so as to zero out the elements of p (work), the elements of U play no role in it. Therefore, when we apply them here to U, the particular combination of sin and cos can produce a negative diagonal element even if the updated matrix is PD. In other words, what this loop does is instead of computing the Cholesky factorization `U^T U = A` it computes a factorization `(DU)^T DU = A` where D is a diagonal matrix with some of its diagonal elements negative. What line 325 does is to undo the matrix D.

Of course, I think it may happen that the updated A is close to singular and that line 293 says that it is PD even if its not, but such case is already covered in the docs by "ill-conditioned" and "inaccurate".

PR #378, opened 2016-04-21
mat64: add Cholesky.Reset and panic on invalid Cholesky

vladimir-ch:
Addresses #377

### DIFF
--- a/mat64/cholesky.go
+++ b/mat64/cholesky.go
@@ -68,10 +68,18 @@ func (c *Cholesky) Factorize(a Symmetric) (ok bool) {
 	if ok {
 		c.updateCond(norm)
 	} else {
-		c.chol.Reset()
-		c.cond = math.Inf(1)
+		c.Reset()
 	}
 	return ok
+}
+
+// Reset resets the factorization so that it can be reused as the receiver of a
+// dimensionally restricted operation.
+func (c *Cholesky) Reset() {
+	if !c.isZero() {
+		c.chol.Reset()
+	}
+	c.cond = math.Inf(1)
 }
 
 // Size returns the dimension of the factorized matrix.
@@ -365,8 +373,7 @@ func (c *Cholesky) SymRankOne(orig *Cholesky, alpha float64, x *Vector) (ok bool
 	if ok {
 		c.updateCond(-1)
 	} else {
-		c.chol.Reset()
-		c.cond = math.Inf(1)
+		c.Reset()
 	}
 	return ok
 }
--- a/mat64/cholesky.go
+++ b/mat64/cholesky.go
@@ -21,6 +21,10 @@ const (
 
 // Cholesky is a type for creating and using the Cholesky factorization of a
 // symmetric positive definite matrix.
+//
+// Cholesky methods may only be called on a value that has been successfully
+// initialized by a call to Factorize that has returned true. Calls to methods
+// of an unsuccessful Cholesky factorization will panic.
 type Cholesky struct {
 	chol *TriDense
 	cond float64
--- a/mat64/cholesky.go
+++ b/mat64/cholesky.go
@@ -14,13 +14,19 @@ import (
 	"github.com/gonum/matrix"
 )
 
-const badTriangle = "mat64: invalid triangle"
+const (
+	badTriangle = "mat64: invalid triangle"
+	badCholesky = "mat64: invalid Cholesky factorization"
+)
 
 // Cholesky is a type for creating and using the Cholesky factorization of a
 // symmetric positive definite matrix.
 type Cholesky struct {
 	chol *TriDense
 	cond float64
+	// valid indicates whether A is positive definite
+	// and the factorization usable.
+	valid bool
 }
 
 // updateCond updates the condition number of the Cholesky decomposition. If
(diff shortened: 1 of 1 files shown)


mat64/cholesky.go
```diff
@@ -14,13 +14,19 @@ import (
...
 )
 
-const badTriangle = "mat64: invalid triangle"
+const (
+	badTriangle = "mat64: invalid triangle"
+	badCholesky = "mat64: invalid Cholesky factorization"
+)
 
 // Cholesky is a type for creating and using the Cholesky factorization of a
 // symmetric positive definite matrix.
```
vladimir-ch: Maybe we could say here that `Cholesky` must be initialized with `Factorize` and it must return true, otherwise the methods will panic. I'm now sure how to formulate it.

mat64/cholesky.go
```diff
@@ -60,27 +67,37 @@ func (c *Cholesky) Factorize(a Symmetric) (ok bool) {
...
 	work := make([]float64, c.chol.mat.N)
 	norm := lapack64.Lansy(matrix.CondNorm, sym, work)
-	_, ok = lapack64.Potrf(sym)
-	if ok {
+	_, c.valid = lapack64.Potrf(sym)
+	if c.valid {
 		c.updateCond(norm)
 	} else {
+		c.chol.Reset()
 		c.cond = math.Inf(1)
```
vladimir-ch: This could be `math.NaN()`, it won't be used anyway.

mat64/cholesky.go
```diff
@@ -14,13 +14,19 @@ import (
 	"github.com/gonum/matrix"
 )
 
-const badTriangle = "mat64: invalid triangle"
+const (
+	badTriangle = "mat64: invalid triangle"
+	badCholesky = "mat64: invalid Cholesky factorization"
```
kortschak: These should be matrix.Error. If the panic happens it's not our fault.

mat64/cholesky.go
kortschak: NaN is good.

kortschak: *Inf* is good.

mat64/cholesky.go
kortschak: Actually, no. I agree with this. These are not soft panics.

kortschak:
LGTM but wait for @btracey.

vladimir-ch:
I modified this a bit: if Cholesky.chol is Reset whenever an invalid factorization is created, we can use it instead of the bool field to indicate validity of the factorization. PTAL

mat64/cholesky.go
kortschak: "Cholesky methods may only be called on a value that has been successfully initialized by a call to Factorize that has returned true."

mat64/cholesky.go
```diff
@@ -47,7 +50,8 @@ func (c *Cholesky) updateCond(norm float64) {
 }
 
 // Factorize calculates the Cholesky decomposition of the matrix A and returns
-// whether the matrix is positive definite.
+// whether the matrix is positive definite. If the matrix is not positive
+// definite, the factorization must not be used.
 func (c *Cholesky) Factorize(a Symmetric) (ok bool) {
```
kortschak: "If Factorize returns false, the factorization must not be used."

mat64/cholesky.go
```diff
@@ -64,23 +68,39 @@ func (c *Cholesky) Factorize(a Symmetric) (ok bool) {
...
-		c.cond = math.Inf(1)
+		c.Reset()
 	}
 	return ok
 }
 
-// Det returns the determinant of the matrix that has been factorized.
-func (c *Cholesky) Det() float64 {
-	return math.Exp(c.LogDet())
+func (c *Cholesky) Reset() {
```
kortschak: Doc comment.

mat64/cholesky.go
```diff
@@ -328,11 +371,15 @@ func (c *Cholesky) SymRankOne(orig *Cholesky, alpha float64, x *Vector) (ok bool
...
+		c.Reset()
 	}
 	return ok
 }
 
 func (c *Cholesky) isZero() bool {
 	return c.chol == nil
 }
+
+func (c *Cholesky) valid() bool {
```
kortschak: Can you check if this gets inlined? If it doesn't make a checker method that includes the panic to save some lines in the preambles above.

vladimir-ch: It seems that it gets inlined for 1.5.4 and newer.

vladimir-ch:
PTAL

btracey:
I think this is good. I only wonder about things like Size and the extraction routines. Let's say you do a rank-one update to a Cholesy matrix, and it's not positive definite. Could it be useful to extract the matrix to see what it looks like? Or is the algorithm somewhat unpredictable, and thus the errors should be handled in different ways?

mat64/cholesky.go
vladimir-ch: Thanks. Added.

vladimir-ch:
In case of Factorize, if the matrix is not positive definite, Lapack returns some information in info. We don't expose this information in mat64. Also, Lapack says that "... the factorization could not be completed.". I'm not sure if in this situation it is useful to enable extraction of the factorization in an unknown/unfinished/implementation-dependent state. On the other hand, enabling it is harmless but documenting it would not be elegant anymore. Now it is: Factorize must return true, otherwise the methods will panic. Maybe the solution is to return Lapack's `info` from Factorize.

In case of SymRankOne, if the updated matrix is not positive definite, it returns without modifying the receiver, so the updated matrix is not available anyway. It can return with a false later on during the update but that would be only due to the effect of floating point and in a general case the probability of getting true in the equal-to-zero check is next to zero (moreover, in the future I'd like to avoid modifying the receiver in this case, too).

kortschak:
I think we leave it up to the user to engineer an alternative if they want to see that information. It is likely not something that they would use in production, but rather use for debugging.

kortschak:
LGTM

btracey:
LGTM

PR #387, opened 2016-06-17
Fix panic message.

btracey:
The panic messages in blas were renamed. This updates the mat64 test to use the correct panic name

### DIFF
--- a/mat64/dense_test.go
+++ b/mat64/dense_test.go
@@ -830,8 +830,12 @@ func TestMul(t *testing.T) {
 		// These probably warrant a better check and failure. They should never happen in the wild though.
 		temp.mat.Data = nil
 		panicked, message := panics(func() { temp.Mul(a, b) })
-		if !panicked || message != "blas: index of c out of range" {
-			t.Error("exected runtime panic for nil data slice")
+		if !panicked || message != "blas: insufficient matrix slice length" {
+			if message != "" {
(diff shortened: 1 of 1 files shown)


mat64/dense_test.go
```diff
@@ -830,7 +830,7 @@ func TestMul(t *testing.T) {
 		// These probably warrant a better check and failure. They should never happen in the wild though.
 		temp.mat.Data = nil
 		panicked, message := panics(func() { temp.Mul(a, b) })
-		if !panicked || message != "blas: index of c out of range" {
+		if !panicked || message != "blas: insufficient matrix slice length" {
 			t.Error("exected runtime panic for nil data slice")
```
kortschak: While you're here, can you make the condition more informative. Adding the panic message to the error when panicked == true would do it.

kortschak:
LGTM with addition to error message.

mat64/dense_test.go
kortschak: Maybe add

```
if message != "" {
    t.Errorf("exected runtime panic for nil data slice: got %q", message)
} else {
   t.Error("exected runtime panic for nil data slice")
}
```

in place of the current `t.Error("exected runtime panic for nil data slice")`.

kortschak:
Also probably fix the spelling of "expected" while you're here.

PR #392, opened 2016-08-28
mat64: tighten definition of Dot

kortschak:
@btracey Please take a look.

@sjwhitworth This will break github.com/sjwhitworth/golearn/metrics/pairwise, but the breakage is fixed by using mat64.Norm [here](https://github.com/sjwhitworth/golearn/blob/master/metrics/pairwise/poly_kernel.go#L32) and [here](https://github.com/sjwhitworth/golearn/blob/master/metrics/pairwise/euclidean.go#L27). (You'll also get a performance improvement by using *mat64.Vector rather than *mat64.Dense if that will work for you.)

### DIFF
--- a/mat64/matrix.go
+++ b/mat64/matrix.go
@@ -338,53 +338,15 @@ func Det(a Matrix) float64 {
 	return math.Exp(det) * sign
 }
 
-// Dot returns the sum of the element-wise products of the elements of a and b.
+// Dot returns the sum of the element-wise product of a and b.
 // Dot panics if the matrix sizes are unequal.
-func Dot(a, b Matrix) float64 {
-	r, c := a.Dims()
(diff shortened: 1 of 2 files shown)


btracey:
The code is so much simpler too.

LGTM

mat64/matrix.go
```diff
@@ -340,51 +340,13 @@ func Det(a Matrix) float64 {
 
 // Dot returns the sum of the element-wise products of the elements of a and b.
 // Dot panics if the matrix sizes are unequal.
```
vladimir-ch: Could you also update the comment here and remove the "element" stutter in the line above?

kortschak: Becomes, "Dot returns the sum of the element-wise product of a and b."

PR #393, opened 2016-08-30
mat64: fix doc comment

kortschak:


mat64/symmetric.go
```diff
@@ -252,7 +252,7 @@ func (s *SymDense) SymRankK(a Symmetric, alpha float64, x Matrix) {
 	blas64.Syrk(t, alpha, g, 1, s.mat)
 }
 
-// SymOuterK calculates the outer product of a times its transpose and stores
+// SymOuterK calculates the outer product of x times its transpose and stores
```
vladimir-ch: The wording here is a little bit strange because "outer product" already carries the information that it is a matrix multiplication x * x^T. I would write something like (please ammend):
```
// SymOuterK calculates the outer product of x with itself and stores
// the result into the receiver. It is equivalent to matrix multiplication
//  s = alpha * x * x'.
// In order to update an existing matrix, see SymRankOne.
```
Also, note the dot after SymRankOne, and do we use `'` or `^T` to denote a transpose in `matrix`?

kortschak: That seems OK to me. @btracey?

btracey: Yea, although I would say "it is equivalent to _the_ matrix multiplication"

btracey:
LGTM with the change.

PR #394, opened 2016-08-30
mat64: add Reset method to SymDense

kortschak:
This is needed for an allocation improvement to stat.CovarianceMatrix and stat.CorrelationMatrix.

PR #396, opened 2016-09-01
mat64: export dense matrix IsZero methods

kortschak:
Blocked on #394.

@btracey Please take a look.

Not wildly happy with the comment wording.

### DIFF
--- a/mat64/dense.go
+++ b/mat64/dense.go
@@ -69,7 +69,7 @@ func (m *Dense) reuseAs(r, c int) {
 		// Panic as a string, not a mat64.Error.
 		panic("mat64: caps not correctly set")
 	}
-	if m.isZero() {
+	if m.IsZero() {
 		m.mat = blas64.General{
 			Rows:   r,
 			Cols:   c,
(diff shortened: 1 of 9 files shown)


mat64/dense.go
```diff
@@ -136,7 +136,22 @@ func (m *Dense) isolatedWorkspace(a Matrix) (w *Dense, restore func()) {
...
+//
+// See the Reseter interface for more information.
+func (m *Dense) Reset() {
+	// Row, Cols and Stride must be zeroed in unison.
+	m.mat.Rows, m.mat.Cols, m.mat.Stride = 0, 0, 0
+	m.capRows, m.capCols = 0, 0
+	m.mat.Data = m.mat.Data[:0]
+}
+
+// IsZero returns whether the receiver is a zero value. A zero value may be
```
btracey: Maybe: `IsZero returns whether the receiver is zero-sized. Zero-sized matrices can be the receiver for size-restricted operations. Dense matrices can be zeroed using Reset()`

btracey:
Just saw your comment in stat. You're right, we should just be able to check the dimensions, so maybe it's better to keep this private.

kortschak:
That works at the moment, but we could change it unless we document that that is the definition.

kortschak:
Pending in gonum/gonum.

PR #399, opened 2016-10-08
mat64: use lapack.EVJob in EigenSym.Factorize

vladimir-ch:


vladimir-ch:
Merged because it fixes build. PTAL

PR #404, opened 2016-11-11
Replace Eigen with lapack-based implementation

btracey:


btracey:
This seems like the best signature for Factorize, but I'm a bit worried about the confusion it will cause. I think many users have only heard of "eigenvectors" and don't realize the two classes. I'm hoping that the documentation and the name asymmetry (which in my vision will eventually become Eigenvectors and LeftEigenvectors as methods on Dense) points users to what they want.

mat64/eigen.go
```diff
@@ -153,824 +207,52 @@ func (e *Eigen) Factorize(a Matrix, vectors bool) (ok bool) {
...
 	return dst
 }
 
-// Vectors returns the eigenvectors of the decomposition.
+// Vectors returns the right eigenvectors of the decomposition. Vectors
+// will panic if the right eigenvectors were not computed during the factorization,
+// or if the factorization was not successful.
 //
-// This signature and behavior will change when issue #308 is resolved.
+// See the documentation in lapack64.Geev for the format of the vectors.
```
kortschak: Can we have documentation for this here rather than pointing users to lapack.

PR #411, opened 2016-12-11
mat64: return Condition(inf) from Dense.Inverse for singular matrix

vladimir-ch:
Fixes #410

Issue #410, opened 2016-12-11
matrix/mat64: Dense.Inverse does not fail on singular matrix


ChristopherRabotin:
One should expect Dense.Inverse to return an error for singular matrices. Instead, it returns garbage: for a 2x2, it returns the input matrix, for a 4x4 it returns the input matrix where some values are changed. I have not checked for other sizes.

Code:
```go
package main

import (
	"fmt"

	"github.com/gonum/matrix/mat64"
)

func main() {
	A := mat64.NewDense(2, 2, []float64{0, 0, 0, 1})
	var InvA mat64.Dense
	err := InvA.Inverse(A)
	if err != nil {
		panic(err)
	}
	fmt.Printf("%v\n", mat64.Formatted(&InvA, mat64.Prefix("")))

	B := mat64.NewDense(4, 4, []float64{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 20, 20, 0, 0, 20, 20})
	var InvB mat64.Dense
	err = InvB.Inverse(B)
	if err != nil {
		panic(err)
	}
	fmt.Printf("%v\n", mat64.Formatted(&InvB, mat64.Prefix("")))
}
```

Result:
```
⎡0  0⎤
⎣0  1⎦
⎡ 0   0   0   0⎤
⎢ 0   0   0   0⎥
⎢ 0   0  20  20⎥
⎣ 0   0   1   0⎦
```

vladimir-ch:
Yes, that should return an error. I've submitted a possible fix in #411 .
At the moment the receiver is always modified even if an error is returned.

vladimir-ch:
Thanks for spotting this, @ChristopherRabotin

ChristopherRabotin:
Sorry to be the bearer of bad news, but now, the Inverse now fails on matrices with small-ish numbers, although it really should not. The matrix used in this example is invertible on Matlab up until the exponent `-15`. I also wonder whether this isn't an issue for lapack instead.

```go
package main

import (
	"fmt"
	"math"

	"github.com/gonum/matrix/mat64"
)

func main() {
	//Q := mat64.NewSymDense(3, []float64{25, 0, 0, 0, 25, 0, 0, 0, 25})
	Q := mat64.NewSymDense(3, []float64{2.5e-9, 6.25e-7, (25e-5) / 3, 6.25e-7, (5e-1) / 3, 2.5e-2, (25e-5) / 3, 2.5e-2, 5})
[...]

PR #413, opened 2016-12-12
mat64: extend TestInverse

### LINKED ISSUE #412: mat64: Dense.Inverse test coverage needs improvement
There is no testing for matrices that are not RawMatrixer or matrices that are a transposed RawMatrixer.

kortschak:
@vladimir-ch @btracey Please take a look for discussion.

The single failing case is one where I have ignored the shadowing restrictions, so we do not guarantee that it will work. I am not entirely sure how I want to handle this. I think that probably leaving the test case in as a comment explaining why it would fail is the correct answer. What are your views?

Fixes #412.

### DIFF
--- a/mat64/dense_arithmetic.go
+++ b/mat64/dense_arithmetic.go
@@ -228,15 +228,7 @@ func (m *Dense) Inverse(a Matrix) error {
 			m.Copy(a)
 		}
 	default:
-		if m != aU {
-			m.Copy(a)
-		} else if aTrans {
-			// m and a share data so Copy cannot be used directly.
-			tmp := getWorkspace(r, c, false)
-			tmp.Copy(a)
-			m.Copy(tmp)
-			putWorkspace(tmp)
-		}
+		m.Copy(a)
 	}
 	ipiv := make([]int, r)
 	lapack64.Getrf(m.mat, ipiv)
--- a/mat64/dense_test.go
+++ b/mat64/dense_test.go
@@ -5,6 +5,7 @@
 package mat64
 
 import (
+	"fmt"
 	"math"
 	"math/rand"
 	"reflect"
@@ -1615,8 +1616,8 @@ func TestOuter(t *testing.T) {
 
 func TestInverse(t *testing.T) {
 	for i, test := range []struct {
-		a    *Dense
-		want *Dense
+		a    Matrix
+		want Matrix
 		tol  float64
 	}{
 		{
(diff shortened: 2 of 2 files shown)


ChristopherRabotin:
I don't see any test on singular matrices, which is how #410 got to slip through the cracks. Maybe a few should be added?

kortschak:
No, I'll leave that for @vladimir-ch in #411. This set of tests is to ensure that our typish system works properly.

mat64/dense_test.go
```diff
@@ -1633,6 +1634,56 @@ func TestInverse(t *testing.T) {
...
+				0.147222222222222, -0.144444444444444, 0.063888888888889,
+				-0.061111111111111, 0.022222222222222, 0.105555555555556,
+				-0.019444444444444, 0.188888888888889, -0.102777777777778,
+			}),
+			tol: 1e-14,
+		},
+
+		// The following test will fail since it does not obey the shadowing
+		// rules. The basicMatrix type allows shadowing of the input data
+		// without providing the Raw mathod required for detection of shadowing.
```
btracey: This is a bit of a contrived example, which is fine, but I think the comment could be adjusted to make that more clear. 

"The following case fails as it disobeys the shadowing rules. Specifically, the test below extracts the underlying *Dense, and uses it as a recevier with the basicMatrix as input. The basicMatrix type allows shadowing of the input data without providing the Raw mathod required for detection of shadowing. We specifically state we cannot check this case."

mat64/dense_arithmetic.go
```diff
@@ -228,15 +228,7 @@ func (m *Dense) Inverse(a Matrix) error {
...
-		if m != aU {
-			m.Copy(a)
-		} else if aTrans {
-			// m and a share data so Copy cannot be used directly.
-			tmp := getWorkspace(r, c, false)
-			tmp.Copy(a)
-			m.Copy(tmp)
-			putWorkspace(tmp)
-		}
+		m.Copy(a)
```
btracey: Does this work for, say RawTriangular? It doesn't look like it would to me

btracey:
I think leaving the test commented is good. It's probably nice to have a few shadowing examples somewhere when we have somewhere to put that kind of extra documentation.

mat64/dense_arithmetic.go
kortschak: What do you mean does it work for RawTriangular? The rules say that we won't check for shadowing unless "the Raw type matches that of the receiver", which it doesn't here.

btracey: Good point. Nice cleanup then.

kortschak:
PTAL

btracey:
LGTM

Sorry for the delay, I missed you had fixed.

PR #414, opened 2016-12-12
Add method to set cholesky from a triangular matrix

btracey:


btracey:
This is needed for the Wishart probability distribution, which I will be creating a PR for shortly.

mat64/cholesky.go
```diff
@@ -86,6 +86,21 @@ func (c *Cholesky) Reset() {
 	c.cond = math.Inf(1)
 }
 
+// SetFromUpper sets the Cholesky decomposition from the given
```
kortschak: s/SetFromUpper/SetFromU/ or the converse.

Also, from the given ...? and presumably there should be a warning about the triangle kind and panicking and note that t is not altered or retained.

btracey: Sorry, I have no idea what I was thinking when I wrote that comment. 

Done.

Issue #405, opened 2016-11-16
matrix/mat64: Should view follow Go slicing syntax


btracey:
Dense.View() asks for a starting point, and a number of rows to extend. It feels like View should take Go slicing syntax; instead specifying {lo, high}

Specifically, rather than `func (m *Dense) View(i, j, r, c int) Matrix`, perhaps it should instead be `func (m *Dense) View(rLo, rHi, cLo, cHi)`.

kortschak:
How often do we take the result of a Dims call and use that in r and c and how often do we view from one element coordinate to another?

If we do make this change I would want it to go through a three step process. 1. Create a Slice method with the new semantics above, 2. replace the calls to View with calls to Slice, and 3. rename Slice to view.

This is to avoid errors during the change - the types are the same so this is real risk.

sbinet:
we should provide an `eg` script to migrate in one go.
see, for example:
 https://github.com/sbinet/envol-2016-go/blob/master/api/api_eg.go

that would probably read, for this api migration:

```go
// +build ignore

// package P is a placeholder to run eg, like so:
//  $> eg -t ./api_eg.go -w -transitive my/pkg/p1 my/pkg/p2
package P

import (
	"github.com/gonum/matrix/mat64"
)

func before(m *mat64.Dense, i, j, r, c int) mat64.Matrix { return m.View(i, j, r, c) }
func after(m *mat64.Dense, i, j, r, c int) mat64.Matrix  { return m.Slice(i, i+r, j, j+c) }
```

kortschak:
This has been done in gonum/gonum.

PR #416, opened 2016-12-15
mat64: replace View use with Slice

kortschak:
Updates #405.

There is an ongoing risk that old View calls remain in dependent code, so it might be worth keeping the new method as Slice and then at some point removing View.

The documentation could do with a polish and I'm not sure if we want to add a Slicer interface.

Grow should probably change too.

@btracey Please take a look.


btracey:
LGTM

I agree that grow should probably change also. I can't think of a similar name recast off of the top of my head. 

I have some code that does use View as an interface, but it needs a lot more than that, so it's not clear that the interface is all that fundamental to the package.

kortschak:
GrowTo?

If you're happy with that, I'll add it to the PR.

Are you happy with the docs? The issue I had was that I don't quite like the wording for the upper slice index. Also, what do we want to do with View? Delete? How long from now?

btracey:
I like GrowTo.

The language in the spec is "The result has indices starting at 0 and length equal to high - low"

Maybe "The new matrix starts at {i,j} and extends k-i rows and l-j columns. Thus, the final row in the resulting matrix is k-1 and the final column is l-1 (this is like Go slicing syntax)."

 Should it really be `Slice(i, k, j, l)` so it matches the `[i:k, j:l]` syntax? If both `View` and `Grow` are going to stay, then they should match, but if we're eliminating `View ` then I personally like the alternate order.

I don't think there's any need to have both View and Slice. Maybe we break it when we break everything?

kortschak:
PTAL

I'm going to leave GrowTo since I'm not happy with the way we should handle the case where the dimensions are less than the current.

btracey:
This reads well now.

LGTM

sbinet:
should we provide an `eg` script to migrate in one go?
see, for example:
 https://github.com/sbinet/envol-2016-go/blob/master/api/api_eg.go

that would probably read, for this api migration:

```go
// +build ignore

// package P is a placeholder to run eg, like so:
//  $> eg -t ./api_eg.go -w -transitive my/pkg/p1 my/pkg/p2
package P

import (
)

func before(m *mat64.Dense, i, j, r, c int) mat64.Matrix { return m.View(i, j, r, c) }
func after(m *mat64.Dense, i, j, r, c int) mat64.Matrix  { return m.Slice(i, i+r, j, j+c) }
```

sbinet:
on this simple `main.go`:

```go
package main

import "github.com/gonum/matrix/mat64"

func main() {
	var m *mat64.Dense
	m.View(0, 1, 0, 3)
}
```

it gives:
```sh
$> eg -t eg.go -v -w .
before: m.View(i, j, r, c)
after: m.Slice(i, i+r, j, j+c)
/home/binet/tmp/go/src/matmig/main.go:7:9: wildcard i -> 0?: primary match
/home/binet/tmp/go/src/matmig/main.go:7:12: wildcard j -> 1?: primary match
/home/binet/tmp/go/src/matmig/main.go:7:15: wildcard r -> 0?: primary match
/home/binet/tmp/go/src/matmig/main.go:7:18: wildcard c -> 3?: primary match
m.View(i, j, r, c) matches m.View(0, 1, 0, 3) with: j->1 r->0 c->3 m->m i->0
=== /home/binet/tmp/go/src/matmig/main.go (1 matches)

$> git diff
```

```diff
diff --git a/main.go b/main.go
index 6443ecc..6354489 100644
--- a/main.go
+++ b/main.go
@@ -4,5 +4,5 @@ import "github.com/gonum/matrix/mat64"
 
 func main() {
        var m *mat64.Dense
-       m.View(0, 1, 0, 3)
+       m.Slice(0, 0+0, 1, 1+3)
 }
```

sbinet:
and according to sourcegraph, `mat64.Dense.View` is referenced in 10 repos:
https://sourcegraph.com/github.com/gonum/matrix/-/info/GoPackage/github.com/gonum/matrix/mat64/-/Dense/View

(even, yes, in JuliaLang! :P)

kortschak:
It's worth putting the eg script into #405 for when it is removed.

All those uses are in either our code or @btracey's code (he wrote most of the current perf.go implementation as it stands now).

btracey:
Sourcegraph is definitely missing some. I know I use View in github.com/btracey/gaussproc, but sourcegraph does not list that.

sbinet:
sourcegraph says 10 repos, but only displays 3, plus the ellipsis `...` :) (I guess you need to pay to get the others)

sbinet:
godoc says 17: https://godoc.org/github.com/gonum/matrix?importers (but it works at the package level, not the func/method level)

btracey:
It lists 10 packages though, so that's confusing if your interpretation is correct.

btracey:
Actually, 195 for mat64

https://godoc.org/github.com/gonum/matrix/mat64?importers

PR #417, opened 2016-12-16
mat64: reduce allocations when SVD is reused

kortschak:
This comes out of discussion in the CCA PR in stat https://github.com/gonum/stat/pull/123#issuecomment-267486046. This change makes it more valuable to allow the CC to be reused. So that should probably happen - the SVD values need to be retained, so the internals need a bit of rework there when that happens.

I am trusting that the u, vt and s do not need to be zeroed - there is no indication in the docs that they do, and the tests reuse the SVD value, so I think this is OK, but @btracey please reassure me.

### DIFF
--- a/mat64/svd.go
+++ b/mat64/svd.go
@@ -57,13 +57,13 @@ func (svd *SVD) Factorize(a Matrix, kind matrix.SVDKind) (ok bool) {
 			Rows:   m,
 			Cols:   m,
 			Stride: m,
-			Data:   make([]float64, m*m),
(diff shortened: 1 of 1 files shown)


btracey:
LGTM

@vladimir-ch would have a better memory than me on SVD, but none of the routines I recall require the data to be zeroed.

kortschak:
I'll merge on the basis that the tests don't fail and your recollection and my reading of the code and the docs suggests it's OK. The fix is easy if I am wrong here.

Issue #419, opened 2016-12-30
matrix/mat64: add `Exp' to TriDense


ChristopherRabotin:
Upper and lower triangular matrices can be assimilated to [dual numbers](https://en.wikipedia.org/wiki/Dual_number), herein represented as `a+b\epsilon`. There's an interesting property with dual numbers where `exp(b\epislon)=1+b\epsilon` (where \epsilon is a N dimensional representation of a dual number), cf. [this paper](http://www.meccanicamente.org/papers/dualispringer2008.pdf). This can be trivially grown to any (2n)x(2n) triangular matrix, but is not trivially applicable to triangular matrices of even rows/columns.

Hence, implementing `Exp` on the Triangular struct would ensure an extremely simple computation as described below in pseudo code:
```
let A be a lower triangular;
if diagonal is non-nil:
    extract diagonal as a scaled identity matrix
    expDiag <- trivially compute its exponential;
endif
expLower <- expDiag * (identity + A)
```

If the matrix is an upper triangular, perform the same computation on the transpose of A (and return the transpose).

If there is no objection to this feature, I volunteer to implement it.

_Edit:_ I forgot to mention that trivial operations also include sums, products and divisions.

_Second edit_: Looking a bit more into what is currently available in `TriDense`, I see that part of what is in this issue would involve enhancing `MulTri` for the specific case where there is an even number of rows/columns.

btracey:
I don't understand what's being suggested. The original title is to add `Exp` to Triangular. That seems fine. By "arithmetic functions" do you mean `Add` etc? That would be fine too, though it would probably better to have a separate tracking issue.

Is the part about dual numbers an implementation suggestion for `Exp`, or a suggestion to modify the implementation of `TriDense` to support the use case of a triangle of dual numbers?

(I do have https://github.com/btracey/hyperdual if you need it)

ChristopherRabotin:
The suggestion is only to add an `Exp` function to TriDense. Subsequently, I think it could be interesting to add `Sub` to TriDense and slightly modify `Add` of TriDense to simplify the operation if the provided matrix has an even number of rows.

There will be no support for dual numbers, I'll just be using some of their properties as a tool to simplify the number of operations needed.

(P.S.: I think there's a bug in your `hyperdual` implementation in the Add function. I think line 44 should be `a.data[i] += v`.)

ChristopherRabotin:
[...]

PR #420, opened 2016-12-31
Add `Exp` to TriDense

ChristopherRabotin:
Implements #419 by using cool properties of triangular matrices which have a size of powers of two (1, 2, 4, 8, ...).

*Note:* This is my first contribution of gonum, so it's very likely that I'm misusing many of the internal functions provided in the library. For example, I may be wrong in understanding that `testOneInput` generates random matrices based on the provided type and size.

### DIFF
--- a/mat64/triangular.go
+++ b/mat64/triangular.go
@@ -402,6 +402,14 @@ func (t *TriDense) MulTri(a, b Triangular) {
 func (t *TriDense) Exp(a Triangular) {
 	n, kind := a.Triangle()
 	t.reuseAs(a.Triangle())
+
+	aU, _ := untransposeTri(a)
+	t.reuseAs(n, kind)
+	var restore func()
+	if t == aU {
+		t, restore = t.isolatedWorkspace(aU)
+		defer restore()
+	}
 	t.Copy(a)
 
 	if n&(n-1) == 0 {
--- a/mat64/triangular.go
+++ b/mat64/triangular.go
@@ -124,6 +124,7 @@ func NewTriDense(n int, kind matrix.TriKind, mat []float64) *TriDense {
 	}
 }
 
+// Dims returns the dimensions of t.
 func (t *TriDense) Dims() (r, c int) {
 	return t.mat.N, t.mat.N
 }
@@ -392,6 +393,62 @@ func (t *TriDense) MulTri(a, b Triangular) {
 			}
 			t.SetTri(i, j, v)
 		}
+	}
+}
+
+// Exp calculates the exponential of the matrix a, e^a, placing the result
+// in the receiver. If a has an odd number of rows, the matrix will be converted
+// to a Dense to perform the calculation.
+func (t *TriDense) Exp(a Triangular) {
+	n, kind := a.Triangle()
+	t.reuseAs(a.Triangle())
+	t.Copy(a)
+
+	if n&(n-1) == 0 {
+		// Exponential is trivial for sizes of powers of two. Calculation performed on lower triangular.
+		aP := NewTriDense(n, kind, nil)
+		isUpper := kind == matrix.Upper
+		if isUpper {
+			aP.Copy(a.T())
+		} else {
+			aP.Copy(a)
+		}
+		computeKind := matrix.Lower
+		*t = *NewTriDense(n, computeKind, nil)
+		diag := NewTriDense(n, computeKind, nil)
+		eye := NewTriDense(n, computeKind, nil)
+		nilDiag := true
+		for i := 0; i < n; i++ {
+			eye.SetTri(i, i, 1)
+			v := aP.At(i, i)
+			if v != 0 {
+				nilDiag = false
+			}
+			diag.SetTri(i, i, math.Exp(v))
+			aP.SetTri(i, i, 0)
+		}
+		// TODO(ChristopherRabotin): Use (*TriDense).Add(a, b) once written.
+		// Compute exponential which is simply diag*(eye + aP) (if diag != [0]).
+		for i := 0; i < n; i++ {
+			for j := 0; j < n; j++ {
+				if i >= j {
+					t.SetTri(i, j, eye.At(i, j)+aP.At(i, j))
+				}
+			}
+		}
+		if !nilDiag {
+			t.MulTri(diag, t)
+		}
+		if isUpper {
+			t.Copy(t.T())
+		}
+	} else {
+		// Convert to Dense then convert back to Triangle.
+		// The exponential of a triangular matrix is always a triangular matrix
+		// of the same kind.
+		var triExp Dense
+		triExp.Exp(DenseCopyOf(t))
+		t.Copy(&triExp) // Store result in receiver.
 	}
 }
 
--- a/mat64/triangular.go
+++ b/mat64/triangular.go
@@ -405,42 +405,48 @@ func (t *TriDense) Exp(a Triangular) {
 	t.Copy(a)
 
 	if n&(n-1) == 0 {
-		// Exponential is trivial for sizes of powers of two. Calculation performed on lower triangular.
-		aP := NewTriDense(n, kind, nil)
-		isUpper := kind == matrix.Upper
-		if isUpper {
+		// Exponential is trivial for sizes of powers of two.
+		// Calculation performed on upper triangular.
+		computeKind := matrix.Upper
+		aP := NewTriDense(n, computeKind, nil)
+		if kind != computeKind {
 			aP.Copy(a.T())
 		} else {
 			aP.Copy(a)
 		}
-		computeKind := matrix.Lower
-		*t = *NewTriDense(n, computeKind, nil)
-		diag := NewTriDense(n, computeKind, nil)
+		tUp := *NewTriDense(n, computeKind, nil)
 		eye := NewTriDense(n, computeKind, nil)
+		diagExp := NewTriDense(n, computeKind, nil)
 		nilDiag := true
 		for i := 0; i < n; i++ {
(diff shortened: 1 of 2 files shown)


ChristopherRabotin:
After running a few benchmarks to compare the Dense `Exp` and the TriDense `Exp` for matrix sizes 2, 8, 128 and 1024, I suppose that there are a few optimizations I can make to the code to make it faster. I suppose these may be related to the several calls I make to `Copy`, but I'm not sure, so any hint is welcome. I'm especially surprised at the time needed for a large 1024*1024 matrix.

For the record, here are the benchmark results, along with the benchmarking code.
### Results
(Re-arranged for readability.)
```
BenchmarkExp2-4                                   200000               7300 ns/op
BenchmarkTriExp2-4                               1000000               2467 ns/op
BenchmarkExp8-4                                    50000              23723 ns/op
BenchmarkTriExp8-4                                200000               9852 ns/op
BenchmarkExp128-4                                    100           18175488 ns/op
BenchmarkTriExp128-4                                 100           12559400 ns/op
BenchmarkExp1024-4                                     1         4787340326 ns/op
BenchmarkTriExp1024-4                                  1        19034451063 ns/op
```

### Benchmarking code
```go
func randTri(size int, rho float64, rnd func() float64) (*TriDense, error) {
	if size == 0 {
		return nil, matrix.ErrZeroLength
	}
	d := NewTriDense(size, matrix.Upper, nil)
	for i := 0; i < size; i++ {
		for j := 0; j < size; j++ {
			if j >= i && rand.Float64() < rho {
				d.SetTri(i, j, rnd())
			}
		}
	}
	return d, nil
}

func BenchmarkTriExp2(b *testing.B)    { expTriBench(b, 2) }
func BenchmarkTriExp8(b *testing.B)    { expTriBench(b, 8) }
func BenchmarkTriExp128(b *testing.B)  { expTriBench(b, 128) }
func BenchmarkTriExp1024(b *testing.B) { expTriBench(b, 1024) }

func expTriBench(b *testing.B, size int) {
	a, _ := randTri(size, 1, rand.NormFloat64)
	b.ResetTimer()
	var m TriDense
	for i := 0; i < b.N; i++ {
		m.Exp(a)
	}
}
```

btracey:
Could you run with GOMAXPROCS = 1? Dense.Mul is coded in parallel while the others are not. Curious if that's the only cause.

ChristopherRabotin:
Using GOMAXPROCS=1 does lead to better results indeed, although TriDense.Exp is still slower for 1024x1024.

(re-arranged results for readability again)
```
BenchmarkExp2                                     200000               5962 ns/op
BenchmarkTriExp2                                 1000000               1658 ns/op
BenchmarkExp8                                     100000              20542 ns/op
BenchmarkTriExp8                                  200000               8188 ns/op
BenchmarkExp128                                      100           20477942 ns/op
BenchmarkTriExp128                                   100           11087205 ns/op
BenchmarkExp1024                                       1        10183199710 ns/op
BenchmarkTriExp1024                                    1        14255916781 ns/op
```

mat64/triangular.go
```diff
@@ -395,6 +396,76 @@ func (t *TriDense) MulTri(a, b Triangular) {
 	}
 }
 
+// Exp calculates the exponential of the matrix a, e^a, placing the result
+// in the receiver. If a has an odd number of rows, the matrix will be converted
+// to a Dense to perform the calculation.
+func (t *TriDense) Exp(a Triangular) {
+	n, kind := a.Triangle()
+	t.reuseAs(a.Triangle())
```
btracey: This needs a check that the kind of `t` is the same as `a`. 

@kortschak maybe reuseAs should check that the kind matches the receiver kind if isZero returns false? This is nice looking code.

mat64/triangular.go
```diff
@@ -395,6 +396,76 @@ func (t *TriDense) MulTri(a, b Triangular) {
...
+
+	aU, _ := untransposeTri(a)
+	t.reuseAs(n, kind)
```
btracey: You shouldn't need to call reuseAs a second time. It'll have no effect.

mat64/triangular.go
```diff
@@ -395,6 +396,76 @@ func (t *TriDense) MulTri(a, b Triangular) {
...
+	var restore func()
```
btracey: This doesn't actually need an isolated workspace (I don't think, especially given the copy below). IsolatedWorkspace exist for things like `Mul` where if you write `A = A*B`, you need a workspace so you don't overwrite values of `A` while doing the multiplication.

mat64/triangular.go
```diff
@@ -395,6 +396,76 @@ func (t *TriDense) MulTri(a, b Triangular) {
...
+			tUp.MulTri(diagExp, &tUp)
+		}
+		*t = *NewTriDense(n, kind, nil)
+		if kind == computeKind {
+			t.Copy(&tUp)
+		} else {
+			t.Copy(tUp.T())
+		}
+	} else {
+		// Convert to Dense then convert back to Triangle.
```
btracey: Add a TODO here to improve the implementation (or improve the implementation) to take advantage of `Tri*Tri` algorithms.

mat64/triangular.go
```diff
@@ -395,6 +396,76 @@ func (t *TriDense) MulTri(a, b Triangular) {
...
+	t.reuseAs(n, kind)
+	var restore func()
+	if t == aU {
+		t, restore = t.isolatedWorkspace(aU)
+		defer restore()
+	}
+	t.Copy(a)
+
+	if n&(n-1) == 0 {
+		// Exponential is trivial for sizes of powers of two.
```
btracey: I would say

````
// Special case exponentiation for sizes that are power-of-two. The exponent can be expressed as 
// [Implementation detail]. See
//  [citation]
// for more information.
````

File: mat64/triangular.go [hunk 6]
```diff
@@ -395,6 +396,76 @@ func (t *TriDense) MulTri(a, b Triangular) {
...
+		// Calculation performed on upper triangular.
+		computeKind := matrix.Upper
+		aP := NewTriDense(n, computeKind, nil)
+		if kind != computeKind {
+			aP.Copy(a.T())
+		} else {
+			aP.Copy(a)
+		}
+		tUp := *NewTriDense(n, computeKind, nil)
```
btracey: It seems like there are a lot of optimizations here. It'll require special casing for the computeKind, but I think even with the duplication we'll still be reducing lines of code. 

Big ideas:
1) Eye is just 1 or 0, so that's easily special cased in setting.
2) diagExp is just a diagonal, so it doesn't need a full TriDense worth of storage.
3) the multiplication with a diagonal is easy, it's just scaling the row.

I think it would look something like
````
if n%(n-1) == 0 {
    if t.kind == matrix.Upper {
         for i := 0; i < n; i++ {
              v := t.At(i,i)
              exp := math.Exp(v)
              t.Set(i,i, (1+v) * exp)
              for j := i+1; j < n; j++ {   // note the indexing optimization
                     v := t.At(i,j)                 
                     t.SetTri(i, j, v * exp)
              }
        }
}
````

With a similar code for the Lower case.

File: mat64/triangular_test.go [hunk 7]
```diff
@@ -315,3 +315,51 @@ func TestTriMul(t *testing.T) {
...
+	legalTypesUpper := func(a Matrix) bool {
+		if !legalTypes(a) {
+			return false
+		}
+		_, kind := a.(Triangular).Triangle()
+		r := kind == matrix.Upper
+		return r
+	}
+
+	for n := 0; n < 8; n++ {
```
btracey: I don't think you need to run this 8 times.

btracey: You should add a test for the power of two code that compares the result to Dense (or something).

btracey:
Adding benchmarks is always welcome.

mat64/triangular.go
ChristopherRabotin: Okay, that makes sense, but if I run the tests without isolating the work space I get the following error repeated a number of times:
```
list_test.go:925: Original data slice not modified when a maybeSame: *mat64.TriDense.Exp(*mat64.TriDense), size: struct { ar int; ac int }{ar:1, ac:1}, atrans false
```

btracey: Ah. This is because of the trickery you're pulling with NewTriDense (and I'm suggesting gets removed).

You're modifying the pointer to the receiver, but not the pointer to the input. If one has `m.Mul(m)`, the expectation would be that the data to `m` would change. Similarly, if one had
````
m := mat64.NewTriDense(blah)
a := m
m.Exp(a)
````
One would expect that the data for both a and m to change since they are pointers to the same data. This is the test you're failing, since you're changing the reference to `m` which breaks that `m` and `a` share the same data.

**Comment by ChristopherRabotin** [comment]
Sorry I've wasted your time: the math is incorrect. I'm canceling this PR.

My main citation was the following [paper](http://www.meccanicamente.org/papers/dualispringer2008.pdf) which does not extend the dual number representation to anything larger than a 2 by 2 for an exponential. It seems intuitive to me, and I was looking for a paper while working on this code. Now I understand why I didn't find any.

Counter examples include the following matrix:
```
     3     4     5     6
     0     7     8     9
     0     0    10    11
     0     0     0    12
```

mat64/triangular_test.go
ChristopherRabotin: For future reference, doesn't `testOneInput` do the comparison since it is provided with the `denseComparison` parameter?

btracey: Yes, it does, but this code had special cased certain sizes, so it should test explicitly for those sizes. `testOneInput` just tests an arbitrary set of sizes.

PR #422, opened 2017-01-02
mat64: Update format example documentation to include extra verbs

### LINKED ISSUE #421: Update format documentation to include additional format verbs
The format documentation only shows examples of using the `%v` verb for format.
It took me looking at issue #189 to figure that you could use `%3.3g` to limit the width of each field. 
I tried forking to update the following test: https://github.com/gonum/matrix/blob/20e0dce1afa810a68a601b427ec7b91ac31234f2/mat64/format_example_test.go#L21

But there is no simple makefile to install all of the repo's dependencies so my test fails.

Anyway, maybe adding something like: 

`fmt.Printf("for large exponents:\na = %3.3g\n\n", fa)`

To the example would help and indicating that you support all verbs in https://golang.org/pkg/fmt/ .

Thanks!

DavidGamba:
Fixes #421

### DIFF
--- a/mat64/format_example_test.go
+++ b/mat64/format_example_test.go
@@ -13,18 +13,19 @@ import (
 func ExampleFormatted() {
 	a := mat64.NewDense(3, 3, []float64{1, 2, 3, 0, 4, 5, 0, 0, 6})
 
-	// Create a matrix formatting value with a prefix ...
-	fa := mat64.Formatted(a, mat64.Prefix("    "))
+	// Create a matrix formatting value with a prefix and calculating each column width individually...
+	fa := mat64.Formatted(a, mat64.Prefix("    "), mat64.Squeeze())
 
 	// and then print with and without zero value elements.
 	fmt.Printf("with all values:\na = %v\n\n", fa)
 	fmt.Printf("with only non-zero values:\na = % v\n\n", fa)
 
 	// Modify the matrix...
-	a.Set(0, 2, 0)
+	a.Set(0, 2, 123.456)
 
-	// and print it without zero value elements.
-	fmt.Printf("after modification with only non-zero values:\na = % v\n\n", fa)
+	// and print it using scientific notation for large exponents.
+	fmt.Printf("after modification for scientific notation:\na = % .2g\n\n", fa)
+	// All golang.org/pkg/fmt/ floating-point verbs should work.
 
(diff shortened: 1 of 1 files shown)


mat64/format_example_test.go
```diff
@@ -13,18 +13,19 @@ import (
 func ExampleFormatted() {
 	a := mat64.NewDense(3, 3, []float64{1, 2, 3, 0, 4, 5, 0, 0, 6})
 
-	// Create a matrix formatting value with a prefix ...
-	fa := mat64.Formatted(a, mat64.Prefix("    "))
+	// Create a matrix formatting value with a prefix and calculating each column width individually...
```
kortschak: Please split this over two lines (think about the formating in godoc - or run `godoc -http :6060` locally and check).

mat64/format_example_test.go
```diff
@@ -13,18 +13,19 @@ import (
...
 
 	// Modify the matrix...
-	a.Set(0, 2, 0)
+	a.Set(0, 2, 123.456)
 
-	// and print it without zero value elements.
-	fmt.Printf("after modification with only non-zero values:\na = % v\n\n", fa)
+	// and print it using scientific notation for large exponents.
+	fmt.Printf("after modification for scientific notation:\na = % .2g\n\n", fa)
+	// All golang.org/pkg/fmt/ floating-point verbs should work.
```
kortschak: I don't like "should" here. They will work if they work for `fmt.Print`. Otherwise it's a bug on us.

mat64/format_example_test.go
```diff
@@ -37,10 +38,10 @@ func ExampleFormatted() {
 	//     ⎢.  4  5⎥
 	//     ⎣.  .  6⎦
 	//
-	// after modification with only non-zero values:
-	// a = ⎡1  2  .⎤
-	//     ⎢.  4  5⎥
-	//     ⎣.  .  6⎦
+	// after modification for scientific notation:
```
kortschak: This breaks the semantics of this sentence. Please do this separately rather than just replacing the zeroing example.

mat64/format_example_test.go
DavidGamba: Thanks, I wasn't sure about that one :-P

kortschak:
Thank you.

PR #423, opened 2017-01-11
mat64: use new internal/asm API

kortschak:
@btracey @vladimir-ch @Kunde21 Please take a look.

Requires https://github.com/gonum/internal/pull/30.


btracey:
LGTM

PR #426, opened 2017-02-08
Add Clone method for Cholesky

btracey:


mat64/cholesky.go
```diff
@@ -103,6 +107,23 @@ func (c *Cholesky) SetFromU(t *TriDense) {
 	c.updateCond(-1)
 }
 
+// Clone makes a copy of the input Cholesky into the receiver, overwriting the
+// previous value of the receiver. Clone does not place any restrictions on receiver
+// shape. Clone panics if the input Cholesky is not the result of a valid decomposition.
+func (c *Cholesky) Clone(chol *Cholesky) {
```
sbinet: API-wise, I would have expected this to return the copy.
I understand gonum/matrix strives to reduce allocation, so this was discarded.

but it makes for some head scratching code as the data flow is not really obvious just by looking at the code.

maybe rename this as: `CloneFrom` ? (to parallel `io.ReaderFrom`)

btracey: I don't think so. This follows the spirit of the invariant: "Matrix input arguments to functions are never directly modified. If an operation changes Matrix data, the mutated matrix will be the receiver of a function."

Given this invariant, the modified `Cholesky` should be the receiver. At that point, the From in `CloneFrom` is redundant. This also matches (*Dense).Clone.

sbinet: ok

PR #433, opened 2017-03-10
mat64: add generalised singular value decomposition

kortschak:

Depends on https://github.com/gonum/lapack/pull/249.

This is functional, but I think the API is open for discussion for a few reasons.

1. Currently there is no interface for the user to gain the permuted alpha/vectors.
2. I would like to have HOGSVD (GSVD with more than 2 matrices) in the long run (this does not use LAPACK) and it's appealing to me to roll HOGSVD and GSVD together (so that they are actually generalised). This has some unfortunate issues, so may not be ideal.

### DIFF
--- a/mat64/cholesky.go
+++ b/mat64/cholesky.go
@@ -176,6 +176,29 @@ func (m *Dense) SolveCholesky(chol *Cholesky, b Matrix) error {
 	return nil
 }
 
+// TODO(kortschak): Export this as SolveTwoChol.
+// solveTwoChol finds the matrix m that solves A * m = B where A and B are represented
+// by their Cholesky decompositions a and b, placing the result in the receiver.
+func (m *Dense) solveTwoChol(a, b *Cholesky) error {
+	if !a.valid() || !b.valid() {
+		panic(badCholesky)
+	}
+	bn := b.chol.mat.N
+	if a.chol.mat.N != bn {
+		panic(matrix.ErrShape)
+	}
+
+	m.reuseAsZeroed(bn, bn)
+	m.Copy(b.chol.T())
+	blas64.Trsm(blas.Left, blas.Trans, 1, a.chol.mat, m.mat)
+	blas64.Trsm(blas.Left, blas.NoTrans, 1, a.chol.mat, m.mat)
+	blas64.Trmm(blas.Right, blas.NoTrans, 1, b.chol.mat, m.mat)
+	if a.cond > matrix.ConditionTolerance {
+		return matrix.Condition(a.cond)
+	}
+	return nil
+}
+
 // SolveCholeskyVec finds the vector v that solves A * v = b where A is represented
 // by the Cholesky decomposition, placing the result in the receiver.
 func (v *Vector) SolveCholeskyVec(chol *Cholesky, b *Vector) error {
--- a/mat64/gsvd.go
+++ b/mat64/gsvd.go
@@ -1,0 +1,321 @@
+// Copyright ©2017 The gonum Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+// Based on the SingularValueDecomposition class from Jama 1.0.3.
+
+package mat64
+
+import (
+	"github.com/gonum/blas/blas64"
+	"github.com/gonum/floats"
+	"github.com/gonum/lapack"
+	"github.com/gonum/lapack/lapack64"
+	"github.com/gonum/matrix"
+)
+
+// GSVD is a type for creating and using the Generalized Singular Value Decomposition
+// (GSVD) of a matrix. See Factorize for a description of the decomposition.
+type GSVD struct {
+	kind matrix.GSVDKind
+
+	r, p, c, k, l int
+	s1, s2        []float64
+	a, b, u, v, q blas64.General
+
+	work  []float64
+	iwork []int
+}
+
+// Factorize computes the generalized singular value decomposition (GSVD) of the input
+// the r×c matrix A and the p×c matrix B. The singular values of A and B are computed
+// in all cases, while the singular vectors are optionally computed depending on the
+// input kind.
+//
+// The full singular value decomposition (kind == GSVDU|GSVDV|GSVDQ) deconstructs A and B as
+//  A = U * Σ₁ * [ 0 R ] * Q^T
+//
+//  B = V * Σ₂ * [ 0 R ] * Q^T
+// where Σ₁ and Σ₂ are r×(k+l) and p×(k+l) diagonal matrices of singular values, and
+// U, V and Q are r×r, p×p and c×c orthogonal matrices of singular vectors. k+l is the
+// effective numerical rank of the matrix [ A^T B^T ]^T.
+//
+// It is frequently not necessary to compute the full GSVD. Computation time and
+// storage costs can be reduced using the appropriate kind. Either only the singular
+// values can be computed (kind == SVDNone), or in conjunction with specific singular
+// vectors (kind bit set according to matrix.GSVDU, matrix.GSVDV and matrix.GSVDQ).
+//
+// Factorize returns whether the decomposition succeeded. If the decomposition
+// failed, routines that require a successful factorization will panic.
+func (gsvd *GSVD) Factorize(a, b Matrix, kind matrix.GSVDKind) (ok bool) {
+	r, c := a.Dims()
+	gsvd.r, gsvd.c = r, c
+	p, c := b.Dims()
+	gsvd.p = p
+	if gsvd.c != c {
+		panic(matrix.ErrShape)
+	}
+	var jobU, jobV, jobQ lapack.GSVDJob
+	switch {
+	default:
+		panic("gsvd: bad input kind")
+	case kind == matrix.GSVDNone:
+		jobU = lapack.GSVDNone
+		jobV = lapack.GSVDNone
+		jobQ = lapack.GSVDNone
+	case (matrix.GSVDU|matrix.GSVDV|matrix.GSVDQ)&kind != 0:
+		if matrix.GSVDU&kind != 0 {
+			jobU = lapack.GSVDU
+			gsvd.u = blas64.General{
+				Rows:   r,
+				Cols:   r,
+				Stride: r,
+				Data:   use(gsvd.u.Data, r*r),
+			}
+		}
+		if matrix.GSVDV&kind != 0 {
+			jobV = lapack.GSVDV
+			gsvd.v = blas64.General{
+				Rows:   p,
+				Cols:   p,
+				Stride: p,
+				Data:   use(gsvd.v.Data, p*p),
+			}
+		}
+		if matrix.GSVDQ&kind != 0 {
+			jobQ = lapack.GSVDQ
+			gsvd.q = blas64.General{
+				Rows:   c,
+				Cols:   c,
+				Stride: c,
+				Data:   use(gsvd.q.Data, c*c),
+			}
+		}
+	}
+
+	// A and B are destroyed on call, so copy the matrices.
+	aCopy := DenseCopyOf(a)
+	bCopy := DenseCopyOf(b)
+
+	gsvd.s1 = use(gsvd.s1, c)
+	gsvd.s2 = use(gsvd.s2, c)
+
+	gsvd.iwork = useInt(gsvd.iwork, c)
+
+	gsvd.work = use(gsvd.work, 1)
+	lapack64.Ggsvd3(jobU, jobV, jobQ, aCopy.mat, bCopy.mat, gsvd.s1, gsvd.s2, gsvd.u, gsvd.v, gsvd.q, gsvd.work, -1, gsvd.iwork)
+	gsvd.work = use(gsvd.work, int(gsvd.work[0]))
+	gsvd.k, gsvd.l, ok = lapack64.Ggsvd3(jobU, jobV, jobQ, aCopy.mat, bCopy.mat, gsvd.s1, gsvd.s2, gsvd.u, gsvd.v, gsvd.q, gsvd.work, len(gsvd.work), gsvd.iwork)
+	if ok {
+		gsvd.a = aCopy.mat
+		gsvd.b = bCopy.mat
+		gsvd.kind = kind
+	}
+	return ok
+}
+
+// Kind returns the matrix.GSVDKind of the decomposition. If no decomposition has been
+// computed, Kind returns 0.
+func (gsvd *GSVD) Kind() matrix.GSVDKind {
+	return gsvd.kind
+}
+
+// Rank returns the k and l terms of the rank of [ A^T B^T ]^T.
+func (gsvd *GSVD) Rank() (k, l int) {
+	return gsvd.k, gsvd.l
+}
+
+// GeneralizedValues returns the generalized singular values of the factorized matrices.
+// If the input slice is non-nil, the values will be stored in-place into the slice.
+// In this case, the slice must have length min(r,c)-k, and GeneralizedValues will
+// panic with matrix.ErrSliceLengthMismatch otherwise. If the input slice is nil,
+// a new slice of the appropriate length will be allocated and returned.
+//
+// GeneralizedValues will panic if the receiver does not contain a successful factorization.
+func (gsvd *GSVD) GeneralizedValues(v []float64) []float64 {
+	if gsvd.kind == 0 {
+		panic("gsvd: no decomposition computed")
+	}
+	r := gsvd.r
+	c := gsvd.c
+	k := gsvd.k
+	d := min(r, c)
+	if v == nil {
+		v = make([]float64, d-k)
+	}
+	if len(v) != d-k {
+		panic(matrix.ErrSliceLengthMismatch)
+	}
+	floats.DivTo(v, gsvd.s1[k:d], gsvd.s2[k:d])
+	return v
+}
+
+// ValuesA returns the singular values of the factorized A matrix.
+// If the input slice is non-nil, the values will be stored in-place into the slice.
+// In this case, the slice must have length min(r,c)-k, and ValuesA will panic with
+// matrix.ErrSliceLengthMismatch otherwise. If the input slice is nil,
+// a new slice of the appropriate length will be allocated and returned.
+//
+// ValuesA will panic if the receiver does not contain a successful factorization.
+func (gsvd *GSVD) ValuesA(s []float64) []float64 {
+	if gsvd.kind == 0 {
+		panic("gsvd: no decomposition computed")
+	}
+	r := gsvd.r
+	c := gsvd.c
+	k := gsvd.k
+	d := min(r, c)
+	if s == nil {
+		s = make([]float64, d-k)
+	}
+	if len(s) != d-k {
+		panic(matrix.ErrSliceLengthMismatch)
+	}
+	copy(s, gsvd.s1[k:min(r, c)])
+	return s
+}
+
+// ValuesB returns the singular values of the factorized B matrix.
+// If the input slice is non-nil, the values will be stored in-place into the slice.
+// In this case, the slice must have length min(r,c)-k, and ValuesB will panic with
+// matrix.ErrSliceLengthMismatch otherwise. If the input slice is nil,
+// a new slice of the appropriate length will be allocated and returned.
+//
+// ValuesB will panic if the receiver does not contain a successful factorization.
+func (gsvd *GSVD) ValuesB(s []float64) []float64 {
+	if gsvd.kind == 0 {
+		panic("gsvd: no decomposition computed")
+	}
+	r := gsvd.r
+	c := gsvd.c
+	k := gsvd.k
+	d := min(r, c)
+	if s == nil {
+		s = make([]float64, d-k)
+	}
+	if len(s) != d-k {
+		panic(matrix.ErrSliceLengthMismatch)
+	}
+	copy(s, gsvd.s2[k:d])
+	return s
+}
+
+// ZeroRFromGSVD extracts the matrix [ 0 R ] from the singular value decomposition, storing
+// the result in-place into the receiver. [ 0 R ] is size (k+l)×c.
+func (m *Dense) ZeroRFromGSVD(gsvd *GSVD) {
+	if gsvd.kind == 0 {
+		panic("gsvd: no decomposition computed")
+	}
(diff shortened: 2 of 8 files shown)


btracey:
Do you have a reference you like for part 2? I've never used this decomposition, so I don't understand the tradeoffs involved. Thanks!

kortschak:
[This](http://journals.plos.org/plosone/article?id=10.1371/journal.pone.0028072) is by the person who originally devised it.

mat64/matrix.go
```diff
@@ -883,3 +883,12 @@ func zero(f []float64) {
 		f[i] = 0
 	}
 }
+
+// useInt returns an int slice with l elements, using i if it
+// has the necessary capacity, otherwise creating a new slice.
+func useInt(i []int, l int) []int {
```
sbinet: nit: I wouldn't use `i` and `l` as look alike for people with bad eye sight (or an aging one, like me)
perhaps replace `l` with `n` ?

kortschak: Done.

... and get off my lawn young whippersnapper :)

btracey:
Oh, hey, Michael Saunders. I saw him last week!

btracey:
This paper says the HO GSVD is the same as the 2-matrix GSVD. Thus, it seems that there's no harm in making them one function. In that light, perhaps it should be `Factorize(mats []Matrix)` , `Values(s []float64, i int)`, etc. We can document that at the moment it only works for two matrices, but this gives us an option to backward-compatibly extend to HO GSVD.

kortschak:
In theory, they would work under the same API, whether that works in practice is not yet clear. I have smooshed together an untested and unoptimised implementation of HOGSVD which I'll add tests to and clean up soon. This may end up in this PR as HOGSVD and we can discuss if and how they can be unified.

kortschak:
I have run tests on the HOGSVD code and it looks like numerical stability gets in the way of it working. The issue is the A_i*A_j^-1 calculation which I cannot do any other way than getting the inverse. Comparing it to a python implementation using numpy I can see where we are starting to deviate. further investigation shows that numpy calculates the inverse using `DGESV` rather than `DGETRI` as we do. @btracey is this likely to be the cause?

btracey:
For poorly conditioned matrices, I would imagine that could be a difference. That's the problem with poor conditioning -- numerical instability. Based on the documentation, it seems like `dgetri` should be better as it is explicitly geared toward inverses.

Why can't you do it another way? `(A_i * A_j^-1) = (A_j^-T * A_i^T)^T`, so 

````
tmp.Solve(A_j.T(), A_i.T())
ans.Clone(tmp.T())
````

**Comment by kortschak** [comment]
Thanks. For posterity, here is the python:

```
import numpy as np
def hogsvd(m):
    a = list()
    for d in m:
        aMatTemp = np.dot(d.T,d)
        a.append(aMatTemp)
    #
    s=np.zeros(m[0].shape[1])
    for i in range(len(a)):
        for j in range(i+1,len(a)):
            s = s + np.dot(a[i],np.linalg.inv(a[j])) + np.dot(a[j],np.linalg.inv(a[i]))
    #
    n = len(m)
    s = (n*(n-1))*s
    #
    _, v = np.linalg.eig(s)
    #
    b = list()
    invV = np.linalg.inv(v)
    for d in m:
        b.append(np.dot(invV,d.T).T)
    #
    sigma = list()
    u = list()
    for x in b:
        sig = np.sqrt(np.sum(x*x, axis=0))
        u.append(x/sig)
        sigma.append(np.diag(sig))
    return(u,sigma,v)
```

**Comment by btracey** [comment]
The eigenvectors are the same, although one of them has a sign flip. However, they are in a different order, and I believe they should be returned with the same ordering as the magnitude of the eigenvalue. Do Go and Python disagree on which eigenvalue is bigger, or do they disagree in the association between eigenvalues and eigenvectors?

**Comment by btracey** [comment]
Ohh, this is using the old Eigen code. Did you try calling the lapack-based Eigenvalue solver to see if that resolves the issue? It's possible there is a bug in the old code.

**Comment by kortschak** [comment]
Yeah, it's all OK now. They differences (which were surprising because I thought we were using LAPACK) are due to the old Eigen code and were making finding my actual errors harder. I massaged the v and then find the errors. It works now and has been committed so you can see the issues with API unification; I don't think they can be unified.

File: mat64/gsvd.go [hunk 2]
```diff
@@ -0,0 +1,296 @@
...
+	if gsvd.kind == 0 {
+		panic("gsvd: no decomposition computed")
+	}
+	m.reuseAsZeroed(gsvd.k+gsvd.l, gsvd.n)
+	a := Dense{
+		mat:     gsvd.a,
+		capRows: gsvd.m,
+		capCols: gsvd.n,
+	}
+	m.Slice(0, min(gsvd.k+gsvd.l, gsvd.m), gsvd.n-gsvd.k-gsvd.l, gsvd.n).(*Dense).
```
vladimir-ch: Extract the fields into local variables for better readability?

mat64/gsvd.go
```diff
@@ -0,0 +1,296 @@
...
+		m.set(i, i+gsvd.k, gsvd.s[gsvd.k+i])
+	}
+	for i := gsvd.m - gsvd.k; i < gsvd.l; i++ {
+		m.set(i, i+gsvd.k, 1)
+	}
+}
+
+// UFromGSVD extracts the matrix U from the singular value decomposition, storing
+// the result in-place into the receiver. U is size m×m.
+func (m *Dense) UFromGSVD(gsvd *GSVD) {
```
vladimir-ch: This signature goes against the pattern established in #436. Also in other places in this PR.

kortschak: This PR predates #436. When #436 is merged, this PR will be updated to follow the pattern set there,

kortschak:
The failure above seems to be real and unexpected. I have reproduced it with a direct translation into the Dggsvd3 and Dggsvp3 tests. Filed https://github.com/gonum/lapack/issues/267.

mat64/gsvd.go
```diff
@@ -0,0 +1,296 @@
...
+
+import (
+	"github.com/gonum/blas/blas64"
+	"github.com/gonum/floats"
+	"github.com/gonum/lapack"
+	"github.com/gonum/lapack/lapack64"
+	"github.com/gonum/matrix"
+)
+
+// GSVD is a type for creating and using the Generalized Singular Value Decomposition
```
btracey: I think there is merit in some kind of description of what it does (combined singular vectors), a link to more information, and some comment about the related HOGSVD

mat64/gsvd.go
```diff
@@ -0,0 +1,296 @@
...
+// Factorize computes the generalized singular value decomposition (GSVD) of the input
+// the m×n matrix A and the p×n matrix B. The singular values of A and B are computed
+// in all cases, while the singular vectors are optionally computed depending on the
+// input kind.
+//
+// The full singular value decomposition (kind == GSVDU|GSVDV|GSVDQ) deconstructs A and B as
+//  A = U * Σ₁ * [ 0 R ] * Q^T
+//
+//  B = V * Σ₂ * [ 0 R ] * Q^T
+// where Σ₁ and Σ₂ are m×(k+l) and p×(k+l) diagonal matrices of singular values, and
```
btracey: What is k? I don't see a description of how I know how big it is.

mat64/gsvd.go
```diff
@@ -0,0 +1,296 @@
...
+	return ok
+}
+
+// Kind returns the matrix.GSVDKind of the decomposition. If no decomposition has been
+// computed, Kind returns 0.
+func (gsvd *GSVD) Kind() matrix.GSVDKind {
+	return gsvd.kind
+}
+
+// Rank returns the k and l terms of the rank of [ A^T B^T ]^T.
```
btracey: This last transpose isn't needed, rank(A) = rank(A^T)

mat64/gsvd.go
btracey: Oh, same with l (github should have better font choices!)

mat64/gsvd.go
```diff
@@ -0,0 +1,296 @@
...
+func (gsvd *GSVD) Rank() (k, l int) {
+	return gsvd.k, gsvd.l
+}
+
+// GeneralizedValues returns the generalized singular values of the factorized matrix
```
btracey: I don't understand this function (reading from documentation). There isn't "a" factorized matrix, there are two. I'm not sure which values I'm getting here, the original documentation has both Sigma1 and Sigma2

mat64/gsvd.go
```diff
@@ -0,0 +1,296 @@
...
+		v = make([]float64, min(gsvd.m, gsvd.n)-gsvd.k)
+	}
+	if len(v) != min(gsvd.m, gsvd.n)-gsvd.k {
+		panic(matrix.ErrSliceLengthMismatch)
+	}
+	floats.DivTo(v, gsvd.c[gsvd.k:min(gsvd.m, gsvd.n)], gsvd.s[gsvd.k:min(gsvd.m, gsvd.n)])
+	return v
+}
+
+// Values1 returns the singular values of the factorized matrix.
```
btracey: Maybe ValuesA and ValuesB?

btracey: "the factorized matrix A". Need to distinguish between the two.

mat64/gsvd.go
```diff
@@ -0,0 +1,296 @@
...
+		c = make([]float64, min(gsvd.m, gsvd.n)-gsvd.k)
+	}
+	if len(c) != min(gsvd.m, gsvd.n)-gsvd.k {
+		panic(matrix.ErrSliceLengthMismatch)
+	}
+	copy(c, gsvd.c[gsvd.k:min(gsvd.m, gsvd.n)])
+	return c
+}
+
+// Values2 returns the singular values of the factorized matrix.
```
btracey: Same here.

mat64/gsvd.go
```diff
@@ -0,0 +1,296 @@
...
+			mat:     gsvd.b,
+			capRows: gsvd.p,
+			capCols: gsvd.n,
+		}
+		m.Slice(gsvd.m, gsvd.k+gsvd.l, gsvd.n+gsvd.m-gsvd.k-gsvd.l, gsvd.n).(*Dense).
+			Copy(b.Slice(gsvd.m-gsvd.k, gsvd.l, gsvd.n+gsvd.m-gsvd.k-gsvd.l, gsvd.n))
+	}
+}
+
+// Sigma1FromGSVD extracts the matrix Σ₁ from the singular value decomposition, storing
```
btracey: If Values becomes A and B this should too.

mat64/gsvd.go
```diff
@@ -0,0 +1,296 @@
...
+//  A = U * Σ₁ * [ 0 R ] * Q^T
+//
+//  B = V * Σ₂ * [ 0 R ] * Q^T
+// where Σ₁ and Σ₂ are m×(k+l) and p×(k+l) diagonal matrices of singular values, and
+// U, V and Q are m×m, p×p and n×n orthogonal matrices of singular vectors.
+//
+// It is frequently not necessary to compute the full GSVD. Computation time and
+// storage costs can be reduced using the appropriate kind. Either only the singular
+// values can be computed (kind == SVDNone), or in conjunction with specific singular
+// vectors (kind bit set according to matrix.GSVDU, matrix.GSVDV and matrix.GSVDQ).
```
btracey: Provide a quick example (for example, gsvd.Factorize(a,b, matrix.GSVDU|matrix.GSVDQ) for both ...)

mat64/gsvd_test.go
```diff
@@ -0,0 +1,118 @@
...
+		{200, 200, 150},
+		{150, 200, 200},
+	} {
+		m := test.m
+		p := test.p
+		n := test.n
+		for trial := 0; trial < 10; trial++ {
+			a := NewDense(m, n, nil)
+			for i := range a.mat.Data {
+				a.mat.Data[i] = rand.NormFloat64()
```
btracey: Shouldn't this be using a seeded number generator?

mat64/hogsvd.go
```diff
@@ -0,0 +1,180 @@
...
+//
+//  M_0 = U_0 * Σ_0 * V^T
+//  M_1 = U_1 * Σ_1 * V^T
+//  .
+//  .
+//  .
+//  M_{n-1} = U_{n-1} * Σ_{n-1} * V^T
+//
+// where U_i are r_i×c matrices of singular vectors, Σ are c×c matrices singular values, and V
+// is a c×c matrix of singular vectors.
```
btracey: Again, some rough description of how they're combined, plus a link to more info.

mat64/hogsvd.go
```diff
@@ -0,0 +1,180 @@
...
+		a[i].Mul(d.T(), d)
+	}
+
+	s := getWorkspace(c, c, true)
+	defer putWorkspace(s)
+	sij := getWorkspace(c, c, false)
+	defer putWorkspace(sij)
+	for i, ai := range a {
+		for _, aj := range a[i+1:] {
+			gsvd.err = sij.Solve(ai.T(), aj.T())
```
btracey: `a[i] = d^T * d`, so a is a positive definite matrix. It seems like this would be better all around if this used the Cholesky decomposition for the solve. For instance, compute all the choleskies before the a_i range, and so the "inverses" are cached for both Solves. This alone should cut out a bunch of N^3 steps. There may also be something tricky about linear solves with two positive definite matrices.

btracey: See for example Wishart.logProbSymChol. It seems like this is a more common primitive, so maybe we could add a symDense.SolveTwoChol(a,b Cholesky) (or something).

mat64/hogsvd.go
```diff
@@ -0,0 +1,180 @@
...
+			if gsvd.err != nil {
+				return false
+			}
+			s.Add(s, sij.T())
```
btracey: Isn't sij actually symmetric? It's the linear solve from two symmetric matrices

mat64/hogsvd.go
```diff
@@ -0,0 +1,180 @@
...
+			gsvd.err = sij.Solve(aj.T(), ai.T())
+			if gsvd.err != nil {
+				return false
+			}
+			s.Add(s, sij.T())
+		}
+	}
+	s.Scale(1/float64(len(m)*(len(m)-1)), s)
+
+	var eig Eigen
```
btracey: As far as I can tell, `s` is also a symmetric matrix, since it's the sum of symmetric matrices. Thus, this can use EigSym?

mat64/gsvd.go
kortschak: This is as described in DGGSVD3.

mat64/gsvd_test.go
kortschak: Fixed. But aren't we lucky I did this; it found the flakiness in `Det` :)

mat64/hogsvd.go
kortschak: No, sij is not symmetric.

mat64/hogsvd.go
kortschak: No, since sij is not.

mat64/gsvd.go
kortschak: This is difficult. It comes from the DGGSVD3 documentation.

mat64/gsvd.go
kortschak: Same here.

kortschak:
I had HOGSVD working a while ago, but the new Eigen appears to have broken it. I will look into this at home where I have the working version.

mat64/gsvd.go
btracey: I want to say "Surely there's a description somewhere of why one wants this". I did some quick looking and failed to find anything more useful. But it still remains that presumably you're implementing this for a purpose. Maybe a link to that is more useful than nothing?

mat64/gsvd.go
btracey: Could you add though, the " Let K+L = the effective numerical rank of the matrix (A**T,B**T)**T" into the main description. There's no comment on what k and l are in the factorize documentation.

mat64/gsvd.go
kortschak: My interest is from [this paper](http://journals.plos.org/plosone/article?id=10.1371/journal.pone.0030098), but there are other uses. I don't think we have put more information on other things (for example SVD) than the properties defined by the equations.

mat64/gsvd.go
kortschak: Done.

mat64/gsvd.go
kortschak: Done.

mat64/gsvd.go
kortschak: Done.

mat64/gsvd.go
kortschak: Mistyped. The doc should say "matrices".

mat64/hogsvd.go
kortschak: Again, in other places we have not provided more information than the properties of the decomposition. The use case I have is in [this paper](http://journals.plos.org/plosone/article?id=10.1371/journal.pone.0028072).

kortschak: I have added the sentence "HOGSV extends the GSVD case from 2 to n input matrices."

mat64/hogsvd.go
kortschak: I'll do that in two steps here. There is an obvious optimisation of using `SymDense` for the `a`s. Then I'll add the `solveTwoChol` as a little helper on `*Dense` (it can't be on `*SymDense` as the results are not symmetric.

kortschak: Done.

kortschak:
PTAL

mat64/hogsvd.go
```diff
@@ -111,6 +118,29 @@ func (gsvd *HOGSVD) Factorize(m ...Matrix) (ok bool) {
...
+	}
+	an := a.chol.mat.N
+	bn := b.chol.mat.N
+	if an != bn {
+		panic(matrix.ErrShape)
+	}
+
+	m.reuseAsZeroed(bn, bn)
+	m.Copy(b.chol)
+	blas64.Trsm(blas.Left, blas.Trans, 1, a.chol.mat, m.mat)
```
kortschak: I don't think this is right, but I'm not sure why it passes.

mat64/gsvd.go
btracey: Okay, I would say to add a reference to Factorize for a description of the decomposition.

mat64/hogsvd.go
```diff
@@ -0,0 +1,211 @@
...
+		panic(badCholesky)
+	}
+	an := a.chol.mat.N
+	bn := b.chol.mat.N
+	if an != bn {
+		panic(matrix.ErrShape)
+	}
+
+	m.reuseAsZeroed(bn, bn)
+	m.Copy(b.chol)
```
btracey: UFromCholesky instead

mat64/hogsvd.go
btracey: Me either. Here is what I think it should be

````
Σ_A ^-1 Σ_B

(A^T * A)^-1 * (B^T * B)
A^-1 * A^-T * B^T * B
A^-1 * C * B
D * B
````
To get C, we have `A^T* C = B^T`, which is Trsm with Trans, Left, but also B.T() instead of just B. You have the trans and left, but it seems like the wrong RHS.
Then we have `A * D = C` which is the Left and NoTrans that you have.
It seems strange that this works despite totally dropping the last B. You seem to be computing `A^-1 * A^-T * B`, which doesn't seem like the same thing.

File: mat64/hogsvd.go (hunk 16 above)
Comment (btracey) [inline mat64/hogsvd.go]: We can though, add up the non-transposed S, and then compute eig.Factorize(s.T()), though, right?

File: mat64/hogsvd.go (hunk 18 above)
Comment (btracey) [inline mat64/hogsvd.go]: Nevermind

File: mat64/gsvd.go (hunk 4 above)
Comment (kortschak) [inline mat64/gsvd.go]: This should probably be applied to other `Factorizor`s then.

File: mat64/hogsvd.go (hunk 16 above)
Comment (kortschak) [inline mat64/hogsvd.go]: Nice. Done.

File: mat64/gsvd.go (hunk 10 above)
Comment (kortschak) [inline mat64/gsvd.go]: Done.

File: mat64/hogsvd.go (hunk 17 above)
Comment (kortschak) [inline mat64/hogsvd.go]: Done. I think.

Comment (kortschak) [inline mat64/hogsvd.go]: Not done. I have added a test for this though - failing.

File: mat64/gsvd.go (hunk 2 above)
Comment (kortschak) [inline mat64/gsvd.go]: I'll go through and s/m/r/g and s/n/c/g since otherwise I have collisions with the receiver name.

Comment (kortschak) [inline mat64/gsvd.go]: Done.

File: mat64/cholesky.go [hunk 19]
```diff
@@ -176,6 +176,29 @@ func (m *Dense) SolveCholesky(chol *Cholesky, b Matrix) error {
...
+	if a.chol.mat.N != bn {
+		panic(matrix.ErrShape)
+	}
+
+	m.reuseAsZeroed(bn, bn)
+	m.Copy(b.chol.T())
+	blas64.Trsm(blas.Left, blas.Trans, 1, a.chol.mat, m.mat)
+	blas64.Trsm(blas.Left, blas.NoTrans, 1, a.chol.mat, m.mat)
+	blas64.Trmm(blas.Right, blas.NoTrans, 1, b.chol.mat, m.mat)
+	if cond := math.Max(a.cond, b.cond); cond > matrix.ConditionTolerance {
```
btracey: The condition number of B doesn't matter here since it's A that has the inverse. B can be a matrix of zeros and the solve is okay (though the Cholesky won't be)

btracey:
LGTM

mat64/cholesky.go
kortschak: If the Cholesky of B won't be OK, doesn't that mean that the final `Trmm` won't adjust the result properly?

kortschak:
I am waiting on some data to make a pair of examples for HOGSVD and GSVD. I can merge this and make that an issue if you want to move ahead with the monorepoisation. But I will wait for the query above.

mat64/cholesky.go
btracey: I don't understand the question, sorry. The function should require that the Cholesky decompositions are okay, but I believe the condition error should only be for the matrix being inverted.

kortschak: Understood now. Done.

PR #436, opened 2017-03-18
mat64: reverse signatures for decomposition matrix extractions

### LINKED ISSUE #435: mat64: rethink XFromY API for extracting matrices from decompositions
When there were few of these it made sense to have the extractions be a method on `*Dense` or other matrix types. However, now (partly due to me) there is a proliferation of these on `*Dense`. Having just had a look at `stat.CC` I'm not averse to the approach that is used there with `Left` and `Right`. We should consider migrating to this approach (at least for some/most? of the decompositions).

This change would have the advantage of keeping the documentation around the decompositions together for each type of decomposition, would reduce the number of methods on `*Dense` and would avoid giving the impression that there are blessed decompositions (that packages outside mat64 cannot match).

kortschak:
Also provides some sugar to ease this.

Closes #435.

### DIFF
--- a/mat64/lq.go
+++ b/mat64/lq.go
@@ -36,7 +36,7 @@ func (lq *LQ) updateCond() {
 //
 // The LQ decomposition is a factorization of the matrix A such that A = L * Q.
 // The matrix Q is an orthonormal n×n matrix, and L is an m×n upper triangular matrix.
-// L and Q can be extracted from the LFromLQ and QFromLQ methods on Dense.
+// L and Q can be extracted from the LTo and QTo methods.
 func (lq *LQ) Factorize(a Matrix) {
 	m, n := a.Dims()
 	if m > n {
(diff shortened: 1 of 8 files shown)


btracey:
Why the "To" endings? It seems like `L` (for instance) is just as clear.

kortschak:
There are two usage syntaxes, `dst = x.PTo(nil)` and `x.PTo(dst)`. The first is obviously an assignment to `dst`, but the second would be less obvious in the case that it were written `x.P(y)`. The `To` suffix clears that use intention (at a minor cost in the first case where `To(nil)` is a little upsetting, but I think the trade-off is worth it (and could be apologised). It lends its origin from `io.WriterTo`.

kortschak:
I'll wait for @btracey to make sure he's happy with my argument above.

btracey:
Thinking about it, I think the related argument is that usually the receiver is modified. This is a difference, so that difference should be clear.

btracey:
LGTM

btracey:
This is a pretty big breaking change, at least for code I've written. Is it possible we could merge repos into gonum/gonum and then make this change? It sounds like there's nothing in the way of migration that will change.

kortschak:
I'm happy to wait. This is relatively small and I'll just copy past over to the new location when we are ready if that's what you want.

kortschak:
Curious though. How does delaying the merge of this help?

btracey:
My suggestion was that this is a (decent-sized) breaking change, and it might be nice to delay that until the merge. This way, old code can remain working without update.

kortschak:
Done in gonum/gonum.

Issue #437, opened 2017-03-29
mat64: TestDet is flakey


kortschak:
With the addition of a new independent test, TestDet now fails with

```
--- FAIL: TestDet (0.00s)

	list_test.go:605: Answer mismatch: DetVsChol(*mat64.basicSymmetric), size: struct { ar int; ac int }{ar:6, ac:6}, atrans true
```

I imagine this is due to an unfortunate random value stream position. Changing the tolerance to 1e-11 allows the test to pass, but it is curious that it only fails with `*mat64.basicSymmetric` and not `*mat64.SymDense` (though maybe the bad stream is very rare), so I don't want to just rush in and broaden the tolerance for this test.

/cc @btracey

kortschak:
`go test -run TestDet -count 1000 | grep -c FAIL`  gives 82 in a variety of types. So it's just rare.

kortschak:
Closed by #438.

PR #438, opened 2017-03-29
mat64: deflake TestDet

kortschak:
This allows 4000 tests of TestDet to pass, but the test is still flakey as it panics due to a Cholesky decomposition failure before 5000 tests.

@btracey Please take a look.

Updates #437.


btracey:
LGTM

PR #451, opened 2017-06-24
improve the dense tranpose and add matnum package, it easy to use

brucexiaok:
I add the new package matnum to improve the matrix operation ,it's easy to use ,but beta,i want to know that why author use type Transpose and type UnTranspose, it is hard to use
and the list of this code also made be confused:

	aU, _ := untranspose(a)
	bU, _ := untranspose(b)
	m.reuseAs(ar, ac)

	if arm, ok := a.(RawMatrixer); ok {
		if brm, ok := b.(RawMatrixer); ok {
			amat, bmat := arm.RawMatrix(), brm.RawMatrix()
			if m != aU {
				m.checkOverlap(amat)
			}
			if m != bU {
				m.checkOverlap(bmat)
			}
			for ja, jb, jm := 0, 0, 0; ja < ar*amat.Stride; ja, jb, jm = ja+amat.Stride, jb+bmat.Stride, jm+m.mat.Stride {
				for i, v := range amat.Data[ja : ja+ac] {
					m.mat.Data[i+jm] = v * bmat.Data[i+jb]
				}
			}
			return
		}
	}

	var restore func()
	if m == aU {
		m, restore = m.isolatedWorkspace(aU)
		defer restore()
	} else if m == bU {
		m, restore = m.isolatedWorkspace(bU)
		defer restore()
	}
code from dense mul func
i think it's useless

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -132,6 +132,7 @@ var (
 	ErrSliceLengthMismatch = Error{"matrix: input slice length mismatch"}
 	ErrNotPSD              = Error{"matrix: input not positive symmetric definite"}
 	ErrFailedEigen         = Error{"matrix: eigendecomposition not successful"}
+	ErrTypeAssert	     = Error{"matrix: type assert error"}
 )
 
 // ErrorStack represents matrix handling errors that have been recovered by Maybe wrappers.
--- a/mat64/dense.go
+++ b/mat64/dense.go
@@ -198,7 +198,23 @@ func (m *Dense) Caps() (r, c int) { return m.capRows, m.capCols }
 
 // T performs an implicit transpose by returning the receiver inside a Transpose.
 func (m *Dense) T() Matrix {
-	return Transpose{m}
+	r, c := m.Dims()
+	data := make([]float64, r*c)
+	for i := 0; i < r; i++ {
+		for j := 0; j < c; j++ {
+			data[j*r+i] = m.mat.Data[i*c+j]
+		}
+	}
+	return &Dense{
+		mat: blas64.General{
+			Rows:   c,
+			Cols:   r,
+			Stride: r,
+			Data:   data,
+		},
+		capRows: c,
+		capCols: r,
+	}
 }
 
 // ColView returns a Vector reflecting the column j, backed by the matrix data.
--- a/mat64/dense_test.go
+++ b/mat64/dense_test.go
@@ -1871,6 +1871,12 @@ func TestInverse(t *testing.T) {
 	}
 }
 
+func TestTranspose(t *testing.T) {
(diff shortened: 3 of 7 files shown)


btracey:
Thank you for the Pull Request. However, we are no longer accepting pull requests to this package. The new code is in the gonum repositiory (github.com/gonum/gonm)

Also
- In the future, please run `go test` before submitting pull requests. This code does not even compile.
- The implementation of transpose means there is a lot of work and memory allocation. The current implementation avoids that
- The current Matrix interface has very few methods for a reason. A symmetric matrix cannot result from adding arbitrary matrices
- I don't understand the matnum package. It looks like it is re-implementing `Dense` and `Vector` from mat64

kortschak:
@brucexiaok Before sending PRs that make wholesale changes to internal implementation it is worth while to file and issue raising the need for the change, or posting to gonum-dev asking whether others agree with the proposal. This is particularly the case when the need for the change is founded on not seeing the utility of an implementation detail.

kortschak:
Sorry, further to this. If there are particular issues you are having using gonum mat or mat64 feel free to post to gonum-dev to ask for advice or help. If you do this, please post your client code or links to it so we can know what you are doing and make suggestions.

From the code that you have in matnum though it looks like you are wanting to use an interface. This is not necessary. If you have matrices that you know are general, then just keep them as a `*Dense`. Then you have access to all the `*Dense` methods. Remember that in Go, interfaces are for code reuse, not a language feature that must be used.

brucexiaok:
@kortschak , thank you very much,and the demand of matnum is that I think some design is hard to use,for example,if i want to multiply the dense ,must NewDense first,why not : 

newDense := newDense.Mul(aDense,bDense)

it easy to write program,if you save result to receiver it means the result must the same dims of receiver,and the type transpose is also have some problem,if i want to convert the type transpose to Dense or others it's very hard,So i think if there is a common interface matrix would be better maybe.

thank you very much !

kortschak:
With the code being `newDense := newDense.Mul(aDense,bDense)` we cannot use interfaces are liberally as we currently can, unless the `Mul` method returns a `mat64.Matrix`, in which case we need to type assert for each operation. Personally, I prefer a new line over `.(*mat64.Dense)`. Note that a similar, approach is used in the math/big package (for similar though not identical reasons).

We have a Reset method on a number of the matrix types exactly to allow the reuse of matrix values without knowing the input dimensions. Though note here that we lean very heavily towards the notion that just being able to assign to a value that may not match dimensionally is probably wrong and will likely cause your calculations to lead you astray. The approach our design favours may be slightly more work up front in the writing, but it clearer to read and will much more likely catch bugs that you have written. We think this is a win.

For examples of how the mat package package is used, look through the stat package.

Finally, note that the design is not going to change in the foreseeable future.

brucexiaok:
Thanks !

